PR #9, opened 2023-05-10
Parser checks, reserved words

dogversioning:
### Description
- Stricter handling of double underscores
- Reserves some keywords in table names
- Lints the test directory

### Checklist

### DIFF
--- a/cumulus_library/study_parser.py
+++ b/cumulus_library/study_parser.py
@@ -20,6 +20,8 @@ from cumulus_library.template_sql.templates import (
 
 StrList = List[str]
 
+RESERVED_TABLE_KEYWORDS = ["etl", "nlp", "lib"]
+
 
 class StudyManifestParsingError(Exception):
     """Basic error for StudyManifestParser"""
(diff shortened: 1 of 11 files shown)


cumulus_library/study_parser.py
```diff
@@ -159,6 +161,15 @@ def clean_study(
         if not view_table_list:
             return view_table_list
 
+        # We'll do a pass to see if any of these tables were created outside of a
+        # study builder, and remove them from the list.
+        for view_table in view_table_list:
+            if any(
+                f"{self.get_study_prefix()}__{word}_" in view_table[0]
```
mikix, line 168: nit: could be a `startswith` check, cause otherwise they could prepend stuff and stop using the prefix. But not important.

mikix (approved):

cumulus_library/study_parser.py
dogversioning: that should be handled before this by prefix checking when the table is created - but i can make that change.

cumulus_library/study_parser.py
```diff
@@ -287,6 +298,34 @@ def _execute_build_queries(
         :param task: a TaskID for a given progress bar
         """
         for query in queries:
+            create_line = query[0].split("\n")[0]
+            if f"{self.get_study_prefix()}__" not in create_line:
```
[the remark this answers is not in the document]
dogversioning: I'll put a single space in front - this should handle it without resorting to regex.

PR #11, opened 2023-05-10
docs: prepare the docs/ folder for ingestion by jekyll

mikix:
### Description
The main documentation repo ([smart-on-fhir/cumulus](https://github.com/smart-on-fhir/cumulus)) will use jekyll and GitHub Pages to build docs from this repo and others into one main site.

But in order to do that, we need some little jekyll metadata markers and to organize our docs in a specific way.

- Add docs/index.md (intro) and docs/README.md (dev explainer)
- Add metadata like titles and nav order for all docs

### How this is going to work technically

Right now, the docs site is rebuilt anytime the main docs repo is changed or on a manual request. But eventually, I will add a GitHub action here that will go poke the main docs repo to rebuild itself, if the docs/ folder here changes.

### How this is going to change how we write/edit docs

- Each Cumulus project will have its own little `docs/` fiefdom in its own repo.
  - This way, docs can update as code changes more easily.
- There will be one main docs repo that pulls docs from all the other repos, following certain conventions below.
  - This will also be able to hold certain project-wide docs or cross-repo docs that don't have a natural fit otherwise.

### Draft of document conventions/advice I'm going to put in the main repo

- Everything in `docs/` is going to be accessible on public docs site (direct URL at minimum)
  - Which might affect the tone of your text as well as which files go in there
  - Except for `docs/README.md` which we won't include, so that you can explain what the
    folder is about and point at this repo.
  - Note that this means you can still put sample json, yaml, etc in there.
- Include a `docs/index.md` as your root & intro doc
- Use a toplevel title that makes sense in context of the whole Cumulus project
  - i.e. "Library" instead of "Cumulus Library" probably (though in your doc prose and headers,
    "Cumulus Library" would make sense -- but your main index.md title will be used in the main
    documentation sidebar)
- Use caution when linking:
  - When linking to _code_, use absolute links
  - When linking to other docs _inside_ your repo, use relative links
  - When linking to other docs _outside_ your repo, only link to the following URLs,
    that we promise will exist:
    - https://docs.smarthealthit.org/cumulus/
    - https://docs.smarthealthit.org/cumulus/library/
- Consider audience & tone when editing docs, to help keep a uniform writing style.
  - It can be helpful to consider what [sort of document](https://diataxis.fr/)
    you are writing.

### Pic of how this looks

Screenshot from 2023-05-10 10-59-10

### Pic of how GitHub renders the metadata

This re-enforces that we can only really have our docs be beautiful in one place, not two - gotta choose the web or GitHub.

image

### Checklist

### DIFF
--- a/docs/aws-setup.md
+++ b/docs/aws-setup.md
@@ -1,7 +1,11 @@
-<!-- Target audience: 
-    - clinical researcher, low AWS familiarity
-    - engineer w/ some AWS experience
-reference tone -->
+---
+title: AWS Setup
+parent: Library
+nav_order: 2
+# audience: clinical researcher, low AWS familiarity or engineer w/ some AWS experience
+# type: reference
+---
+
 # AWS setup
 
 Cumulus library executes queries against an 
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 # project specific
 data_export/
+/.idea/
 .python-version
 .DS_Store
 cumulus_library_columns.json
--- a/README.md
+++ b/README.md
@@ -2,4 +2,4 @@
 
 A framework for designing, executing, and distributing SQL queries packaged as "studies", e.g., for quality monitoring, clinical research, or public health. Part of the [SMART on FHIR Cumulus Project](https://smarthealthit.org/cumulus-a-universal-sidecar-for-a-smart-learning-healthcare-system/)
 
-See [first time setup](./docs/first-time-setup.md) for configuring and running, [creating studies](./docs/creating-studies.md) for using Cumulus Library as part of clinical research, and [core study details](./docs/core-study-details.md) for information about the core dataset.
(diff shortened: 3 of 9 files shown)


docs/aws-setup.md
```diff
@@ -1,7 +1,11 @@
-<!-- Target audience: 
-    - clinical researcher, low AWS familiarity
-    - engineer w/ some AWS experience
-reference tone -->
+---
+title: AWS Setup
+parent: Library
+nav_order: 2
+# audience: clinical researcher, low AWS familiarity or engineer w/ some AWS experience
+# type: reference
```
mikix, line 6: I converted all the little audience & tone markers to audience & type comments here -- dropped all the "helpful direct" comments as the default value -- if we want to be casual or some other tone, we can say that explicitly.

dogversioning (approved):
Honestly, the PR description here is so good it should go in the main docs repo as a 'how do we manage this' description.

PR #13, opened 2023-05-11
Added flexible paths & study template

dogversioning:
### Description
- Adds a new CLI option for quick creation of manifest templates
- Adds a new CLI option for adding paths for studies
- Removes shortcut options for config items which are ususally specified by environment variable

### Checklist

### DIFF
--- a/cumulus_library/cli.py
+++ b/cumulus_library/cli.py
@@ -6,7 +6,7 @@ import os
 import sys
 
 from pathlib import Path, PosixPath
-from typing import Dict
+from typing import Dict, List, Optional
 
 import pyathena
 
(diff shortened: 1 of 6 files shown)


cumulus_library/studies/template/manifest.toml
```diff
@@ -1,8 +1,3 @@
-# If you'd like to run this template study as a proof of concept, you can load
```
dogversioning, line 1: I deleted this since we're now copying it when you're creating a template. The equivalent instructions are now in first-time-setup.md

docs/creating-studies.md
```diff
@@ -80,8 +80,13 @@ Most users have a workflow that looks like this:
   - Build your study with the CLI to make sure your queries load correctly.
 
 We use [sqlfluff](https://github.com/sqlfluff/sqlfluff) to help maintain a consistent
-style across many different SQL query authors. There are two commands you may want to
-run inside your study's directory:
+style across many different SQL query authors. You don't have to if you are working
+on development, but if you're going to submit your study when completed back to
+the Cumulus project, it may be worth mimicing our linting strategy. You can copy our 
+[sqlfluff configuration rules](https://github.com/smart-on-fhir/cumulus-library-core/blob/main/pyproject.toml)
+into your study directory to configure the right behavior.
```
dogversioning, line 87: so - i :could: automatically create a .sqlfluff file in study directories? I'd have to move the config out of the pyproject.toml though, which i'm softly reluctant to do. We also might expect that we just have to do this on our end during a PR.

cumulus_library/cli.py
```diff
@@ -113,73 +113,126 @@ def export_all(self, study_dict: Dict):
             self.export_study(study_dict[key])
 
 
-def get_study_dict() -> Dict[str, PosixPath]:
+def get_abs_posix_path(path: str) -> PosixPath:
+    """Conveince method for hanlding abs vs rel paths"""
+    if path[0] == "/":
+        return Path(path)
+    else:
+        return Path(Path.cwd(), path)
```
mikix, line 121: yelling at clouds: "current working directory" feels redundant. You just need c or w but not both

cumulus_library/cli.py
```diff
@@ -113,73 +113,126 @@ def export_all(self, study_dict: Dict):
...
+
+
+def create_template(path: str) -> None:
+    """Creates a manifest in target dir if one doesn't exist"""
+    abs_path = get_abs_posix_path(path)
+    manifest_path = Path(abs_path, "manifest.toml")
+    if manifest_path.exists():
```
mikix, line 128: nit: The following might be too strict for your use case, but as prior art: if you give the ETL an export directory, it errors out if there are any files in the export folder at all. What you have is good and maybe they already made some SQL there...?

cumulus_library/cli.py
```diff
@@ -113,73 +113,126 @@ def export_all(self, study_dict: Dict):
...
 
     :returns: A list of pathlib.PosixPath objects
     """
     manifest_studies = {}
-    library_path = Path(__file__).resolve().parents[0]
-    for path in Path(f"{library_path}/studies").iterdir():
-        if path.is_dir():
-            manifest_studies[path.name] = path
+    paths = [Path(Path(__file__).resolve().parents[0], "studies")]
+    if alt_paths is not None:
```
mikix, line 144: nit: eh let em error out if they break your type hinting of a required arg  :smile:

cumulus_library/cli.py
```diff
@@ -113,73 +113,126 @@ def export_all(self, study_dict: Dict):
...
-        args["profile"],
-        args["database"],
-    )
-    if args["verbose"]:
-        builder.verbose = True
+    if not args["build"] and not args["export"] and not args["create"]:
+        sys.exit(
+            (
+                "Expecting at least one of build, export or create as arguments. "
+                "See `cumulus_library --help` for more information"
```
mikix, line 161: Hyphen here

cumulus_library/cli.py
```diff
@@ -113,73 +113,126 @@ def export_all(self, study_dict: Dict):
...
+    for parent_path in paths:
+        for child_path in parent_path.iterdir():
+            if child_path.is_dir():
+                manifest_studies[child_path.name] = child_path
+            elif child_path.name == "manifest.toml":
+                manifest_studies[parent_path.name] = parent_path
     return manifest_studies
 
 
 def run_cli(args: Dict):  # pylint: disable=too-many-branches
```
mikix, line 155: lol I didn't see this pylint disable before -- it may honestly be trying to help here

cumulus_library/cli.py
```diff
@@ -113,73 +113,126 @@ def export_all(self, study_dict: Dict):
...
+            if "all" in args["target"]:
+                builder.export_all(study_dict)
+            else:
+                for target in args["target"]:
+                    builder.export_study(study_dict[target])
 
-    # returning the builder for ease of unit testing
-    return builder
+        # returning the builder for ease of unit testing
+        return builder
```
mikix, line 199: You might not want to indent the return here

cumulus_library/cli.py
```diff
@@ -113,73 +113,126 @@ def export_all(self, study_dict: Dict):
...
+    if not args["build"] and not args["export"] and not args["create"]:
+        sys.exit(
+            (
+                "Expecting at least one of build, export or create as arguments. "
+                "See `cumulus_library --help` for more information"
+            )
+        )
+
+    if args["create"]:
+        create_template(args["path"])
```
mikix, line 166: nit: If you just returned after this, you wouldn't need to modify the whole rest of this method.

nit: Honestly, this func would probably be more readable if you just did:
```
if create:
  do_create()
elif build:
  do_build()
elif export:
  do_export()
else:
  error out
```

And you could add a shared method like `make_builder(args)` to share that bit.

cumulus_library/cli.py
```diff
@@ -113,73 +113,126 @@ def export_all(self, study_dict: Dict):
...
+  - explict command line arguments
+  - cumulus environment variables (CUMULUS_LIBRARY_PROFILE, 
+    CUMULUS_LIBRARY_SCHEMA, CUMULUS_LIBRARY_S3, CUMULUS_LIBRARY_REGION)
+  - Normal boto profile order (AWS env vars, ~/.aws/credentials, ~/.aws/config)""",
+    )
+
+    studygen = parser.add_argument_group("Creating study templates")
+    studygen.add_argument(
+        "-c",
+        "--create",
```
mikix, line 219: nit: Can the user pass multiples of create/build/export? I think you can mark them as an exclusive group. But maybe that would conflict with other groupings. They might make more sense as subcommands, but not an important thought.

docs/creating-studies.md
```diff
@@ -11,33 +11,30 @@ nav_order: 3
...
+```bash
+cumulus-library -c -p ./path/to/your/study/dir
+```
+We'll create that folder if it doesn't already exist. We recommend you use a name
+relevant to your study (we'll use `my_study` forthis document). The folder name is
+the same thing you will use as a target with `cumulus_library` to run your study's
+queries.
 
-Once you've made a new study, the `manifest.toml` is the place you let cumulus
+Once you've made a new study, the `manifest.toml` fiile is the place you let cumulus
```
mikix, line 37: nit: typo

docs/creating-studies.md
mikix: I'm fine with that as a maintainer burder. Only risk is that we'd be making SQL changes to an already proven pile of SQL, and folks may be pretty reluctant at that point.

mikix (approved):

cumulus_library/cli.py
dogversioning: yeah that was kind of my thought? like, you :could: start futzing around in athena and want to make a manifest after the fact?

cumulus_library/cli.py
dogversioning: well this wass more 'did you even supply one'? so that way i can rely on list typing for addition. though i guess i could make the default `[]` instead of None and not worry about it?

cumulus_library/cli.py
dogversioning: oh that was from when this was :much: grosser, and i meant to trim it - i removed it and it is now not complaining.

cumulus_library/cli.py
dogversioning: I explicitly do so i don't have to worry about checking if i'm in the build/export group of functions. I don't need this for testing create, and don't want to worry about forcing some create logic to output something that adds nicely to the builder.

cumulus_library/cli.py
mikix: There is no default right now, you have to supply the argument. And you type hinted it as a List, not Optional[List], so the hinting suggests you need to pass a real list in, no None allowed.

cumulus_library/cli.py
mikix: Yeah I had missed that you only create the builder in this branch of the code.

cumulus_library/cli.py
dogversioning: makes sense for create, though it hurts my brain.

for build/export, i want to allow a user to do both of those in one go - i think that's a pretty common use case?

cumulus_library/cli.py
mikix: Hmm, does our linting config turn off warning about unnecessary lint disables? That's a nice one.

cumulus_library/cli.py
dogversioning: see above - currently they can.

cumulus_library/cli.py
mikix: Whatever works for you here, I was just throwing out suggestions.

It bumped my code senses that adding an extra mode here would cause a giant indentation change -- often a sign that code isn't maybe organized in the same way you are trying to use it. (not always true, just saying)

cumulus_library/cli.py
dogversioning: `useless-suppression` is in the disabled list.

mikix: Why, Google. Such a helpful mistake-catcher, and keeps you from backsliding. I don't feel strongly enough to advocate for a delta, but I wouldn't hate it.

cumulus_library/cli.py
dogversioning: my inclination for the time being is to leave it for now and if this gets more complex/that linting warning comes back, move things off into submethods then.

dogversioning: oth the other thing i was trying to dodge here - i don't want to create a builder if it's not needed, just so i don't throw out an athena error trying to create a cursor if a user isn't to that point yet?

PR #17, opened 2023-05-11
initial release prep

dogversioning:
### Description
- Some changes to the pyproject.toml for release prep
  - some hooks for pypi links
- manifest for getting the sqlfluff config in the distribution

Deployed to test at https://test.pypi.org/manage/project/cumulus-library-core/release/0.1.0a1/

### Checklist


pyproject.toml
```diff
@@ -18,30 +18,36 @@ dependencies = [
...
 test = [
     "pytest",
 ]
+
+[project.urls]
+Documentation = "https://docs.smarthealthit.org/cumulus/"
+Source = "https://github.com/smart-on-fhir/cumulus-library-core"
+
+[project.scripts]
+cumulus-library = "cumulus_library.cli:main"
```
mikix, line 46: We talked about this a little on slack, but just putting it here for posterity - I think this needs more discussion. It's really going against convention to have a pip package install the binary `cumulus-library` and the module `cumulus_library` but not have the package name be `cumulus-library`.

That's bad for discoverability, and it's bad namespace-management conventions. I know we talked about having other studies be like `cumulus-library-<study>` and that's fine. But the core package should be, IMO, `cumulus-library`.

Let's chat more about this.

dogversioning: ok, the project name is now reverted to `cumulus-library`.

mikix (approved):
Love it! Choo choo all aboard the open source Library train!

PR #23, opened 2023-05-26
Allowlist for pip studies

dogversioning:
### Description
- Adds an allowlist for autojoining select studies from site-packages
  - New docs for exposing this to users. 
- Added error condition reporting if study is not found
  - which exposed some tests that weren't really doing much, whoops.

### Checklist

### DIFF
--- a/cumulus_library/cli.py
+++ b/cumulus_library/cli.py
@@ -2,8 +2,10 @@
 """Utility for building/retrieving data views in AWS Athena"""
 
 import argparse
+import json
 import os
 import sys
+import sysconfig
 
 from pathlib import Path, PosixPath
 from typing import Dict, List, Optional
@@ -138,12 +140,26 @@ def create_template(path: str) -> None:
 
 
 def get_study_dict(alt_dir_paths: List) -> Optional[Dict[str, PosixPath]]:
-    """Convenience function for getting directories in ./studies/
+    """Gets valid study targets from ./studies/, and any pip installed studies
 
     :returns: A list of pathlib.PosixPath objects
     """
     manifest_studies = {}
-    paths = [Path(Path(__file__).resolve().parents[0], "studies")]
+    cli_path = Path(__file__).resolve().parents[0]
+
+    # first, we'll get any installed public studies
+    with open(Path(cli_path, "./module_allowlist.json"), "r") as study_allowlist_json:
+        study_allowlist = json.load(study_allowlist_json)
+    study_allowlist.pop("__desc__")
+    site_packages_dir = sysconfig.get_path("purelib")
+    for study, subdir in study_allowlist.items():
+        study_path = Path(site_packages_dir, subdir)
+        if study_path.exists():
+            manifest_studies[study] = study_path
+
+    # then we'll get all studies inside the project directory, followed by
+    # any user supplied paths last. These take precedence.
+    paths = [Path(cli_path, "studies")]
     if alt_dir_paths is not None:
         paths = paths + alt_dir_paths
     for parent_path in paths:
@@ -168,7 +184,7 @@ def run_cli(args: Dict):
     if args["create"]:
         create_template(args["path"])
 
-    if args["build"] or args["export"]:
+    elif args["build"] or args["export"]:
         builder = StudyBuilder(
             args["region"],
             args["workgroup"],
@@ -179,6 +195,15 @@ def run_cli(args: Dict):
             builder.verbose = True
 
         study_dict = get_study_dict(args["study_dir"])
+        if args["target"]:
+            for target in args["target"]:
+                if target not in study_dict:
+                    sys.exit(
+                        f"{target} was not found in available studies: "
+                        f"{list(study_dict.keys())}.\n\n"
+                        "If you are trying to run a custom study, make sure "
+                        "you include `-s path/to/study/dir` as an arugment."
+                    )
         # here we invoke the cursor once to confirm valid connections
         builder.cursor.execute("show tables")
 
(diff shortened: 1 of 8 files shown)


cumulus_library/cli.py
```diff
@@ -2,8 +2,10 @@
 """Utility for building/retrieving data views in AWS Athena"""
 
 import argparse
+import json
 import os
 import sys
+import sysconfig
```
mikix, line 8: lol always scary to see this import :smile:

cumulus_library/cli.py
```diff
@@ -179,6 +195,15 @@ def run_cli(args: Dict):
             builder.verbose = True
 
         study_dict = get_study_dict(args["study_dir"])
+        if args["target"]:
+            for target in args["target"]:
+                if target not in study_dict:
```
mikix, line 200: This feels awkward, but I'm not sure I love some alternatives I thought of (do a `set` difference, maybe define `target = args["target"] or []` so you don't need the `if` AND `for`....)

pyproject.toml
```diff
@@ -21,6 +21,7 @@ dependencies = [
 description = "Clinical study SQL generation for data derived from bulk FHIR"
 readme = "README.md"
 license = { text="Apache License 2.0" }
+homepage = "https://smarthealthit.org/cumulus-a-universal-sidecar-for-a-smart-learning-healthcare-system/"
```
mikix, line 24: I think this needs to be in its own url section?

mikix (approved):

cumulus_library/cli.py
dogversioning: i'm open to suggestions, i'd love something higher level.

cumulus_library/cli.py
```diff
@@ -138,12 +140,26 @@ def create_template(path: str) -> None:
...
     :returns: A list of pathlib.PosixPath objects
     """
     manifest_studies = {}
-    paths = [Path(Path(__file__).resolve().parents[0], "studies")]
+    cli_path = Path(__file__).resolve().parents[0]
+
+    # first, we'll get any installed public studies
+    with open(Path(cli_path, "./module_allowlist.json"), "r") as study_allowlist_json:
+        study_allowlist = json.load(study_allowlist_json)
+    study_allowlist.pop("__desc__")
```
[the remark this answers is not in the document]
dogversioning: oh yes that's much cleaner, i'll make that change.

cumulus_library/cli.py
```diff
@@ -138,12 +140,26 @@ def create_template(path: str) -> None:
...
+    site_packages_dir = sysconfig.get_path("purelib")
```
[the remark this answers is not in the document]
dogversioning: ok so - lemme get to the second part first. right now, for ease of end users, even though we're installing studies with pip, they aren't really pure python modules at the moment, since we're using the manifest.toml as a starting point to disambiguate out. I :could: add a wrapper at the base level to expose the toml infrastructure, but i kind of like keeping it out for two reasons:
- changes to the loader infrastructure only happen in the core repo
- one less bit of plumbing exposed to the 'analyst' user role

so, with that said: i'm targeting purelib based on the [list of python installation paths(https://docs.python.org/3/library/sysconfig.html#installation-paths). This path picks up the correct locationbased on whether i'm invoking with system python or a virtual env, so i believe it would work with package manager pythons as well.

since this path is down in the guts of the system, I'm somewhat loathe to expose it to the analyst class of users as an env var, though I would be ok doing so if we provided a command for them to run to get the contents of that var. 

if you buy this approach, i'll document this more in code, though please feel free to push back.

cumulus_library/cli.py
dogversioning: i could move it into a function to hide the awardness from this primary entrypoint function?

pyproject.toml
dogversioning: this one has some legacy support but i can move it into the URL dict.

cumulus_library/cli.py
mikix: Naw yeah, for the way you're doing this, this is appropriate. Just one of those imports that's like "we're about to get crazy, buckle up"

pyproject.toml
mikix: Oh really that's nice

dogversioning: i've actually got in the right place as well so i'm just removing this one.

cumulus_library/cli.py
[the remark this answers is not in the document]
mikix: > purelib

Sure, that works until someone has some *serious* study that ships some compiled code. Or we become popular and debian packages us, so the analyst apt-installs us, but pip-installs a specific study they want. Or the analyst installed the study into a venv that also sees their user-pip location. Just saying Python is specifically designed to avoid making assumptions about where imports live - the import flow chart is complicated.

You talk about not wanting wrappers - but isn't that what the `cumulus_library_covid` package is? Like it's a pip-installable wrapper package? I would have thought the right side of the dict could just be `cumulus_library_covid`, you import it, and then grab the toml dir -- either as a well-known folder under the wrapper package, or a defined variable in the wrapper or something, or a second component of the dictionary value.

> I'm somewhat loathe to expose it to the analyst class of users as an env var

Yeah totes agree. My mention of `PYTHONPATH` was just one example of the ways that python path importing is not a simple "one directory" thing.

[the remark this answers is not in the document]
dogversioning: > I would have thought the right side of the dict could just be cumulus_library_covid, you import it, and then grab the toml dir -- either as a well-known folder under the wrapper package, or a defined variable in the wrapper or something, or a second component of the dictionary value.

honestly, i hadn't thought of this approach - lemme give this a go.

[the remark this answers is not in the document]
dogversioning: although - this does create a nested 'import this thing if it exists' headache - right now these things are agnostic to importing.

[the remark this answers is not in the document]
mikix: > right now these things are agnostic to importing.

Says the guy inspecting `purelib` :smile:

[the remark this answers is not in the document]
mikix: Sure yeah

PR #27, opened 2023-05-31
Undo linter overcorrection

dogversioning:
### Description
This PR makes the following chage:
- SqlFluff trims out names of unnests that are used in a where clause, but not in the query itself. This undoes that correction and adds an ignore which should prevent it from happening in the future.

### Checklist


dogversioning:
for posterity - this issue is already in the SqlFluff queue: https://github.com/sqlfluff/sqlfluff/issues/4802

PR #39, opened 2023-06-09
37 observation vital signs

comorbidity:
### Description
"core observations" previously was scoped to only laboratory. 
This PR attempts to both 
* define VITAL SIGNS category 
* define "core observation" so that other cateogories can be added later more easily.

### DIFF
--- a/cumulus_library/studies/core/observation.sql
+++ b/cumulus_library/studies/core/observation.sql
@@ -1,0 +1,40 @@
+drop table if exists core__observation;
+
+create TABLE core__observation as
+with temp_observation as (
+    SELECT
+          category
+        , component
+        , status
+        , code_row as obs_code
+        , code
+        , interpretation
+        , referenceRange
+        , valueQuantity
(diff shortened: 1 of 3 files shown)


cumulus_library/studies/core/observation.sql
```diff
@@ -0,0 +1,40 @@
+drop table if exists core__observation;
```
dogversioning, line 1: `drop` statements are no longer required to be explicitly defined - all tables/views beginning with the study prefix will be discovered directly from the DB and auto dropped.

cumulus_library/studies/core/observation.sql
dogversioning, line 1: we should have a side conversation about linting/test requirements and how you want to handle them - i'm happy to have this be an 'lobbed over the wall' approach if that's what makes the most sense, but we should agree on that before i start appending commits to PRs by default.

dogversioning (approved):

PR #45, opened 2023-06-15
43 fhir valueset support

comorbidity:
provided FHIR Valuset template for generating SQL to use in each "study"

fixed wrong import of "library", now "cumulus_library" 

increased version number to 0.2.2


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "cumulus-library"
-version = "0.2.1"
+version = "0.2.2"
```
dogversioning, line 3: this is fine this time, but in the future we may want to do these increments as standalone commits just to avoid merge issues - what i'm working on over on the CLI side is going to become a 0.3.0 release, for example.

cumulus_library/template_sql/parse_valueset.py
```diff
@@ -0,0 +1,45 @@
+import os
+from typing import List
+from cumulus_library.helper import load_json
+from fhirclient.models.coding import Coding
```
dogversioning, line 4: we don't have this as a library dependency right now - i'll add it as part of test+lint

cumulus_library/template_sql/parse_valueset.py
```diff
@@ -0,0 +1,45 @@
...
+    :param view_name: like study__define_type
+    :param concept_list: list of concepts to include in definition
+    :return: sql statement to execute
+    """
+    header = f"create or replace view {view_name} as select * from (values"
+    footer = ") AS t (system, code, display) ;"
+    content = list()
+    for concept in concept_list:
+        content.append(f"('{concept.system}', '{concept.code}', '{concept.display}')")
+    content = '\n,'.join(content)
```
dogversioning, line 40: im going to convert this to jinja so it can be auto validated.

dogversioning (commented):
Lemme temporarily lock this for changes while doing the test + lint pass

dogversioning (approved):
ok, @comorbidity - take a look and make sure this is sensible to you. I moved the parser to a designated parser directory, and made a new single entrypoint in template_sql/templates for create view queries, which hopefully decouples things a bit.

Issue #20, opened 2023-05-23
Support grabbing AWS credentials from environment variables


mikix:
Rather than only from a ~/.aws/credentials profile.

(Variables like `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, and `AWS_SESSION_TOKEN`)

PR #48, opened 2023-06-21
Read AWS creds from profile, central error lib

dogversioning:
Adds support for global AWS env settings #20 
Moves existing custom error into cumulus_library.errors #3 

### Checklist

### DIFF
--- a/cumulus_library/cli.py
+++ b/cumulus_library/cli.py
@@ -47,18 +47,30 @@ class StudyBuilder:
     def __init__(  # pylint: disable=too-many-arguments
         self, region: str, workgroup: str, profile: str, schema: str
     ):
+        connect_kwargs = {}
+        for aws_env_name in [
+            "AWS_ACCESS_KEY_ID",
+            "AWS_SECRET_ACCESS_KEY",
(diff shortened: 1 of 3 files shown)


cumulus_library/cli.py
```diff
@@ -47,18 +47,30 @@ class StudyBuilder:
...
         self, region: str, workgroup: str, profile: str, schema: str
     ):
+        connect_kwargs = {}
+        for aws_env_name in [
+            "AWS_ACCESS_KEY_ID",
+            "AWS_SECRET_ACCESS_KEY",
+            "AWS_SESSION_TOKEN",
+        ]:
+            if aws_env_val := os.environ.get(aws_env_name):
+                connect_kwargs[aws_env_name.lower()] = aws_env_val
```
mikix, line 57: pyathena doesn't grab those variables itself as fallbacks? I would have guessed that it would, but maybe it considers itself lower-level than that.

mikix (approved):

cumulus_library/cli.py
dogversioning: yeah it doesn't, i've just verified. 🤷

Issue #49, opened 2023-06-21
Technical Debt: replace views with tables


dogversioning:
since we need to rerun study generation due to the presence of tables, we may as well make everything a table for performance reasons.

Issue #38, opened 2023-06-09
Core Encounter timing out in Athena


comorbidity:
Workaround: use a table instead of view, drop postalcode which isn't in use. 

`drop table core__count_encounter_month;

create table core__count_encounter_month as
with powerset as
(
    select distinct
          count(distinct P.subject_id) as cnt_subject,
          count(distinct E.encounter_id) as cnt_encounter
        , enc_class.code as enc_class_code
        , start_month
        , age_at_visit
        , P.gender
        , P.race_display
    from core__encounter E, core__patient P
    where E.subject_ref = P.subject_ref
    group by CUBE (enc_class, start_month, age_at_visit, gender, race_display)
)
select  distinct cnt_encounter as cnt
        , enc_class_code, start_month
        , age_at_visit, gender, race_display
from    powerset
where    cnt_subject >= 10
order by start_month asc, enc_class_code, age_at_visit asc
;`

comorbidity:
first reported by Regenstrief (david)

dogversioning:
Resolved via #50

PR #50, opened 2023-06-21
replace views with tables

dogversioning:
- Everything is now a table #49 #38 
- Caught one dangling usage of the old progress bar in the vocab generator
- Incremented version - i'll cut a bugfix release after this to make things simpler for Regienstrief

### Checklist

### DIFF
--- a/cumulus_library/studies/core/encounter.sql
+++ b/cumulus_library/studies/core/encounter.sql
@@ -36,7 +36,7 @@ WHERE
     e.subject_ref = p.subject_ref
     AND start_date BETWEEN date('2016-06-01') AND current_date;
 
-CREATE OR REPLACE VIEW core__join_encounter_patient AS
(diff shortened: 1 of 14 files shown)


cumulus_library/studies/core/encounter.sql
```diff
@@ -68,8 +68,7 @@ WITH powerset AS (
         ce.age_at_visit,
         cp.gender,
         cp.race_display,
-        cp.ethnicity_display,
-        cp.postalcode3
+        cp.ethnicity_display
```
mikix, line 71: Just confirming this was intentional

mikix (approved):

cumulus_library/studies/core/encounter.sql
dogversioning: yep, sorry -  it was part of the request on #38

PR #54, opened 2023-06-22
Add --version CLI argument

dogversioning:
This PR adds a --version argument to the CLI for coherence checking.

### Checklist

### DIFF
--- a/cumulus_library/__init__.py
+++ b/cumulus_library/__init__.py
@@ -1,0 +1,5 @@
+from importlib import metadata
+
+__version__ = metadata.version(__package__)
+
+del metadata
--- a/cumulus_library/__init__.py
+++ b/cumulus_library/__init__.py
@@ -1,5 +1,1 @@
-from importlib import metadata
-
-__version__ = metadata.version(__package__)
-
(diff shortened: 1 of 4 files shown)


cumulus_library/__init__.py
```diff
@@ -0,0 +1,5 @@
+from importlib import metadata
+
+__version__ = metadata.version(__package__)
```
mikix, line 3: Taste may vary, but I've found it more convenient to just define `__version__ = "0.3.2"` here in this file and set `dynamic = ["version"]` in pyproject.toml, and it will then read it from this file. (see ctakesclient or cumulus-etl for examples).

mikix (approved):

cumulus_library/cli.py
```diff
@@ -3,6 +3,7 @@
 
 import json
 import os
+import pkg_resources
```
mikix, line 6: :scissors:

cumulus_library/__init__.py
dogversioning: ok, i came around on this.

PR #62, opened 2023-06-30
performance tuning for codeableConcepts

dogversioning:
This improves the codableConcept query for large datasets by moving the ordering step to the partitioned subtable, so it can be better map/reduced.

### Checklist
- [X ] Consider if documentation (like in `docs/`) needs to be updated


mikix (approved):
Sure - we talked offline about this and it's been tested

PR #69, opened 2023-07-05
Better procecure encounter support

comorbidity:
core__procedure now compiles with limitations
* procedure "codes" are empty 70% of the time 
* procedure "performed on" dates are not true dates 

core__encounter_type now defines extra type metadata beyond "encounter class"
* encounter type 
* encounter service
* encounter priority 
* encounter reasoncode

Note that reasoncode is typically free text and should not be "counted".


cumulus_library/studies/core/count_core.py
dogversioning, line 1: Since this repo is public, I'm going to suggest not including this file in this PR, for a couple of reasons:
- we know we need to move the underlying logic this to a central location (we didn't have a ticket that i could see, so I added #70)
- we know this does not output sql compliant with the [mozilla sql style guide](https://docs.telemetry.mozilla.org/concepts/sql_style.html) - making it a bad starting point if someone else is going to come along and pick it up.
- We don't have a mechanism of validating the template syntax is correct without manually running the table, which has been an error source in the past
- it's generating views instead of tables, so it may not be performant on larger data sets.

in the short term until we can get to technical debt, it's probably easier to work with these tables as sql artifacts, rather than regenerating them from this script.

cumulus_library/studies/core/count_core.sql
```diff
@@ -0,0 +1,151 @@
...
+    )
+    select
+          cnt_encounter  as cnt 
+        , start_month, enc_class_display, age_at_visit, gender, race_display, ethnicity_display
+    from powerset 
+    WHERE cnt_subject >= 10 
+    ORDER BY cnt desc;
+
+-- ###########################################################
+CREATE or replace VIEW core__count_encounter_type AS
```
dogversioning, line 59: checking - are these encounter counts intentionally using a different binning minimum?

cumulus_library/studies/core/encounter.sql
```diff
@@ -30,87 +36,61 @@ SELECT DISTINCT
...
+    E.enc_class_display,
+    COALESCE(join_enc_type.as_coding.system, '?')   as enc_type_system,
+    COALESCE(join_enc_type.as_coding.code, '?')     as enc_type_code,
+    COALESCE(join_enc_type.as_coding.display, '?')  as enc_type_display,
+    COALESCE(join_service.as_coding.system, '?')    as enc_service_system,
+    COALESCE(join_service.as_coding.code, '?')      as enc_service_code,
+    COALESCE(join_service.as_coding.display, '?')   as enc_service_display,
+    COALESCE(join_priority.as_coding.system, '?')   as enc_priority_system,
+    COALESCE(join_priority.as_coding.code, '?')     as enc_priority_code,
+    COALESCE(lower(join_priority.as_coding.display), '?')  as enc_priority_display,
```
dogversioning, line 77: I'm leery of not just using nulls for stuff like this - it's the expected behavior, so it adds an unusual condition if you're trying to check for presence/absence later in a downstream query against this table.

dogversioning (commented):
We'll need to do a linting/unit test pass on this before allowing it in.

cumulus_library/studies/core/count_core.sql
comorbidity: yes, because encounters are for everyone and we dont want to overwhelm with small bin sizes. This also allows us to run "monthly" counts which is impossible (athena query dies) when we set the min size to 10

cumulus_library/studies/core/count_core.sql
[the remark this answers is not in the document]
comorbidity: agree to one pattern and stick to it. I didn't have time to convert all the other views.

cumulus_library/studies/core/encounter.sql
comorbidity: it is intentional. We only sometimes get Encounter type, service, priority.

cumulus_library/studies/core/fhir_define.sql
```diff
@@ -64,14 +64,17 @@ SELECT * FROM
             'PostalCode',
             'http://hl7.org/fhir/datatypes-definitions.html#Address.postalCode'
         ),
+        (
+            'PatientClass', 'http://terminology.hl7.org/CodeSystem/v2-0004')
```
dogversioning, line 68: missing a comma here.

cumulus_library/studies/core/count_core.py
dogversioning: i'm leaving this one alone for time reasons but let the record show i still think it should be excluded.

cumulus_library/studies/core/count_core.sql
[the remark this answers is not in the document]
dogversioning: deferred - #70

cumulus_library/studies/core/encounter.sql
dogversioning: was resolved via #75

dogversioning (approved):

Issue #66, opened 2023-07-03
Suggestion: Encounter.class.display will be more useful than code


James-R-Jones:
Epic uses an internal code system for encounter.class that has numbers as codes.

Suggest updating cumulus_library/studies/core/encounter.sql to use `display` rather than `code` on line 52

comorbidity:
Related to additional Encounter elements #31
https://github.com/smart-on-fhir/cumulus-library/issues/31

comorbidity:
![image](https://github.com/smart-on-fhir/cumulus-library/assets/96439283/5774d247-f367-41ab-9b84-a767cc671e64)

PR #75, opened 2023-07-07
Per Jamie's good suggestion, replace Encounter class code with displa…

comorbidity:
Suggestion: **Encounter.class.display** will be more useful than **code** #66
https://github.com/smart-on-fhir/cumulus-library/issues/66

This is a PATCH to force the use of Encounter.class.display and 'None" instead of '?' everywhere in library CORE


dogversioning (approved):
We need a ticket for resolving this naming discrepancy post 7-11.

PR #79, opened 2023-07-14
Table builder refactor

dogversioning:
- Refactors the generic python entrypoint in the manifest to be targeted at a specific kind of table building task
- splits out table building tasks into targeted files
- replaces the deprecated module loading with the best practice recipe from [the importlib docs](https://docs.python.org/3/library/importlib.html#importing-a-source-file-directly)
  - removes cap on python version as a result

### Checklist
- [ ] Update template repo if there are changes to study configuration <-- this needs to be done for :all: study repos. will do this after.


cumulus_library/base_table_builder.py
```diff
@@ -0,0 +1,49 @@
...
+        :param verbose: toggle for verbose output mode
+        """
+        with get_progress_bar(disable=verbose) as progress:
+            task = progress.add_task(
+                self.display_text,
+                total=len(self.queries),
+                visible=not verbose,
+            )
+            self.prepare_queries(cursor, schema)
+            for query in self.queries:
```
mikix, line 47: nit shower thought: maybe `prepare_queries` could return an array of queries, so you could avoid having a shared state variable between super and sub classes, and the lifecycle of the queries might be clearer

cumulus_library/base_table_builder.py
```diff
@@ -0,0 +1,49 @@
...
+class BaseTableBuilder(ABC):
+    """Generic base class for python table builders.
+
+    To use a table builder, extend this class exactly once in a new module.
+    See ./studies/vocab or ./studies/core for example usage.
+    """
+
+    def __init__(self):
+        self.queries = []
+        self.display_text = "Building custom tables..."
```
mikix, line 17: nit shower thought: you could make this a class variable, so that subclasses don't have to define `__init__` at all, saving some boilerplate. They could still define it per-instance if they really wanted, but I bet the 99% case is a static string.

cumulus_library/studies/core/builder_condition_codes.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def prepare_queries(self, cursor: object, schema: str):
+        """Constructs queries related to condition codeableConcept
+
+        :param cursor: A database cursor object
+        :param schema: the schema/db name, matching the cursor
+
+        """
+        self.queries.append(
+            get_codeable_concept_denormalize_query(
```
mikix, line 22: so fresh, so clean

cumulus_library/__init__.py
```diff
@@ -1,2 +1,2 @@
 """Package metadata"""
-__version__ = "0.4.0"
+__version__ = "0.5.0"
```
mikix, line 2: You made this sound like a breaking change in slack - that the manifest changed in a way that would require study changes.

I'd think that would warrant a major version bump, but I don't know what your versioning policy is. Maybe it's not semantic versioning.

cumulus_library/studies/core/manifest.toml
```diff
@@ -1,8 +1,9 @@
 study_prefix = "core"
 
-[python_config]
+[table_builder_config]
```
mikix, line 3: :+1: good name change

cumulus_library/studies/template/manifest.toml
```diff
@@ -29,13 +29,10 @@ export_list = [
...
+# build tables, you can provide a list of files implementing BaseTableBuilder.
+# See vocab and core studies for examples of this pattern
+# [table_builder_config]
 # file_names = [
-#     "my_script.py",
+#     "my_table_builder.py",
 # ]
-
-# If you'd like another language available, please reach out - we can likely 
-# accomadate, but we'd need to understand the use cases a little better.
```
mikix, line 41: :+1:

cumulus_library/base_table_builder.py
```diff
@@ -0,0 +1,49 @@
...
+        When completed, prepare_queries should populate self.queries with sql
+        statements to execute. This array will the be read by run queries.
+
+        :param cursor: A PEP-249 compatible cursor
+        :param schema: A schema name
+        :param verbose: toggle for verbose output mode
+        """
+        raise NotImplementedError
+
+    @final
```
mikix, line 32: I didn't know this existed, that's nice to have.

cumulus_library/base_table_builder.py
```diff
@@ -0,0 +1,49 @@
...
+    def execute_queries(self, cursor: object, schema: str, verbose: bool):
+        """Executes queries set up by a prepare_queries call
+
+        :param cursor: A PEP-249 compatible cursor
+        :param schema: A schema name
+        :param verbose: toggle for verbose output mode
+        """
+        with get_progress_bar(disable=verbose) as progress:
+            task = progress.add_task(
```
mikix, line 41: This is a behavior change, right? Previously there was one progress bar for the whole study? Is this desirable, or just a compromise with how the refactor went? Because I could imagine you having code that sat above all the study's table builders, easily enough.

mikix (approved):

cumulus_library/__init__.py
dogversioning: if we had a 1.X release, I would make this a 2.X. Since we're still in 0.X, I think [it's ok](https://semver.org/#spec-item-4)?

cumulus_library/base_table_builder.py
dogversioning: in my head, if you had one builder referencing a bunch of submodules generating queries, it would be slightly easier to dump things into a state. this is a weakly held opinion.

cumulus_library/base_table_builder.py
mikix: Naw, my comment was less about best practices and more "I think this will be easier for users" but just a nit

cumulus_library/__init__.py
mikix: I mean sure, fine - the semantic versioning authorities give you an out because you have avoided bumping to a real version so far. But... you're not alpha software moving fast and loose - you have real partners using this for real things. The 0.x is increasingly a lie.

cumulus_library/base_table_builder.py
mikix: I mean, ultimately who cares whether there's one or two progress bars - I wouldn't worry too much about the effect on the user in this case. Your approach is fine

cumulus_library/__init__.py
mikix: (and more to the point - the goal of semantic versioning is to be kind to consumers by warning about big changes - why not just do that?)

dogversioning: per our long in person discussion we agreed to do this, and it does not imply that someone should be writing a press release.

PR #81, opened 2023-07-19
CodeableConcept generic handling

dogversioning:
- Added templates for handling codeableConcepts
- Added core builder for extracting codeableConcepts from encounters
- Updated core tables to use new concept table
- Alphasorting of some lists that have gotten very long

### Checklist
- [ ] Update template repo if there are changes to study configuration <- todo on release


cumulus_library/.sqlfluff
```diff
@@ -3,7 +3,7 @@ templater = jinja
 dialect = athena
 sql_file_exts = .sql,.sql.jinja
 # this rule overfires on athena nested arrays
-exclude_rules=references.from,structure.column_order
+exclude_rules=references.from,structure.column_order,aliasing.unused
```
dogversioning, line 6: The AL05 rule is so over-agressive that, rather than continuing to ask users to remember to apply QA rules for it, I'm disabling it for now. This means that a query like this will no longer trigger a correction:

```
SELECT t2.row2
FROM table,
unnest (table.array) as t1 (row1) <--- this would trigger this rule even though it's valid
unnest (t1.row1.array) as t2 (row2)
```

I'm watching a couple issues on the sqlfluff repository related to this.

cumulus_library/.sqlfluff
```diff
@@ -16,21 +16,25 @@ ignore_comment_lines = true
...
+prefix = Test
 schema_name = test_schema
+source_table = source_table
+source_id = source_id
+table_cols = ["a","b"]
 table_name = test_table
+target_col_prefix = prefix
+target_table = target_table
+unnests = [{"source col": "g", "table_alias": "i", "row_alias":"j"}, {"source col": "k", "table_alias": "l", "row_alias":"m"},]
+view_cols = ["c","d"]
```
dogversioning, line 37: cc_columns and unnests are the only new sqlfluff aliases here - otherwise, this is just trying to apply order to chaos.

cumulus_library/base_table_builder.py
```diff
@@ -38,13 +38,13 @@ def execute_queries(self, cursor: object, schema: str, verbose: bool):
         :param schema: A schema name
         :param verbose: toggle for verbose output mode
         """
+        self.prepare_queries(cursor, schema)
```
dogversioning, line 41: this was moved due to needing to create an extra progress bar inside a builder, since only one progress object can be active at a time (though it can have multiple tasks).

cumulus_library/studies/core/builder_condition_codableconcept.py
```diff
@@ -0,0 +1,31 @@
...
+        """Constructs queries related to condition codeableConcept
+
+        :param cursor: A database cursor object
+        :param schema: the schema/db name, matching the cursor
+
+        """
+        config = CodeableConceptConfig(
+            source_table="condition",
+            source_id="id",
+            cc_columns=[{"name": "code", "is_array": False}],
```
dogversioning, line 23: these are the only two lines changed in this file, which is renamed from builder_condition_codes.py. not sure why git isn't properly detecting this one.

cumulus_library/studies/core/count_core.py
dogversioning, line 1: I did not run this - this is more about documenting the expected state for when we move these to a count table generic generator.

cumulus_library/template_sql/codeable_concept_denormalize.sql.jinja
```diff
@@ -1,35 +1,47 @@
...
+            UNNEST(s.{{ cc_column.name }}) AS cc (cc_row),
+            UNNEST(cc.cc_row.coding) AS u (codeable_concept)
+        {% else %}
+            UNNEST(s.{{ cc_column.name }}.coding) AS u (codeable_concept)
+        {%- endif %}
         WHERE
             u.codeable_concept.system = '{{ system }}'
     ), --noqa: LT07
     {%- endfor %}
+    {% endfor %}
```
mikix, line 25: Why does the opening `{%` sometimes have a hyphen and sometimes not?

cumulus_library/studies/core/count_core.py
```diff
@@ -58,29 +58,11 @@ def _count_encounter_type(view_name, cols, duration):
 
 def count_encounter_type(duration=None):
     cols = [
-        "enc_class_display",
-        "enc_type_display",
-        "enc_service_display",
-        "enc_priority_display",
+        "codableconcept_display",
```
mikix, line 61: You frequently use the actual English word `codable` vs the (afaict) not-a-word `codeable` from the spec (codable is in Merriam Webster, but it still gets squiggle marks in my browser, so maybe they are both barely words). That's fine, just calling it out in case you were fat-fingering these cases. But I'd also defend your right to use a maybe-more-correct word in code and only use the spec word when you have to.

cumulus_library/studies/core/builder_encounter_coding.py
```diff
@@ -0,0 +1,123 @@
...
+                        cols_with_data.append(code_source)
+                else:
+                    progress.advance(task)
+            else:
+                progress.advance(task)
+                progress.advance(task)
+        return cols_with_data
+
+    def prepare_queries(self, cursor: object, schema: str):
+        """Constructs queries related to condition codeableConcept
```
mikix, line 90: copy/paste

cumulus_library/studies/core/builder_encounter_coding.py
```diff
@@ -0,0 +1,123 @@
...
+        :param cursor: A database cursor object
+        :param schema: the schema/db name, matching the cursor
+
+        """
+        code_sources = [
+            {"name": "type", "is_array": True},
+            {"name": "servicetype", "is_array": False},
+            {"name": "priority", "is_array": False},
+        ]
+        discovery_length = len(code_sources) * 3
```
mikix, line 101: I don't love this construction - a magic value 3 that requires this line to know a lot about the internals of a separate method.

What if you pass `total=None` when constructing the task, and when advancing, you can just call `progress.update(task, completed=2, total=3)` (which also lets you skip double-advance calls by just setting `completed` or `total` directly).

cumulus_library/studies/core/builder_encounter_coding.py
```diff
@@ -0,0 +1,123 @@
...
+class EncounterCodingBuilder(BaseTableBuilder):
+    display_text = "Creating encounter type code table..."
+
+    def _check_data_in_fields(self, code_sources: List[Dict], cursor, progress, task):
+        """checks if CodableConcept fields actually have data available"""
+
+        # TODO: consider moving to a utility library if we have another case like
+        # this one
+        cols_with_data = []
+        for code_source in code_sources:
```
mikix, line 23: This method contains some hairy nested logic. A newcomer might appreciate (a) a dollop of "what is happening" comments maybe for the separate execution steps and (b) a dollop of "why is this happening" maybe in the docstring, for like, why this method is relevant/useful in the broader context.

mikix (approved):

cumulus_library/studies/core/builder_encounter_coding.py
dogversioning: i mean, :i: hate it, so lemme try to be more descriptive here.

cumulus_library/studies/core/builder_encounter_coding.py
```diff
@@ -0,0 +1,123 @@
...
+    CodeableConceptConfig,
+    get_codeable_concept_denormalize_query,
+    get_is_table_not_empty_query,
+)
+
+
+class EncounterCodingBuilder(BaseTableBuilder):
+    display_text = "Creating encounter type code table..."
+
+    def _check_data_in_fields(self, code_sources: List[Dict], cursor, progress, task):
```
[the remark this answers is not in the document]
dogversioning: i had that thought but was like 'wait, is that 3.10?' - will fix.

cumulus_library/studies/core/builder_encounter_coding.py
dogversioning: it was more sensible (alebit less readable) before i pulled the ladder logic out into its own function.

but my thought was this:
- you will always need to run three queries to prove conclusively that data exists, without introducing intentional error catching for misfired queries, which I think we don't want to do because it masks bug behavior
- We want someone to be able to add/remove codings from this easily without thinking too hard about it.

I could do the `total=None` route, though I selfishly like an advancing one for communicating that this is a known length of stuff we're doing here. I haven't used this in this way before so i don't fully know what the UX looks like.

what if 3 was replaced by a semantically named constant?

cumulus_library/studies/core/count_core.py
dogversioning: i actually think i should use the spec name here, despite what my spellchecker is telling me to do otherwise, because an analytics user would look for it in the non-english way.

fun fact, since i am the kind of person to go look this sort of thing up for fun: while Miriam Webster does not thing 'codeable' is a word, the OED does recognize it as valid, so maybe this is a color/colour situation. Anyway, taking a lorry back to my flat.

cumulus_library/template_sql/codeable_concept_denormalize.sql.jinja
dogversioning: it's jinja's way of determing what to do about continuing whitespace -[ this SO post ](https://stackoverflow.com/questions/11813422/what-does-this-in-jinja2-template-engine-do)is the most parsable explanation i've seen.

cumulus_library/studies/core/builder_encounter_coding.py
```diff
@@ -0,0 +1,123 @@
+""" Module for generating condition codeableConcept table"""
```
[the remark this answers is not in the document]
dogversioning: thank you.

cumulus_library/template_sql/codeable_concept_denormalize.sql.jinja
mikix: Of course :roll_eyes:

cumulus_library/studies/core/count_core.py
mikix: Merriam Webster is my household's (of one, granted) official dictionary for Boggle. So, I don't know why anyone would bother looking at another authority.

cumulus_library/template_sql/codeable_concept_denormalize.sql.jinja
dogversioning: i don't like it either.

cumulus_library/studies/core/builder_encounter_coding.py
mikix: My beef is less with wanting a semantically named magic constant, and more with the separation of logic between two different methods. If the `* 3` were at the top of `_check_data_in_fields` (like maybe `progress.update(task, total=len()*3)`?), I'd be happy enough. I just want that method boundary to actually be a semantic boundary, and allow the method to be rewritten in the future.

I think if the `* 3` is close to the code that does something thrice and there's a little comment, I don't think the addition of a named constant would add much more.

mikix: Or heck, have the method own & make its own task?

dogversioning: Ok, i went ahead and pushed the task up into this method, so we've got a full boundary between 'check if data exists' and 'build a table with columns that have data'

cumulus_library/studies/core/builder_encounter_coding.py
comorbidity: Why would ICD9/10 have anything to do with an Encounter code? 
That isn't the coding system (namespace) of encounters 

        config = CodeableConceptConfig(
            source_table="encounter",
            source_id="id",
            cc_columns=cols_with_data,
            target_table="core__encounter_coding",
            code_systems=[
                "http://snomed.info/sct",
                "http://hl7.org/fhir/sid/icd-10-cm",
                "http://hl7.org/fhir/sid/icd-9-cm",
            ],
        )

comorbidity: Here is what we get from Cerner for Encounter CLASS system 

http://terminology.hl7.org/CodeSystem/v3-ActCode
http://terminology.hl7.org/CodeSystem/v2-0004

comorbidity: for **servicetype**

Cerner says 
http://snomed.info/sct

and really should be 
http://hl7.org/fhir/ValueSet/service-type
http://terminology.hl7.org/CodeSystem/service-type

I'm not convinced though that we can expect to know the namespace (system) URL though.

comorbidity: http://hl7.org/fhir/R4/valueset-service-type.html

comorbidity: standard for **PRIORITY** is here, also not ICD9/10 

https://terminology.hl7.org/1.0.0//ValueSet-v3-ActPriority.html

cumulus_library/studies/core/count_core.py
comorbidity: is codeableconcept_display intended to have all 4 parts? 

Encounter Type, Class, Service, Priority?

comorbidity: Opinion: count_core.py and count_core.sql shouldn't be committed in this PR, I think it will break.

cumulus_library/studies/core/encounter_type.sql
comorbidity, line 1: I might be missing something, sorry, but I think this is incorrect? 
I dont see how you are going to get 4 variables into one field 

Encounter Type, Class, Priority, Service

comorbidity: As per USCDIv1 spec 
http://hl7.org/fhir/us/core/STU4/StructureDefinition-us-core-encounter.html

Each Encounter must have:

1 A status
2 An classification such as inpatient, outpatient or emergency
3 An encounter type
4 A patient

If that means we have to drop support for Priority and Service, we can live with that to not hold up support for Condition code, which is more important.

comorbidity (changes requested):
Condition looks good 
Encounter is not going to work at present

comorbidity:
List of known SYSTEM for Encounter Fields 

The standard = USCDIv1 = STU4 = http://hl7.org/fhir/us/core/STU4/StructureDefinition-us-core-encounter.html

**CLASS** (http://hl7.org/fhir/R4/encounter-definitions.html#Encounter.classHistory.class)
http://terminology.hl7.org/ValueSet/v3-ActEncounterCode
http://terminology.hl7.org/CodeSystem/v3-ActCode
http://terminology.hl7.org/CodeSystem/v2-0004
2.16.840.1.113883.1.11.13955

**TYPE** (http://hl7.org/fhir/R4/encounter-definitions.html#Encounter.type)
http://hl7.org/fhir/ValueSet/encounter-type
http://terminology.hl7.org/CodeSystem/encounter-type
http://terminology.hl7.org/CodeSystem/v2-0004
2.16.840.1.113883.4.642.3.248

**SERVICE** (https://hl7.org/fhir/R4/encounter-definitions.html#Encounter.serviceType)
http://hl7.org/fhir/ValueSet/service-type
http://terminology.hl7.org/CodeSystem/service-type
2.16.840.1.113883.4.642.3.518

**PRIORITY** (https://hl7.org/fhir/R4/encounter-definitions.html#Encounter.priority)
http://terminology.hl7.org/ValueSet/v3-ActPriority
http://terminology.hl7.org/CodeSystem/v3-ActPriority

cumulus_library/studies/core/builder_encounter_coding.py
dogversioning: these systems are now being used in the codeableconcept generator

comorbidity (approved):
@dogversioning addressed my concerns about Encounter meta

PR #85, opened 2023-07-21
Soe sequence of events

comorbidity:
Created SOE Sequence Of Events because reasons 
1) linking FHIR resources based on FHIR **Resource date** "period" when encounter reference is not available 
2) QA/verification that FHIR resources are not missing links 

Currently there is only ONE file that does this and reports with weekly counts. 
Future work may include SQL templates and other refactoring, this was done in isolation to not break main branch.


cumulus_library/studies/core/soe_sequence_of_events.sql
```diff
@@ -0,0 +1,343 @@
...
+        c.category,
+        concat('Condition/', c.id) AS condition_ref
+    FROM condition AS c
+),
+cond as
+(
+    select distinct rawdata.*, category_row
+    from rawdata,
+        unnest(category) AS t1 (category_coding),
+        unnest(category_coding.coding) AS t2 (category_row)
```
dogversioning, line 60: This unnest means we need to go through the various investigations of 'how does this come back in epic vs cerner vs others' - we may have to move this statement into a table builder.

cumulus_library/studies/core/soe_sequence_of_events.sql
```diff
@@ -0,0 +1,343 @@
...
+            link_reference.encounter_ref as encounter_ref,
+            link_period.encounter_ref as period_ref,
+            -- link REF via Encounter.reference when provided, else link Encounter.period
+            case when link_reference.encounter_ref is not null
+            then link_reference.encounter_ref
+            else link_period.encounter_ref end as encounter_link,
+            -- link CLASS via Encounter.reference when provided, else link Encounter.period
+            case when link_reference.enc_class is not null
+            then link_reference.enc_class
+            else link_period.enc_class end as enc_class,
```
dogversioning, line 103: how are we going to tell which of these took place? does it matter?

dogversioning (changes requested):
A couple of general things:

- it would be nice to use unique names for the intermediate CTAS queries
- `SELECT table.*` should be avoided for unplanned consequences
- We'll have to do a SqlFluff pass, though that can be right at the end.

cumulus_library/template_sql/templates.py
```diff
@@ -288,6 +289,34 @@ def get_is_table_not_empty_query(
...
 
+def get_object_denormalize_query(
+    schema_name: str,
+    source_table: str,
+    source_id: str,
+    field: str,
+    field_config: dict,
+    target_table: str,
+):
+    """Generates a table by expanding a specified row element
```
mikix, line 300: nit: "get object denormalize query" and "expanding a specified row element" is a lot of jargon, which I can see makes total sense for someone embedded here. But as a sometimes reviewer, it gave me pause.

The wording is fine, but could maybe like a tiny example in the docstring be appropriate? It seems to do a tiny flatten of a target field, yeah?

cumulus_library/studies/core/builder_soe.py
```diff
@@ -0,0 +1,57 @@
...
+    def prepare_queries(self, cursor: object, schema: str):
+        """Constructs tables for SOE QA verification
+
+        :param cursor: A database cursor object
+        :param schema: the schema/db name, matching the cursor
+
+        """
+        table = "documentreference"
+        column = "context"
+        with get_progress_bar(transient=True) as progress:
```
mikix, line 23: reminder: did you want to squeeze in the time-elapsed instead of time-remaining change?

mikix (approved):
I did not really review the sql - but the rest looks good!

cumulus_library/studies/core/builder_soe.py
dogversioning: I want to leave that as its own thing rather that tie it to another feature.

cumulus_library/template_sql/templates.py
dogversioning: take a look at this update and see what you think.

mikix: Sure! :+1:

Issue #88, opened 2023-07-21
core__medication should unnest codeableconcept to be used by studies


comorbidity:
The **code** field appears differently on Cerner/Epic. 

Cerner shows struct without code or system

`desc core__medication`
id                  	string              	                    
resourcetype        	string              	                    
**code**                	struct<text:string> 	                    
form                	struct<coding:array<struct<code:string,display:string,system:string,userselected:boolean>>,text:string>	                    
ingredient          	array<struct<itemcodeableconcept:struct<coding:array<struct<code:string,display:string,system:string,userselected:boolean>>,text:string>,strength:struct<denominator:struct<code:string,system:string,unit:string,value:double>,numerator:struct<unit:string,value:double>>>>	                    

Epic core__medication is struct with coding and array of CodeableConcepts. 
Example discovered during Opioid study here: https://github.com/smart-on-fhir/cumulus-library-opioid/issues/7

comorbidity:
Note that the **id** field in the schema is for Medication, not MedicationRequest, so there is no possible linkage to Subject. 
See https://github.com/smart-on-fhir/cumulus-library/issues/87

dogversioning:
relates to #83

Issue #87, opened 2023-07-21
core__medication has no reference to subject (or encounter)


comorbidity:
MedicationRequest.subject.reference is a required field 
http://hl7.org/fhir/us/core/STU4/StructureDefinition-us-core-medicationrequest.html

At minimum, core__medication needs to provide a reference to Patient.

Issue #83, opened 2023-07-20
medicationrequest.medicationcodeableconcept NULL handling


dogversioning:
This field has the same absence issues as other codeableconcept fields in the EPIC implementation. We'll have to move this to programmatic detection to guarantee query execution.

PR #89, opened 2023-07-24
Added inline codeable concept, patient/encounter ids

dogversioning:
- Adds inline medication codes from `medication` to `core__medications` for cerner #87
- Refactors the structure of the `core__medications` table to include ids and be more coding centric #88 #83 
- Uses a slightly faster query for checking for valid DB credentials for large numbers of tables
- light whitespace modifications for is_table_not_empty.sql.jinja

### Checklist
- [ ] Update template repo if there are changes to study configuration


cumulus_library/studies/core/builder_core_medication.py
```diff
@@ -18,17 +18,50 @@ def _check_data_in_fields(self, cursor, schema: str):
...
+                        ],
+                    )
+                    cursor.execute(query)
+                    progress.advance(task)
+                    if cursor.fetchone() is not None:
+                        data_types["inline"] = True
+                else:
+                    progress.advance(task)
+            else:
+                progress.advance(task, advance=2)
```
dogversioning, line 62: so this is the technical debt I was discussing - this looks very similar to the code in builder_encounter_coding, but it's structurally different in one important way: it uses the table schema for determining if coding is an available field, rather than relying on table length. This is slightly safer for the case where we've got unusual 'data is not present' conditions.

My intent is to move this to some kind of template_sql/utils kind of location, and import it here and in the encounter builder - but in order to do that, i'll need to split apart the array vs non array handling, and write a similar version for array handling. I'm not doing this as part of this PR due to time reasons.

cumulus_library/studies/core/medication_request.sql
```diff
@@ -8,14 +8,14 @@ SELECT
     date_trunc('month', date(from_iso8601_timestamp(mr.authoredon)))
     AS authoredon_month,
     mr.category,
-    med_coding.code AS code,
-    med_coding.display AS display,
-    med_coding.system AS system, --noqa: RF04
+    cm.code AS medication_code,
+    cm.display AS medication_display,
+    cm.code_system AS medication_system,
```
dogversioning, line 13: @comorbidity I'm labelling these because i am assuming, at some point, we're also going to want to unnest mr.category, which is also a codeable concept, so we'll need to keep these distinct from each other

comorbidity: elsewhere these fields are called 
rx_code
rx_display
rx_system

cumulus_library/studies/core/medication_request.sql
```diff
@@ -25,9 +25,9 @@ WITH powerset AS (
         cmr.status,
         cmr.intent,
         cmr.authoredon_month,
-        cmr.display
+        cmr.medication_display AS display
```
comorbidity, line 28: "rx_display"

cumulus_library/studies/core/medication_request.sql
comorbidity: alternatively 
med_code 
med_display 
med_system

comorbidity (approved):
overall looks good. Preference to use the names "rx_code, rx_display, rx_system" since that is common elsewhere in cumulus. medication_code, medication_display, and medication_system is not used anywhere.

cumulus_library/studies/core/builder_core_medication.sql
```diff
@@ -0,0 +1,153 @@
+-- noqa: disable=all
+/*
+This is a reference output of the SQL generated by builder_core_medication.py 
+that is used by the core__encounter_type table. It is not invoked directly.
+*/
```
mikix, line 5: Good comment, but I have a further clarifying question - is this used at all by a production run? Like, _could_ this be in a `tests/` directory, or is not invoked directly, but is still used somehow?

cumulus_library/template_sql/core_medication.sql.jinja
```diff
@@ -1,105 +1,203 @@
...
+        medication_datasources.by_external_ref)
+    -%}
+    ,
+    {%- endif %}
+    {# This section is also Cerner specific. Sometimes medication codes are blank
+    and we can get :some: info from the contains field.
+
+    As of this writing, the code coming back via contained when code is not
+    present contains text only. In these cases, we'll cast the text
+    as a display element.
```
mikix, line 44: Oh is that right? I'd not noticed or looked for that. No codeable concepts inside the contained resources? Rude.

mikix (approved):

cumulus_library/studies/core/builder_core_medication.sql
dogversioning: This is more about documentation at @comorbidity 's request - the idea being that a researcher can get an idea what the builder is doing in the maximal case, without actually needing to understand the builder's logic. But it is never run in any form.

Issue #92, opened 2023-07-25
medicationrequest Medication Display may be null, needs "coalesce" 


comorbidity:
When counting on Display (which can be null) this can cause problems in Dashboard.

PR #93, opened 2023-07-25
fix for potentially null values rx_display

comorbidity:
Minor fix for potentially NULL rx_display 
https://github.com/smart-on-fhir/cumulus-library/issues/92


dogversioning (approved):
This is fine, but I do sort of feel like long term this is something we should expect the dashboard to gracefully handle.

PR #97, opened 2023-07-26
refactor codeableConcept handling to central location

dogversioning:
- codeableConcept detection now handled via template_sql/utils
- `--table-builder` CLI option allows running just one python file implementing BaseTableBuilder in a study
- some improvements to handling for data set edge cases

### Checklist


cumulus_library/studies/core/manifest.toml
```diff
@@ -2,7 +2,7 @@ study_prefix = "core"
 
 [table_builder_config]
 file_names = [
-    "builder_condition_codableconcept.py",
+    "builder_condition_codeableconcept.py",
```
dogversioning, line 5: whoops whoops whoops

cumulus_library/study_parser.py
```diff
@@ -210,6 +210,46 @@ def _execute_drop_queries(
             cursor.execute(drop_view_table)
             query_console_output(verbose, drop_view_table, progress, task)
 
+    def load_builder_from_file(self, filename) -> (BaseTableBuilder, object):
```
dogversioning, line 213: no logic changes here, just moved it out so it can be used for building all studies/building one studies

cumulus_library/template_sql/core_medication.sql.jinja
```diff
@@ -23,7 +23,11 @@ CREATE TABLE core__medication AS (
             t.r.code,
             t.r.display,
             t.r.system AS code_system,
+            {% if missing_userselected %}
+            false AS userselected
+            {% else %}
             t.r.userselected
+            {% endif %}
```
dogversioning, line 30: i think? this case is only present in the synthea dataset, but it doesn't hurt to check for (and, y'know, allows us to test against synthea data).

cumulus_library/template_sql/core_medication.sql.jinja
```diff
@@ -142,12 +150,16 @@ CREATE TABLE core__medication AS (
     external_medication AS (
         SELECT
             mmn.id,
-            mmn.encounter AS encounter_ref,
-            mmn.subject AS patient_ref,
+            mmn.encounter.reference AS encounter_ref,
+            mmn.subject.reference AS patient_ref,
```
dogversioning, line 154: this is now consistent with the other two subtables

cumulus_library/base_table_builder.py
```diff
@@ -32,14 +34,31 @@ def prepare_queries(self, cursor: object, schema: str):
...
         :param cursor: A PEP-249 compatible cursor
         :param schema: A schema name
         :param verbose: toggle for verbose output mode
+        :param drop_table: drops any tables found in prepared_queries results
         """
         self.prepare_queries(cursor, schema)
+        if drop_table:
+            table_names = []
+            for query in self.queries:
+                # Get the first non-whitespace word after create table
```
mikix, line 51: I'm sure this is the most reasonable box we could put ourselves into, but all this manual, regex-based SQL parsing gives me the heebie-jeebies. At least sqlfluff makes our own SQL fairly regular, right?

Maybe it's not worth the effort, but is there an actual SQL parser we could import & use to tokenize the string? That still leaves us with assuming some SQL structure, but it's one less assumption-about-input. (This is all a nit suggestion)

cumulus_library/cli.py
```diff
@@ -250,7 +262,12 @@ def run_cli(args: Dict):
                     builder.clean_and_build_all(study_dict)
                 else:
                     for target in args["target"]:
-                        builder.clean_and_build_study(study_dict[target])
+                        if args["table_builder"] is not None:
```
mikix, line 265: nit: This feels like it would read better and be more correct if you also handled the empty string case with a simple `if args["table_builder"]:`

cumulus_library/cli_parser.py
```diff
@@ -15,6 +15,18 @@ def add_target_argument(parser: argparse.ArgumentParser) -> None:
...
 
 
+def add_table_builder_argument(parser: argparse.ArgumentParser) -> None:
+    """Adds --table_builder arg to a subparser"""
+    parser.add_argument(
+        "--table-builder",
+        help=(
+            "Within a study, specify a specific table builder to run, skipping all "
+            "other files. Primarily intended for development/debugging. Will still "
+            "remove study tables."
```
mikix, line 25: nit: consider using `help=argparse.SUPPRESS` instead and just hiding this if you think this arg is more dangerous than helpful for n00bs.

If you keep it, is a "table builder" anything a researcher would know? I know that's the internal name for it, but this might be a good point to consider if there's a better end-user terminology. (it's a fine name, I have no complaints, but just my brain fired this question off when exposing something for the first time)

cumulus_library/study_parser.py
```diff
@@ -219,49 +259,25 @@ def run_table_builder(
...
+            table_builder, table_builder_module = self.load_builder_from_file(file)
+            # After runnning the executor code, we'll remove the builder
             # remove it so it doesn't interfere with the next python module to
             # execute, since the subclass would otherwise hang around.
-            table_builder_class = table_builder_subclasses[0]
-            table_builder = table_builder_class()
+
             table_builder.execute_queries(cursor, schema, verbose)
             del sys.modules[table_builder_module.__name__]
             del table_builder_module
```
mikix, line 270: This cleanup feels like it belongs to the moved function - and you'd want this same cleanup for `run_single_table_builder` in theory yeah? (if the program didn't just exit after - e.g unit tests at least). Is there a clean way to group this with the moved logic?

Maybe a contextlib.contextmanager would let you handle cleanup cleanly - so like:

```
with load_builder() as builder:
    table_builder.execute_queries(cursor, schema, verbose)
```

Or just combine the loading and executing together anyway? Is the only difference between these two methods whether `drop_table` gets passed?

cumulus_library/studies/core/builder_core_medication.py
```diff
@@ -29,60 +30,45 @@ def _check_data_in_fields(self, cursor, schema: str):
...
-                                "row_alias": "row1",
-                            }
-                        ],
-                    )
-                    cursor.execute(query)
-                    progress.advance(task)
-                    if cursor.fetchone() is not None:
-                        data_types["inline"] = True
+                if "userselected" not in str(cursor.fetchone()[0]):
+                    missing_userselected = True
```
mikix, line 45: nit: flip this meaning to `has_userselected` - positive variables are way easier to reason about, I'd say

cumulus_library/studies/core/builder_core_medication.py
```diff
@@ -113,12 +99,17 @@ def prepare_queries(self, cursor: object, schema: str) -> dict:
         :param schema: the schema/db name, matching the cursor
 
         """
-        medication_datasources = self._check_data_in_fields(cursor, schema)
+        medication_datasources, missing_userselected = self._check_data_in_fields(
```
mikix, line 102: nit: one more of these and maybe just return a dict and pass it down like `**query_args`

cumulus_library/studies/core/builder_encounter_coding.py
```diff
@@ -168,3 +117,4 @@ def prepare_queries(self, cursor: object, schema: str):
                         table_cols=["id", "code", "code_system", "display"],
                     )
                 )
+        self.write_queries()
```
mikix, line 120: Is this a debug tool that got left in or was this an important step that was missing before? :smile:

cumulus_library/studies/core/manifest.toml
mikix: You gotta just hack your dictionary to remove the one and add the other :smile:

cumulus_library/template_sql/utils.py
```diff
@@ -0,0 +1,131 @@
...
+    base_col: str,
+    cursor,
+    progress: Progress,
+    task: Task,
+    coding_element="coding",
+) -> bool:
+    """Check db to see if codeableconcept data exists.
+
+    Will execute several exploratory queries to see if the column in question
+    can be queried naively. Will advance the associated progress's task by 3 steps.
```
mikix, line 31: There is a lot of implicit task total tracking happening - I don't have a better solution right now, but I'm registering my unease. As I add more progress bars to the ETL, prepare yourself to mock me when I do the same thing.

(One possible thought - at this point, trade precision-of-bar for maintainability and have this just advance 1 and/or bring the advancing up a layer to the caller who advances 1)

cumulus_library/template_sql/utils.py
```diff
@@ -0,0 +1,131 @@
...
+    :param progress: a Rich progress bar manager
+    :param task: The TaskID of the progress bar to update
+    :param coding_element: the place inside the code element to look for coding info.
+        default: 'coding' (and :hopefully: this is always right)
+    :returns: a boolean indicating if valid data is present.
+    """
+    query = get_is_table_not_empty_query(table, base_col)
+    cursor.execute(query)
+    progress.advance(task)
+    if cursor.fetchone() is not None:
```
mikix, line 47: nit: you could avoid indenting the whole function if you just early-exited instead and flipped this condition to `if not cursor.fetchone(): advance; return;` (and same comment later for the second logic branch)

That (to me) would make this more readable instead of a tree of "OK which path am I in?". This feels like a brain difference though, because I feel like we've talked about early-exiting vs if-trees before.

mikix (approved):

cumulus_library/base_table_builder.py
dogversioning: so - i hear you, this is a little fast and dirty. here is my thought:
- We are only using the table creator to, well, create tables
- It's case insensitive

but theoretically if you're doing something besides creating a table, we could modify this to do something like 'if the number of matches is just one, continue without trying to drop tables'.

I'm kind of willing to live with it being a bit risky because this is really just a development tool for iterating on a builder without commenting things out. That said, there are sql parsers we could try to use for this - lemme spend 15 mins looking into it.

cumulus_library/studies/core/builder_encounter_coding.py
dogversioning: debug, will trim.

cumulus_library/template_sql/utils.py
dogversioning: yeah i don't love it either - it could be a constant that's imported? but then we get into `total=len(dict)*utils.ADVANCE_NUM+2` or something.

I suppose in this case - we could consider this an 'atomic' operation and just advance once in the caller

cumulus_library/template_sql/utils.py
dogversioning: i think in this case it's because of the advance=2 call, which, per the other comment, if we treat this as atomic early exiting is fine.

mikix: Well regardless of atomic or not, you can just do:
```
if not cursor.fetchone():
    advance
    return
```

cumulus_library/base_table_builder.py
dogversioning: python-sqlparse is very unpleasant - i'm leaving the regex, but will check to see if the regex passes.

cumulus_library/template_sql/utils.py
dogversioning: oh of course

i'm still pulling it out, i think tracking the iterations across files is too much.

PR #103, opened 2023-08-08
Added core versioning

dogversioning:
This PR makes the following change:
- Added new core__meta_version table with basic version number (i.e. major only)
- Upload will now look for an exported {study}__meta_version table
- If found, will add version to aggregator post. If not, will assume no version (i.e. version 0)

This is one half of the implementation for versioning - the other half is at [this aggregator PR](https://github.com/smart-on-fhir/cumulus-aggregator/pull/99).

### Checklist

### DIFF
--- a/cumulus_library/studies/core/manifest.toml
+++ b/cumulus_library/studies/core/manifest.toml
@@ -10,6 +10,7 @@ file_names = [
 
 [sql_config]
 file_names = [
+    "version.sql",
     "setup.sql",
     "fhir_define.sql",
     "patient.sql",
@@ -36,4 +37,5 @@ export_list = [
     "core__count_condition_month",
     "core__count_medicationrequest_month",
     "core__meta_date",
+    "core__meta_version",
 ]
--- a/cumulus_library/studies/core/version.sql
+++ b/cumulus_library/studies/core/version.sql
@@ -1,0 +1,2 @@
+CREATE TABLE core__meta_version AS
+SELECT 1 AS data_package_version;
--- a/cumulus_library/upload.py
+++ b/cumulus_library/upload.py
@@ -28,7 +28,7 @@ def upload_data(
         json={
             "study": study,
             "data_package": data_package,
-            "version": f"{version:03d}",
+            "data_package_version": version,
             "filename": f"{args['user']}_{file_name}",
         },
         auth=(args["user"], args["id"]),
@@ -77,13 +77,13 @@ def upload_files(args: dict):
         sys.exit("user/id not provided, please pass --user and --id")
     with get_progress_bar() as progress:
         file_upload_progress = progress.add_task("Uploading", total=num_uploads)
-        meta_version = next(
-            filter(lambda x: "__meta_version" in str(x), file_paths), None
-        )
-        if meta_version:
+        try:
+            meta_version = next(
+                filter(lambda x: str(x).endswith("__meta_version"), file_paths)
+            )
             version = read_parquet(meta_version)["data_package_version"][0]
             file_paths.remove(meta_version)
-        else:
+        except StopIteration:
             version = 0
         for file_path in file_paths:
             upload_data(progress, file_upload_progress, file_path, version, args)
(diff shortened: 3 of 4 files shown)


cumulus_library/studies/core/version.sql
dogversioning, line 1: @comorbidity Bringing this pattern to your attention - other studies can copy this almost verbatim.

cumulus_library/studies/core/manifest.toml
```diff
@@ -36,4 +37,5 @@ export_list = [
     "core__count_condition_month",
     "core__count_medicationrequest_month",
     "core__meta_date",
+    "core__meta_version",
```
mikix, line 40: This might be a dumb question - but these meta tables are for the whole study, right? Why not just have one meta table with columns like `date` and `version`? I could imagine either approach making your life easier, so I assume that sort of reason is why.

cumulus_library/upload.py
```diff
@@ -4,24 +4,31 @@
...
     progress.update(file_upload_progress, description=f"Uploading {study}/{file_name}")
-    subscription = file_name.split(".")[0]
+    data_package = file_name.split(".")[0]
     prefetch_res = requests.post(
         args["url"],
         json={
             "study": study,
-            "data_package": subscription,
+            "data_package": data_package,
+            "version": f"{version:03d}",
```
mikix, line 31: Why the `:03d` formatting? That feels like it's for some implicit reason about what the aggregator would want to see? If so, I'd like to see either a comment about it or just send the aggregator an int and it can format it like it wants.

mikix (approved):

cumulus_library/studies/core/manifest.toml
dogversioning: I guess my thought was this: these are both quasi-optional, and may be present or absent, and this way there isn't a table schema dependency issue if we ever have to touch one or the other individually?

cumulus_library/upload.py
dogversioning: So my thought was this:
- I'd like to have sensible alphasorting version numbers, i.e. 001-009, then 010, when viewing data in buckets/transaction metadata
- 3 decimals is a paranoid number about how many changes we should see.
- If you are only looking at the library codebase, it might be more explicit as to what is happening

But - I can see an argument for doing this formatting on the receive side instead of the end side. I could go either way.

cumulus_library/upload.py
```diff
@@ -70,5 +77,13 @@ def upload_files(args: dict):
         sys.exit("user/id not provided, please pass --user and --id")
     with get_progress_bar() as progress:
         file_upload_progress = progress.add_task("Uploading", total=num_uploads)
+        meta_version = next(
+            filter(lambda x: "__meta_version" in str(x), file_paths), None
```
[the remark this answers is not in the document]
dogversioning: on 1: i'm happy to change to ends with, and that should be explicit
on 2: This basically prevents the validation/array extraction requirement, which I kind of like for brevity? but i can be talked out of it.

cumulus_library/upload.py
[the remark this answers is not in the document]
dogversioning: great call

mikix: I'm not done with the other PR, but as far as I can tell, it doesn't care about the format of the string at all - it just places the version on disk (though I suppose it could care about the format in so far as any slashes in it would mess it up real good).

Since the aggregator doesn't care, I think what you've done here makes sense. I just bumped on it because it was an unexpected place for the concern of "the admin of the aggregator would like visually pleasing sorted versions" to appear. (that's a totally valid concern, I just didn't expect it on the library side).

I do this kind of formatting in the ETL too - when writing out ndjson formats, I use 3 decimals. I get the desire.

Your code is fine as-is. You could send an int, which solves some problems (formatting can more properly live on the aggregator side and it doesn't have to check the contents of that string), but the library loses the ability to do minor versions or some other goofy thing. Trade-offs. I might land on the side of "send an int and have aggregator own the how-does-it-look-on-disk logic", but I'm not mad about it.

cumulus_library/upload.py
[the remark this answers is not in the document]
dogversioning: for posterity - taking suggestion 1, after some slack sidebar modified 2 to be try/except based for 'what :really: is the most "pythonic" way to do this?'

PR #105, opened 2023-08-15
flattened conditions, added new code table

dogversioning:
- adds a flag to allow codeable concepts tables to be priority filtered (i.e select the first found code, primarily for display reasons for core summary statistics) or just dump all found codes
- Adds both instances of these tables for condition
- Removes code from the core__condition table
  - The intent by now basically requiring a left join against one of the coding tables is to eliminate accidental double counting - this will require some work on the analyst's part
- condition__count now points at the explicitly filtered code list.

### Checklist

### DIFF
--- a/cumulus_library/studies/core/condition.sql
+++ b/cumulus_library/studies/core/condition.sql
@@ -38,7 +38,7 @@ SELECT
     t_category_coding.category_row.display AS category_display,
     tc.code,
     tc.code_system,
-    tc.display,
+    tc.code_display,
     tc.subject_ref,
     tc.encounter_ref,
     tc.condition_id,
--- a/cumulus_library/studies/core/condition.sql
+++ b/cumulus_library/studies/core/condition.sql
@@ -2,11 +2,12 @@
 -- Condition
 -- https://build.fhir.org/ig/HL7/US-Core/StructureDefinition-us-core-condition-encounter-diagnosis.html
 -- 
---Each Condition must have:
+-- Each Condition must have:
 --    a category code of “problem-list-item” or “health-concern”
---    a code that identifies the condition
+--    a code that identifies the condition (this is available in 
+--        core__condition_codable_concepts_all)
 --    a patient
---Each Condition must support:
+-- Each Condition must support:
 --    a clinical status of the condition (e.g., active or resolved)
 --    a verification status
 --    a category code of ‘sdoh’
@@ -14,12 +15,10 @@
 --    abatement date (in other words, date of resolution or remission)
 --    a date when recorded
 
-
 CREATE TABLE core__condition AS
 WITH temp_condition AS (
     SELECT
         c.category,
-        c.code,
         c.clinicalstatus,
         c.verificationstatus,
         c.subject.reference AS subject_ref,
@@ -31,8 +30,8 @@ WITH temp_condition AS (
 )
 
 SELECT
-    t_category_coding.category_row AS category,
-    tc.code AS cond_code,
+    t_category_coding.category_row.code AS category_code,
+    t_category_coding.category_row.display AS category_display,
     tc.subject_ref,
     tc.encounter_ref,
     tc.condition_id,
@@ -42,9 +41,8 @@ SELECT
     date_trunc('month', date(tc.recordeddate)) AS recorded_month,
     date_trunc('year', date(tc.recordeddate)) AS recorded_year
 FROM temp_condition AS tc,
-    unnest(category) AS t_category (category_coding), --noqa
-    unnest(category_coding.coding) AS t_category_coding (category_row), --noqa
-    unnest(code.coding) AS t_coding (code_row) --noqa
+    unnest(category) AS t_category (category_coding),
+    unnest(category_coding.coding) AS t_category_coding (category_row)
 WHERE tc.recordeddate BETWEEN date('2016-01-01') AND current_date;
 
 -- ###########################################################################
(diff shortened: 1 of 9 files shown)


dogversioning:
@comorbidity the core__condition table now looks like this:

```
category_code, category_display, subject_ref, encounter_ref, condition_id, condition_ref, recordeddate, recorded_week, recorded_month, recorded_year,
encounter-diagnosis, Encounter Diagnosis, Patient/d4c5c3e03a4b08f72a9f90f47313064893f93938491424d313e9220f103c6932, Encounter/9439b2b2495a4fe036c87fb215f39cb228f916d7e7272fce64ce5177ed933162, f245a896b2be1178c8542865fd72e0f74b0064f55927c937713e0e92679619ca, Condition/f245a896b2be1178c8542865fd72e0f74b0064f55927c937713e0e92679619ca, 2020-01-24, 2020-01-20, 2020-01-01, 2020-01-01
encounter-diagnosis, Encounter Diagnosis, Patient/0c6944dc40f5d40ad8c4296362289ca1cb28b106ac6053cc04d22e5ce73f169a, Encounter/63a9447e2e4499c4cd00fbefe472d6c9388e532abf047d0c454dfa3eb16808bf, f4e0b629eedd9427a9d2ae7c6f3db73836a73c37137c14d3571b12d9ab6e0791, Condition/f4e0b629eedd9427a9d2ae7c6f3db73836a73c37137c14d3571b12d9ab6e0791, 2017-12-21, 2017-12-18, 2017-12-01, 2017-01-01
```

cumulus_library/template_sql/templates.py
```diff
@@ -194,24 +194,37 @@ class CodeableConceptConfig:
...
     def __init__(
-        self, source_table: str, source_id: str, cc_column: dict, target_table: str
+        self,
+        source_table: str,
+        source_id: str,
+        column_name: str,
+        is_array: bool,
+        target_table: str,
+        filter_priority: bool = False,
+        code_systems: list = ["all"],
```
mikix, line 214: nit: I would strongly caution against any container as a default argument. It's only a _problem_ if you then try to modify that container, which isn't always the case (and doesn't seem to be here?). But it's a lingering rake to trip over.

(The problem is that Python creates the default argument _once_ and re-uses it. So any modifications to `code_systems` will carry over to future calls.)

I usually prefer to use `None` for this sort of thing, then in the body, do something like `code_systems = code_systems or ["all"]` to get a similar result without the risk and with only one line of fuss.

mikix (approved):

cumulus_library/studies/core/condition_codeable_concept.sql
```diff
@@ -59,43 +58,49 @@ CREATE TABLE core__condition_codable_concepts AS (
...
+        code,
+        code_system,
+        display
+    FROM partitioned_table
     WHERE available_priority = 1
-);
\ No newline at end of file
+);
+
+#non-filtering case
```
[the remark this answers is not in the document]
dogversioning: i meant to paste something after this and forgot after fixing it up - will update

cumulus_library/template_sql/templates.py
dogversioning: I had a similar thought about this really needing to be none - this ended up coming down to me being lazy w.r.t. how the jinja template is updated, and i just covid-brained shoved it here.

I do think there's some DRY repercussions in the template, but I can moved this sort of proxy for 'i need the for loop in the template to run exactly once' down into the function that uses this config on a 'is None' detection.

dogversioning: maybe not as much as i really thought - i'll just do it the right way rather than trying to hack around it.

dogversioning:
@comorbidity ok, now with inline code:

```
"category_code","category_display","code","code_system","display","subject_ref","encounter_ref","condition_id","condition_ref","recordeddate","recorded_week","recorded_month","recorded_year"
"encounter-diagnosis","Encounter Diagnosis","424393004","http://snomed.info/sct","Reports of violence in the environment (finding)","Patient/e98f5e2d69d22c0cbbdbb29b7ddd2343986b573634231ea3c7c872f1f97f6783","Encounter/850d9eefdee81130a93d44adb1b9192ac11556edfe1f135d0ef7fddf31b6ddfb","3ae75243f6a440b5cffbf4ddd5fdacea1b341e75bda511a0a2e6ba42dfa29ccb","Condition/3ae75243f6a440b5cffbf4ddd5fdacea1b341e75bda511a0a2e6ba42dfa29ccb","2017-06-17","2017-06-12","2017-06-01","2017-01-01"
"encounter-diagnosis","Encounter Diagnosis","195662009","http://snomed.info/sct","Acute viral pharyngitis (disorder)","Patient/96a9d32610487c94a7ea57a9b9514d675b4fe46567f1049e8e01a05862dcd5c2","Encounter/a97e0329993788e8cbec6896da6aee4cea3f15f6ef803cd6972662b360b5bd56","28d7658d49060f13ad987849998b52d9d131b39ff4772be21a7a7a1c1e1e9771","Condition/28d7658d49060f13ad987849998b52d9d131b39ff4772be21a7a7a1c1e1e9771","2016-09-22","2016-09-19","2016-09-01","2016-01-01"
```

cumulus_library/studies/core/condition.sql
```diff
@@ -2,37 +2,43 @@
...
-
 CREATE TABLE core__condition AS
 WITH temp_condition AS (
     SELECT
         c.category,
-        c.code,
         c.clinicalstatus,
+        cca.code,
+        cca.code_system,
+        cca.display,
```
comorbidity, line 25: for consistency, this should be called **code_display**

comorbidity (commented):
for consistency please prefix "code" for the **code_system** and **code_display**. Otherwise I reviewed with human eyeballs the SQL artifact but have nothing to share about the builder_*.py classes, that part is black magic to me. :) 

`Select ....
    tc.code,
    tc.code_system,
    tc.code_display,
....`

PR #109, opened 2023-08-21
Moved count gen to template

dogversioning:
- Reorg of counts generator to leverage BaseTableBuilder
- Updates CLI to dynamically build counts tables

### Checklist

### DIFF
--- a/cumulus_library/schema/counts.py
+++ b/cumulus_library/schema/counts.py
@@ -1,128 +1,64 @@
-# pylint: disable=W,C,R
+from typing import Union
+
 from cumulus_library.schema.valueset import DurationUnits
 from cumulus_library.schema.columns import ColumnEnum as Column
+from cumulus_library.template_sql import templates
 from cumulus_library.errors import LibraryError
 
-##################################################
-# Google Style Guide
-#
-# https://google.github.io/styleguide/pyguide.html
-# Design by composition composition for
-#
-# Query Grammar (in progress)
-#
-# Functions = support iterable lists/types
-# Objects = typesafety only
-##################################################
 
-
-def drop(table_or_view) -> str:
-    return f"DROP table if exists {table_or_view};"
-
-
-def create_view(view_name) -> str:
-    return f"CREATE or replace VIEW {view_name} AS "
-
-
-def name_view(count_from_table: str, duration_col=None):
-    if duration_col:
-        return f"{count_from_table}_{as_duration(duration_col).name}"
+def get_table_name(study_prefix: str, table_name: str, duration=None) -> str:
+    if duration:
+        return f"{study_prefix}__{table_name}_{duration}"
     else:
-        return f"{count_from_table}"
+        return f"{study_prefix}__{table_name}"
 
 
-def as_duration(col: str) -> DurationUnits:
-    if "_months" in col:
-        return DurationUnits.months
-    elif "_weeks" in col:
-        return DurationUnits.weeks
-    elif "_days" in col:
-        return DurationUnits.days
-
-
-def str_columns(cols) -> str:
-    if isinstance(cols, str):
-        return cols
-    if isinstance(cols, Column):
-        return cols.name
-    if isinstance(cols, list):
-        targets = [str_columns(c) for c in cols]
-        return ", ".join(targets)
-
-
-def where_clauses(clause=None, min_subject=10) -> str:
-    if not clause:
-        return f"cnt_subject >= {min_subject}"
+def get_where_clauses(
+    clause: Union[list, str, None] = None, min_subject: int = 10
+) -> str:
+    print(clause)
+    print(min_subject)
+    if clause is None:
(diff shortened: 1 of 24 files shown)


cumulus_library/studies/core/count_core.sql
dogversioning, line 1: 99% of the changes here are indentations just to support the addition of parenthesis after the start of the CTAS statement

cumulus_library/studies/core/manifest.toml
```diff
@@ -23,7 +23,11 @@ file_names = [
     "observation_vital_signs.sql",
     "medication_request.sql",
     "study_period.sql",
-    "count_core.sql"
+]
+
+[count_table_builder_config]
```
mikix, line 28: nit: you named the class `CountsBuilder` - maybe `counts_builder_config` would be a good symmetry? But no opinion myself

tests/test_cli.py
```diff
@@ -71,6 +71,22 @@ def test_cli_path_mapping(
...
+            cli_args=[
+                "build",
+                "-t",
+                "study_python_counts_valid",
+                "-s" "./tests/test_data",
+                "--database",
+                "test",
+            ]
+        )
+        builder.cursor.execute.assert_called()
```
mikix, line 87: might be worth testing that it actually did anything interesting too (like make an expected output file)

mikix (approved):

cumulus_library/schema/counts.py
```diff
@@ -1,128 +1,98 @@
...
+
+
+class CountsBuilder(BaseTableBuilder):
+    """Extends BaseTableBuilder for counts-related use cases"""
+
+    def __init__(self, study_prefix: str = None):
+        if study_prefix is None:
+            # This slightly wonky approach will give us the path of the
+            # directory of a class extending the CountsBuilder
+            study_path = f"{Path(sys.modules[self.__module__].__file__).parent}"
```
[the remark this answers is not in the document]
dogversioning: good catch, that was a carryover

cumulus_library/schema/counts.py
```diff
@@ -1,128 +1,98 @@
...
+        self, table_name: str, source_table: str, table_cols: list, **kwargs
+    ) -> str:
+        """Wrapper method for generating a counts table from a template"""
+        if not table_name or not source_table or not table_cols:
+            raise CountsBuilderError(
+                "count_query missing required arguments. " f"output table: {table_name}"
+            )
+        for key in kwargs:
+            if key not in ["min_subject", "where_clauses", "cnt_encounter"]:
+                raise CountsBuilderError(f"count_query recieved unexpected key: {key}")
```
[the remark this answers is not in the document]
dogversioning: there's a particular logic where, depending on the entry point you're coming in from, not all of these are going to be required, which i think is author friendly, so i'm ok just specifying the list of optional params here explicitly.

cumulus_library/studies/core/manifest.toml
dogversioning: good call

tests/test_cli.py
dogversioning: yeah - i sort of treated this as a second class test, since it is just testing a different entrypoint into a method that's tested elsewhere, but lemme try and improve this.

mikix: I hear ya, and you don't necessarily want to go overboard on an end-to-end unit test. But I like having all the glue between code tested in this way. Which you did! I would just throw a couple more checks in and call it a day.

dogversioning: the hard part about this is that the output of all this stuff is 'we executed some queries against athena' - if we had a mock, we could check that data existed.

mikix: Ah yeah - I live in the privileged ETL world, where I have files on disk.

If it's hard, don't stress about it, just a thought.

dogversioning: I did get in here and at least checked to make sure a sensible query exists in the call arg, which did root out an edge case bug, so it was worth doing.

I was thinking about this a little in the background, and I may open up a ticket with BCH IT to try and enable this approach for regression: https://github.com/smart-on-fhir/cumulus-library/issues/110

cumulus_library/study_parser.py
```diff
@@ -233,19 +241,24 @@ def _load_and_execute_builder(
...
                 f"Error loading {self._study_path}{filename}\n"
-                "Custom builders must extend the BaseTableBuilder "
-                "class exactly once per module."
+                "Custom builders must extend the BaseTableBuilder class."
             )
 
+        # Remove instances of intermediate classes, if present
+        for classref in table_builder_subclasses:
+            if classref.__name__ == "CountsBuilder":
+                table_builder_subclasses.remove(classref)
```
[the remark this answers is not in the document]
dogversioning: yeah i had that same concern - the lambda approach definitely feels safer.

PR #112, opened 2023-08-28
CI regression

dogversioning:
- Adds a basic regression test against a dedicated athena instance
  - Adds a CI action to run said test on PR
  - Adds a example IAM profile for configuring the AWS side of things
- Bugfix: added missing duration field to the appropriate counts generator (thanks regression)

### Checklist


.github/workflows/ci.yaml
```diff
@@ -47,3 +47,31 @@ jobs:
...
+    permissions:
+      id-token: write # This is required for requesting the JWT
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install .
+      - name: Configure AWS Credentials
+        uses: mcblair/configure-aws-profile-action@v0.1.1
```
dogversioning, line 61: See https://github.com/smart-on-fhir/cumulus-library/issues/113 for a discussion of this marketplace action.

tests/regression/run_regression.py
```diff
@@ -0,0 +1,65 @@
...
+                    filename,
+                    ("Rows differ between reference and export:\n" f"{compared}"),
+                ]
+            )
+if len(diffs) > 0:
+    for row in diffs:
+        print(f"--- {row[0]} ---")
+        print(row[1])
+    sys.exit(f"❌ Found {len(diffs)} difference(s). ❌")
+print("✅ Reference and export matched ✅")
```
mikix, line 65: :man_dancing: yay for fancy text :1st_place_medal:

tests/regression/run_regression.py
```diff
@@ -0,0 +1,65 @@
+"""Checks export against known dataset.
+
+This is intended to be primarily run as part of CI only, and does not leverage
+pytest due to the requirements for AWS secrets/federated tokens/network 
+connectivity."""
```
mikix, line 5: nit: I have not read rest of PR - so forgive me if this is answered elsewhere - but the "due to" is not immediately obvious to me why pytest wouldn't work. Maybe just like, one more sentence on that.

tests/regression/run_regression.py
```diff
@@ -0,0 +1,65 @@
...
+            export_missing.append(export)
+    sys.exit(
+        "Found differences in files present:\n"
+        f"Files present in reference not in export: {str(ref_missing)}\n"
+        f"Files present in export not in reference: {str(export_missing)}"
+    )
+diffs = []
+for filename in references:
+    if filename.endswith(".parquet"):
+        ref_df = read_parquet(f"{Path(__file__).resolve().parent}/reference/{filename}")
```
mikix, line 34: nit: maybe save these awkward paths in a variable at the top? You use them in two places, so just on safe side.

tests/regression/reference/core__count_condition_month.csv
mikix, line 1: I am just blindly trusting that these csv and parquets all make sense.

.github/workflows/ci.yaml
```diff
@@ -47,3 +47,33 @@ jobs:
...
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install .
+      # See https://github.com/aws-actions/configure-aws-credentials for configuring
+      # the aws side of this - the below action is just a light wrapper
+      - name: Configure AWS Credentials
+        uses: mcblair/configure-aws-profile-action@v0.1.1
+        with:
+          role-arn: arn:aws:iam::316333106222:role/cumulus-library-ci
```
mikix, line 65: Should this be a repo secret? I don't know how sensitive IAM numbers are.

Wait, speaking of - this is actually hitting athena, right? Is this cumulus-library-ci role wide open? (i.e. if I fork this repo and run CI, I'll be able to hit the same files) I don't see obvious repo secrets.

mikix (approved):

.github/workflows/ci.yaml
dogversioning: From aws: _While account IDs, like any identifying information, should be used and shared carefully, they are not considered secret, sensitive, or confidential information._ - that said, i could make this one a secret.

The real security comes in on the AWS side - over there, we have an identity provider to allow AWS traffic from github's official integration, and only when it comes from our github org and the cumulus-library repo, and it only gives them access to the sythetic regression S3 locations.

tests/regression/reference/core__count_condition_month.csv
dogversioning: they have been compared to each other, at least - but it's the same stuff that's going to the aggregator, just with synthetic data.

tests/regression/regression-iam.json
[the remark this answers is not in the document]
dogversioning: I can add a readme in this directory? otherwise i think the iam validator would complain

tests/regression/run_regression.py
dogversioning: I guess - my thought was 'if you are running pytest locally, and you want this to work, you either have to be signed into the BCH VPN, and have an auth token, or we have to open a much less secure way to use these credentials from another source'. The second is a non starter. The first is irritating enough that I kind of want to not have this picked up during a normal pytest run.

I could make this pytest-based and have this directory be normally excluded? but then you could run it manually with pytest if so inclined?

tests/regression/run_regression.py
```diff
@@ -0,0 +1,65 @@
...
+
+from pathlib import Path
+
+from pandas import read_parquet
+
+references = os.listdir(f"{Path(__file__).resolve().parent}/reference")
+
+exports = os.listdir(f"{Path(__file__).resolve().parent}/data_export/core")
+
+if not references == exports:
```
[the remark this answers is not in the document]
dogversioning: let's blame late day brain. will fix.

tests/regression/run_regression.py
```diff
@@ -0,0 +1,65 @@
...
+    ref_missing = []
+    for ref in references:
+        if ref not in exports:
+            ref_missing.append(ref)
+    export_missing = []
+    for export in exports:
+        if export not in references:
+            export_missing.append(export)
```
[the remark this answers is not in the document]
dogversioning: oh ho ho i like it.

tests/regression/run_regression.py
dogversioning: yeah i buy it

tests/regression/regression-iam.json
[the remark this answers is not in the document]
mikix: Maybe a readme would be good - "notes for running this yourself" and could include the mention of AWS-side restrictions to this repo, etc. that you should do.

.github/workflows/ci.yaml
mikix: Ah cool, that makes sense.

tests/regression/reference/core__count_condition_month.csv
mikix: Yeah moreso that the values themselves make sense, given the input data. But even if not manually checked, an automated baseline is still good to establish.

tests/regression/run_regression.py
mikix: I don't need it to be pytest, was just confused by the docstring here.

How about something like "this isn't in pytest because it isn't run as part of a test suite, and it's awkward to connect correctly to AWS (see ci.yaml)" -- I think I was struck by the "due to the requirements" making it sound impossible, but the real answer is "oy it's too much of a pain"

dogversioning: something in that flavor should work, yeah

tests/regression/regression-iam.json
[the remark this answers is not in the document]
dogversioning: ok, take a look at the added readme and let me know if you think it needs anything else.

tests/regression/run_regression.py
dogversioning: OK, this is more explicit now.

PR #119, opened 2023-09-06
Removed extra indexing on schema removal

dogversioning:
This PR removes a second instance of array indexing in the base_table_builder, which was causing us to search only the first character of a potential string for a name like `"schema"."table"`

This is currently only an issue when we are creating a completely empty table.
### Checklist


dogversioning:
Since this is affecting sites right now, I'm electing to not write a unit test this second, but am tracking the need for catching this case at https://github.com/smart-on-fhir/cumulus-library/issues/120

mikix (approved):

Issue #108, opened 2023-08-16
Better path traversal


dogversioning:
- adding a path should allow for arbitrary depth retrieval of manifest
- Check env var behavior
- Check behavior on retrieving study name (build/export)

dogversioning:
resolved via #123

PR #123, opened 2023-09-12
Better discovery of manifest location

dogversioning:
This PR addresses #108 by recursively searching for manifest files, rather than assuming the user is running this search in a specific place.

### Checklist

### DIFF
--- a/cumulus_library/cli.py
+++ b/cumulus_library/cli.py
@@ -213,13 +213,21 @@ def get_study_dict(alt_dir_paths: List) -> Optional[Dict[str, PosixPath]]:
     paths = [Path(cli_path, "studies")]
     if alt_dir_paths is not None:
         paths = paths + alt_dir_paths
-    for parent_path in paths:
-        for child_path in parent_path.iterdir():
-            if child_path.is_dir():
-                manifest_studies[child_path.name] = child_path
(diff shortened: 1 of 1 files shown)


cumulus_library/cli.py
```diff
@@ -213,15 +213,23 @@ def get_study_dict(alt_dir_paths: List) -> Optional[Dict[str, PosixPath]]:
...
-                manifest_studies[child_path.name] = child_path
-            elif child_path.name == "manifest.toml":
-                manifest_studies[parent_path.name] = parent_path
+    for path in paths:
+        found_studies = get_studies_by_manifest_path(path)
+        manifest_studies.update(found_studies)
     return manifest_studies
 
 
+def get_studies_by_manifest_path(path: PosixPath) -> dict:
```
mikix, line 222: This seems like something I would go to `os.walk` for, but PosixPath is doing a lot of the work for you too. :+1:

mikix (approved):

cumulus_library/cli.py
dogversioning: There's def more than one way to do this - I'm making a conscious decision over in this to use pathlib.PosixPath when there's rough equivalence just so there's one pattern for someone to pick up.

PR #125, opened 2023-09-18
Added upload check for targets

dogversioning:
This PR updates the upload `--target` argument to actually filter the list of files to upload

### Checklist

### DIFF
--- a/cumulus_library/upload.py
+++ b/cumulus_library/upload.py
@@ -72,7 +72,7 @@ def upload_files(args: dict):
             "study export folder."
         )
     file_paths = list(args["data_path"].glob("**/*.parquet"))
-    if args["target"] is not None:
+    if args["target"]:
         filtered_paths = []
         for path in file_paths:
             if any(study in str(path) for study in args["target"]):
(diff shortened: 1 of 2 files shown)


tests/test_cli.py
```diff
@@ -217,3 +218,32 @@ def test_cli_upload_studies(mock_glob, args, status, login_error, raises):
...
+            Path(
+                str(Path(__file__).parent) + "/test_data/count_synthea_patient.parquet"
+            ),
+            Path(
+                str(Path(__file__).parent) + "/other_data/count_synthea_patient.parquet"
+            ),
+        ],
+    ]
+    cli.main(cli_args=args)
+    assert mock_upload_data.call_count == calls
```
mikix, line 249: nit: I might prefer to also see a comparison of the actual paths, to confirm the filtering got the right ones.

mikix (approved):

cumulus_library/upload.py
```diff
@@ -72,6 +72,13 @@ def upload_files(args: dict):
             "study export folder."
         )
     file_paths = list(args["data_path"].glob("**/*.parquet"))
+    if args["target"] is not None:
```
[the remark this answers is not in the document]
dogversioning: yeah good point.

tests/test_cli.py
dogversioning: Ok, this is lightly contrived, but there is a check now making sure the right path was selected.

PR #126, opened 2023-09-29
Updates for i2b2 export support

dogversioning:
- Updates to the `clean` arg group (really only used by us, and infrequently) to adapt to prefixes in manifests, and cleaning up dangling old tables/user created tables.
- Updates supporting i2b2 injestion
  - Moved encounter coding output to link to denormalized tables
  - Added static table for encounter class display fields
  - Fixed namespace class vs display distinction

### Checklist


cumulus_library/study_parser.py
```diff
@@ -212,8 +226,12 @@ def _execute_drop_queries(
         :param task: a TaskID for a given progress bar
         """
         for view_table in view_table_list:
+            if explicit_prefix:
+                prefix = f"{view_table[0]}"
+            else:
+                prefix = f"{view_table[0]}__"
             drop_view_table = get_drop_view_table(
-                name=view_table[0], view_or_table=view_table[1]
+                name=f"{view_table[0]}", view_or_table=view_table[1]
```
mikix, line 234: Did you mean to pass in `prefix` here? I don't see it being used.

(And if so, and tests didn't catch this - maybe add a test that would)

cumulus_library/study_parser.py
```diff
@@ -179,20 +183,29 @@ def clean_study(
         # We'll do a pass to see if any of these tables were created outside of a
         # study builder, and remove them from the list.
         for view_table in view_table_list.copy():
-            if any(
-                view_table[0].startswith(f"{self.get_study_prefix()}__{word}_")
-                for word in RESERVED_TABLE_KEYWORDS
-            ):
+            if any((f"_{word}_") in view_table[0] for word in RESERVED_TABLE_KEYWORDS):
```
mikix, line 186: So harsh! No more `covid_study__this_is_my_nlp_table`?

We might want to also catch the word at the end (which we didn't do before either) -- `prefix__nlp` or `prefix__this_is_nlp`)

cumulus_library/cli.py
```diff
@@ -89,7 +89,7 @@ def reset_data_path(self, study: PosixPath) -> None:
 
     ### Creating studies
 
-    def clean_study(self, targets: List[str]) -> None:
+    def clean_study(self, targets: List[str], explicit_prefix=False) -> None:
```
mikix, line 92: Can you expand a bit on what the explicit prefix stuff is all about? It makes me lightly nervous to allow cleaning on any old prefix - just seems very "shoot yourself in the foot" territory.

If it's to clean up for oddball historical tables... Maybe instead of an explicit prefix, we could do explicit full-name matching. I.e. Don't allow cleaning on `^m*` but rather just `^mistake$` -- or is that not a real issue / not a good idea?

cumulus_library/cli_parser.py
```diff
@@ -124,8 +121,15 @@ def get_parser() -> argparse.ArgumentParser:
     clean = actions.add_parser(
         "clean", help="Removes tables & views beginning with '[target]__' from Athena"
     )
+
     add_target_argument(clean)
+    add_verbose_argument(clean)
     add_aws_config(clean)
+    clean.add_argument(
+        "--explicit_prefix",
```
mikix, line 129: A) hyphen not underscore my dude

B) I know why you named it this - in your head you were thinking "this is the explicit string I want to check against".

But my head bumped against this a little bit, feeling like the intent should actually be reversed. With this flag enabled, it's actually a really loosey goosey prefix. Potentially matching multiple studies. Without this flag, you have a tightly defined explicit one-study prefix.

Would users of the CLI even really think of it in terms of a "prefix?" Like, to me I think of it as "I'm passing a study name, and the tool is managing how it cleans that up." Internally, we often use it as a prefix. But that's not what I would be thinking of when I pass in the CLI arg.

I don't have a better arg name though. `--wildcard-target`... `--regex-target` (and actually allow a regex?)

I dunno. Maybe the possible confusion here is light and your arg is perfect. Will defer to your brain on this.

mikix (approved):

cumulus_library/study_parser.py
mikix: Also, maybe add a test for this new restriction.

cumulus_library/cli_parser.py
dogversioning: A) i flipped it, and then went back on it, because i've already committed the sin here elsewhere. i should make a ticket to clean it up.

B) this is definetely a 'i am thinking about this based on the way the non-flag version of this works' - i.e. you infer the prefix from the study name in the manifest, and then add a dunder to it, so you always have a coerced prefix. When you're explicitly saying 'use this prefix i'm giving you and trust me on it and don't bother checking a manifest' with this flag.

Does that help? I don't love the name either, open to suggestions.

cumulus_library/study_parser.py
dogversioning: we had some of those in the historical stuff i was cleaning up - that's the source of this change, but i think i stand by it if the reserved with is small. can add a endswith check and a test for this.

cumulus_library/study_parser.py
dogversioning: yeah, good catch and call

cumulus_library/study_parser.py
mikix: Yeah, easier to be strict than not

cumulus_library/cli_parser.py
mikix: B) Yeah I guess I was bumping on the fact that this name is referencing some internal details instead of abstracting it. Like, in Matt's brain, deeply steeped in this code, obviously the `--target` is a prefix that we append a dunder to. But a poor little user might just be thinking "pls delete this study's tables, I don't care or know how that works behind the scenes". The fact that it is internally a prefix may not occur to them, so `--explicit-prefix` might not register as anything. That's all I was getting at.

Plus to my external brain, I was thinking that the default behavior actually uses a more explicit prefix than _with_ the flag. So even for a half-steeped person, this was confusing.

That all said, :shrug: you can just document it well enough and it's fine (this is a high risk place to have some confusion, but casuals hopefully aren't running this or adding arguments they don't understand).

cumulus_library/cli.py
mikix: My concern was just avoiding accidental deletion.

- Yeah you could manually drop the tables in Athena, but that's more technical - I had assumed this was exactly that, but for casuals who aren't 100% how to do that.
  - Whether this "full name" idea makes sense depends on how many tables we're talking about in those use cases. Like 10 might be annoying to do individually. But 3? :shrug:
- It might be nice to have like, a guard against destructive actions, since this can affect tables the user might not be thinking about (like, they forget that covid_symptoms__ exists when they're trying to delete some old covid_junk)
  - Something like `This command will delete these tables: [...] Continue? [y/N]`
  - Doesn't have to be this PR, but it feels like a nice guard, now that we're moving past single study actions.

Regardless, this is fine - I just get real nervous about wildcard deletes with no ability to undo it :smile:

mikix: (I know that's ironic coming from someone who recently had the NdjsonFormat class uncritically delete a dir. But that did get fixed for a reason.)

cumulus_library/cli_parser.py
dogversioning: `--specify-prefix`? and then a big warning comment about how 99% of the time you shouldn't need to use it?

mikix: I think I like `--explicit-prefix` better, despite my whinging. Docs can make it clear. Though... what about just `--prefix`? If you don't like that, I'd say stick with `--explicit-prefix`

cumulus_library/cli.py
dogversioning: i like the list & confirm idea - i'll either squeeze it in now or make a ticket for it.

cumulus_library/cli_parser.py
dogversioning: `--prefix` it is

cumulus_library/study_parser.py
```diff
@@ -427,6 +445,9 @@ def export_study(self, cursor: object, data_path: PosixPath) -> List:
             dataframe = cursor.execute(query).as_pandas()
             path = Path(f"{str(data_path)}/{self.get_study_prefix()}/")
             path.mkdir(parents=True, exist_ok=True)
+            dataframe = dataframe.sort_values(
+                by=list(dataframe.columns), ascending=False, na_position="first"
+            )
```
dogversioning, line 450: After seeing the churn in the reference data files, I threw this in, which gets us back to the original format for exports before we removed the ORDER BY clauses for performance reasons.

cumulus_library/cli.py
dogversioning: #129

Issue #78, opened 2023-07-11
Multiple Race is confusing "White American Indian or Alaska Native"


comorbidity:
If this is supposed to be "White" AND "American Indian or Alaska Native", then there should be a separator between the labels like "[White][American Indian or Alaska Native]"

dogversioning:
So we need an example of a dataset containing more than one code per patient before we can do any design work here. we haven't found this in either the cerner data or synthea data. But right now, this field is an array, per patient, of every code available from the dataset. If this was present, it would look like this: "[White,American Indian or Alaska Native]"

comorbidity:
UCDavis is where I saw this happen (Epic EHR)

dogversioning:
i don't know if they're comfortable with this, but i'd love to see the post-ETL json/have them run a specific query against the etl patient table for something with multiple race extensions and send us a data sample.

PR #133, opened 2023-10-17
Data quality updates

dogversioning:

- More robust handling of multiple race/ethnicity medical codes #78
- Added mapping of arbitrary encounter codes to reference codes
  - Adjusted the naming of `fhir_define.sql` to two seperate files w/ better semantic naming.
- Added filtering for FHIR resources to exclude things like future/cancelled encounters.
- Made relative links in docs absolute for external doc site.
- Regenned some reference sql, since it had been a bit

### Checklist


cumulus_library/studies/core/documentreference.sql
```diff
@@ -67,7 +67,7 @@ WITH powerset AS (
         d.author_month,
         e.enc_class_display
     FROM core__documentreference AS d, core__encounter AS e
-    WHERE d.encounter_ref = e.encounter_ref
+    WHERE d.encounter_ref = e.encounter_ref AND d.status = 'current'
```
mikix, line 70: [docStatus](https://www.hl7.org/fhir/R4/documentreference-definitions.html#DocumentReference.docStatus) might also be useful to filter on? Like only allow `final` or `amended` docs.

cumulus_library/studies/core/encounter.sql
```diff
@@ -56,7 +57,9 @@ SELECT DISTINCT
     p.ethnicity_display,
     p.postalcode3
 FROM temp_encounter AS e
-LEFT JOIN core__act_encounter_code_v3 AS ac ON ac.code = e.class.code
+LEFT JOIN core__fhir_mapping_expected_act_encounter_code_v3 AS eac
+    ON eac.found = e.class.code
+LEFT JOIN core__fhir_act_encounter_code_v3 AS ac ON eac.expected = ac.code
```
mikix, line 62: Both these lines should also be checking the `system` yeah? Otherwise who knows what you are matching on.

cumulus_library/studies/core/fhir_mapping_tables.sql
```diff
@@ -111,10 +113,28 @@ SELECT * FROM
...
-    ) AS t (define, url); --noqa: AL05
+        ('VitalSign', 'http://hl7.org/fhir/observation-vitalsigns.html')
+    ) AS t (resource, uri); --noqa: AL05
+
+-- ############################################################
+-- FHIR mapping of as found Encounter codes to the expected encounter code from
+-- http://hl7.org/fhir/STU3/v3/ActEncounterCode/vs.html
+
+CREATE TABLE core__fhir_mapping_expected_act_encounter_code_v3 AS
+SELECT * FROM
```
mikix, line 124: Same comment as above: this table should include systems for the found side.

cumulus_library/template_sql/count.sql.jinja
```diff
@@ -4,7 +4,37 @@ it simultaneously asks for indentation of 4 and 8 spaces.
...
+            {%- endif -%}
+            {% for col in table_cols %}
+            "{{ col }}"
+            {%- if not loop.last -%}
+            , 
+            {%- endif -%}
+            {%- endfor %}
+        FROM {{ source_table }}
+        {%- if fhir_resource=='document' and filter_resource %}
+        WHERE (status = 'current')
```
mikix, line 29: Similar comment as above - maybe consider checking `docStatus` too

cumulus_library/template_sql/templates.py
```diff
@@ -151,7 +154,9 @@ def get_count_query(
     min_subject: int = 10,
     where_clauses: Optional[list] = None,
     fhir_resource: Optional[str] = None,
+    filter_resource: Optional[bool] = True,
 ) -> str:
+    print(filter_resource)
```
mikix, line 159: :scissors:

tests/regression/reference/core__count_documentreference_month.csv
mikix, line 1: What happened here? I assume it's fine, just I didn't expect this kind of change from the above python changes (was this from `status` filtering?)

dogversioning: look at this nonsense:
```#	status	instances
1	current	1144
2	superseded	86100
```

cumulus_library/studies/core/encounter.sql
dogversioning: So, on the one hand, i agree; on the other, i think the namespace for this one is small enough that it likely doesn't matter.

I'd almost rather wait to see if someone responds to the request i made in slack for running a query from a different EHR and see how bad this looks in practice. But I can do it out of maximum parinoia.

tests/regression/reference/core__count_documentreference_month.csv
mikix: Is this from synthea? That's a lot of superceding! I wonder if that was an attempt to be realistic.

cumulus_library/studies/core/documentreference.sql
dogversioning: status is a required field, so i think that's ok to add without analysis, but since this is optional, i'm not sold without getting some info about other EHR implementations.

tests/regression/reference/core__count_documentreference_month.csv
dogversioning: it is - so either they were trying to be realistic or they got a very skewed sample.

cumulus_library/studies/core/documentreference.sql
mikix: Let's clarify what my suggestion was - I'm surprised to hear pushback, so maybe I wasn't clear.

I was suggesting a line like `WHERE d.docStatus in (null, 'final', 'amended')` or actual good syntactical equivalent.

Is your concern that `preliminary` or `entered-in-error` are misused in some EHRs?

(FYI, the NLP tasks we have do the filtering I suggest on the ETL side - check status and docStatus)

cumulus_library/studies/core/encounter.sql
mikix: Oh OK let me pause approval until we iron this out - this is the only real concern I had.

What's the argument for not checking system? The `R` code could be used in some random EHR-vendor-specific-system and mean a totally different thing. You _have_ to always treat codes & systems as a cohesive unit. Otherwise the code doesn't mean anything.

mikix (changes requested):

cumulus_library/studies/core/documentreference.sql
mikix: Ah rereading your comment, you thought I wanted to disallow null. No - that would be crazy, IMO, for an optional field.

But there is low hanging fruit of disallowing entered-in-error and preliminary docs from cluttering up our dashboards.

dogversioning: ok, yeah, with null this is more reasonable

cumulus_library/studies/core/encounter.sql
dogversioning: as an aside - i'm thinking about creating some kind of runnable artifact as part of this repo we can ask folks something like 'run this and send it back to us so we can confirm data issues' to gather information like this.

mikix: We chatted out of band - here's my understanding of our convo, Matt, for posterity:

- We both want to eventually get to the place I described above.
- We can't do it today because some Epic sites provide the same _code_ but with an Epic-internal _system_ -- and the code in this PR allows for that. If we changed it now to require a system, we might be tossing some Epic data aside (though Mike thinking now -- we don't believe it's the existing Epic encounter system we use in library? `urn:oid:2.16.840.1.113883.4.642.3.248`)
- We're asking some sites to run some SQL to provide systems for their codes, and once that's in hand we can come back and update this.

So I'll re-approve. It sounds like we're both moving toward the same vision. I'm just more impatient than Matt.

mikix (approved):

cumulus_library/studies/core/encounter.sql
dogversioning: for posterity - we believe that that urn is :probably: equivalent to http://hl7.org/fhir/ValueSet/encounter-type, which is not what we're trying to use here.

Issue #132, opened 2023-10-17
Technical Debt - seperate unit test expected queries from test logic


dogversioning:
Since these queries can get very large, it becomes hard to manage the actual logic of the unit test - it might be better to move these to individual files instead.

PR #134, opened 2023-10-19
Discovery study

dogversioning:
- Adds a new study for data discovery at remote sites
- Adds a builder for extracting coding systems from FHIR resources

### Checklist


cumulus_library/studies/discovery/code_definitions.py
```diff
@@ -0,0 +1,53 @@
+# A collection of codes & codeableConcepts to extract available codes from.
```
dogversioning, line 1: Philosophizing: I'm starting to get to the point over here where I'm considering more general seperation of table business logic from the bits that do the building. Somewhat releated, I created this: #132 - looking for a temperature read about this generally.

cumulus_library/studies/discovery/code_detection.py
```diff
@@ -0,0 +1,84 @@
...
+        code_sources = []
+        for code_definiton in code_list:
+            if any(
+                x not in code_definiton.keys() for x in ["table_name", "column_name"]
+            ):
+                raise KeyError(
+                    "Expected table_name and column_name keys in "
+                    f"{str(code_definiton)}"
+                )
+            code_source = {
```
dogversioning, line 73: this :could: be a class, either in here or in the data object above. I don't think it's quite there yet but I could be convinced otherwise.

cumulus_library/studies/discovery/manifest.toml
```diff
@@ -0,0 +1,12 @@
+study_prefix = "discovery"
```
dogversioning, line 1: this name is just ok - soliciting options.

cumulus_library/.sqlfluff
```diff
@@ -20,6 +20,7 @@ code_systems = ["http://snomed.info/sct", "http://hl7.org/fhir/sid/icd-10-cm"]
 col_type_list = ["a string","b string"]
 cc_columns = [{"name": "baz", "is_array": True}, {"name": "foobar", "is_array": False}]
 cc_column = 'code'
+code_system_tables = [{table_name":"hasarray","column_name":"acol","is_bare_coding":False,"is_array":True, "has_data": True},{"table_name":"noarray","column_name":"col","is_bare_coding":False,"is_array":False, "has_data": True}{"table_name":"bare","column_name":"bcol","is_bare_coding":True,"is_array":False, "has_data": True},{"table_name":"empty","column_name":"empty","is_bare_coding":False,"is_array":False, "has_data": False}]
```
mikix, line 23: I remember this whole setup being explained to me before and it making sense. But it definitely feels awkward.

cumulus_library/template_sql/utils.py
```diff
@@ -46,13 +49,20 @@ def is_codeable_concept_populated(
...
-
-    query = get_column_datatype_query(schema, table, base_col)
-    cursor.execute(query)
 
-    if coding_element not in str(cursor.fetchone()[0]):
+        query = get_column_datatype_query(schema, table, base_col)
+        cursor.execute(query)
+        schema_str = str(cursor.fetchone()[0])
+        if allow_partial:
+            if coding_element not in schema_str:
```
mikix, line 57: nit: totes ignorable, but a way to have less logic forks here would be something like:
```
required_fields = [coding_element]
if not allow_partial:
    required_fields += ["code", "system", "display"]
if any(x not in schema_str for x in required_fields):
    return False
```

cumulus_library/template_sql/utils.py
```diff
@@ -91,21 +102,35 @@ def is_codeable_concept_array_populated(
...
         if cursor.fetchone() is None:
             return False
-    except:
-        return False
 
-    query = get_column_datatype_query(schema, table, base_col)
-    cursor.execute(query)
-    if coding_element not in str(cursor.fetchone()[0]):
-        return False
+        query = get_column_datatype_query(schema, table, base_col)
```
mikix, line 119: nit: This whole method feels like a lot of tricky duplicated logic with the above.

cumulus_library/template_sql/utils.py
```diff
@@ -126,3 +151,52 @@ def is_codeable_concept_array_populated(
...
+        This can be a nested element, like column.object.code
+    :param cursor: a PEP-249 compliant database cursor
+    :allow_partial: If true, codings which do not have fields expected by the library
+        will still be included, and will need to be manually coerced.
+    :returns: a boolean indicating if valid data is present.
+    """
+
+    # if the source column is missing for some reason (i.e. we're dealing with
+    # conversion to FHIR rather than a true FHIR source and it's incomplete),
+    # we'll return false
```
mikix, line 180: I take slight umbrage with calling the hard-working ETL's i2b2-converted FHIR _incomplete_ and not a _true FHIR source_. :smile:

cumulus_library/template_sql/utils.py
```diff
@@ -126,3 +151,52 @@ def is_codeable_concept_array_populated(
...
+        if cursor.fetchone() is None:
+            return False
+
+        query = get_column_datatype_query(schema, table, base_col)
+        cursor.execute(query)
+        schema_str = str(cursor.fetchone()[0])
+    except:
+        return False
+    if any(x not in schema_str for x in ["code", "system", "display"]):
+        return False
```
mikix, line 193: nit: again, duplicated code with the above two methods. Feels like a utility method that takes in a list of required fields to check would be able to be shared?

cumulus_library/studies/discovery/manifest.toml
mikix: For context, will this study also maybe be used in the future for other sort of "is your site's data ready for Cumulus Library or do we need to make tweaks" kind of stuff (i.e. it will be a component of site on-boarding)?

Brainstorming: `onboarding`, `data-quality`, `data-qa`, `data-report`, `fhir-quality`, `library-prep`

I dunno. Discovery works for me, though it is lightly generic. But that might be good, depending on how kitchen sink we want it to be able to get.

cumulus_library/studies/discovery/code_detection.sql
```diff
@@ -0,0 +1,41 @@
+-- noqa: disable=all
+/*
+This is a reference output of the SQL generated by builder_code_detection.py 
+that is used by the core__encounter_type table. It is not invoked directly.
```
mikix, line 4: This makes me wonder: why is it checked in? Does anything reference it? How is it generated? Can it be out of date?

This sql is also in the unit test as an expected string, yeah?

cumulus_library/studies/discovery/code_detection.py
```diff
@@ -0,0 +1,84 @@
...
+    is_codeable_concept_array_populated,
+    is_codeable_concept_populated,
+    is_code_populated,
+)
+
+from cumulus_library.studies.discovery.code_definitions import code_list
+
+
+class CodeDetectionBuilder(BaseTableBuilder):
+    display_text = "Selecting unique code systems..."
```
mikix, line 16: unicode nerd nit: don't you know there's a character for that?  :smile:  U+2026: … or in monospace `…`

cumulus_library/studies/discovery/code_detection.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def _check_codes_in_fields(self, code_sources: list[dict], schema, cursor) -> dict:
+        """checks if Coding/CodeableConcept fields are present and populated"""
+
+        with get_progress_bar() as progress:
+            task = progress.add_task(
+                "Discovering available coding systems...",
+                # Each column in code_sources requires at most 3 queries to
+                # detect valid data is in the DB
+                total=len(code_sources),
```
mikix, line 26: nit: The comment feels like it harkens from a time when you had `len() * 3`

cumulus_library/studies/discovery/code_definitions.py
```diff
@@ -0,0 +1,53 @@
...
+# is_array and is_bare_code are assumed to be false unless explicitly set true.
```
mikix, line 2: nit: Maybe a quick comment that they both also can't be true - which is presumably obvious, but still.

And/or maybe like a light explanation of each option:
- `is_array`: array of [CodeableConcepts](https://www.hl7.org/fhir/R4/datatypes.html#CodeableConcept)
- `is_bare_code`: direct [Coding](https://www.hl7.org/fhir/R4/datatypes.html#Coding), not a full CodeableConcept
- neither: one CodeableConcept

As/if/when you add future field types, would you add more booleans? Maybe a string/enum field makes sense here instead of a boolean tree. Especially since the default option isn't an intuitive lack of the other booleans being true, but like, just a third option.

cumulus_library/studies/discovery/code_detection.py
```diff
@@ -0,0 +1,84 @@
...
+    def prepare_queries(self, cursor: object, schema: str):
+        """Constructs queries related to condition codeableConcept
+
+        :param cursor: A database cursor object
+        :param schema: the schema/db name, matching the cursor
+
+        """
+
+        code_sources = []
+        for code_definiton in code_list:
```
mikix, line 65: typo: definition

cumulus_library/studies/discovery/code_detection.py
```diff
@@ -0,0 +1,84 @@
...
+            if any(
+                x not in code_definiton.keys() for x in ["table_name", "column_name"]
+            ):
```
mikix, line 68: nit: This is a curious check - you are basically validating that you yourself didn't make a mistake when writing `code_list`? Maybe a unit test makes more sense for that?

cumulus_library/studies/discovery/code_detection.py
mikix: Is this a place where a `dataclass` or `namedtuple` would be the right level of complexity? I'm fine with it as-is, but just a thought.

cumulus_library/studies/discovery/code_definitions.py
mikix: I'm not 100% understanding your options here, maybe an in-person rubber-ducking would help. Lunch talk or something.

mikix (approved):

cumulus_library/studies/discovery/code_detection.py
dogversioning: This is due to the seperation and/or having another user less familiar editing the definitions file - basically a consequence of not using a class.

cumulus_library/template_sql/utils.py
mikix: But the FHIR isn't incomplete. It's totes valid spec-compliant FHIR right? The Library is annoyed that it leaves out some optional elements, but that's not a fault of the FHIR.

cumulus_library/studies/discovery/code_detection.sql
mikix: Yeah but maybe comment some of that stuff - like why it exists. How to reproduce it - maybe a comment in the unit test that also uses this same SQL like "if you change this unit test, also go regenerate our sample in this location" -- all that sort of stuff. Or better yet, use this SQL directly in the unit test so it won't be possible to get stale.

Should this file be in the shipped python or in the test code?

All these questions swirl in my mind when I see a pile of SQL that might bitrot in a week. I'd vote for some more comments/safeguards etc.

cumulus_library/studies/discovery/code_detection.py
mikix: Sure I love the check. No complaints about that. It just struck me as an odd place to do it - more of a unit test than a runtime test in my head. But just a nit - location doesn't matter too much and this isn't performance sensitive code.

cumulus_library/template_sql/utils.py
dogversioning: I guess the FHIR itself is valid - the schema we're creating from that FHIR from a SQL perspective causes headaches, but that's a seperate thing. trimmed to something more concise

cumulus_library/studies/discovery/code_detection.py
dogversioning: reminder that the contract we have in this repo is that studies themselves are exempt from unit tests, for ease of use by clinical folks - that's the reason for here rather than a unit test

mikix: Oh sure but like. This is a study we control, in the main library repo, where there are unit tests. Like we could still use one for us right?

I'm not trying to advocate hard for a unit test in this specific case, who cares. But that distinction is troubling to me -- I'd think while we don't require others to have production-quality full QA practices etc, we'd want them as much as possible for our own sanity. None of our code under studies/ has any unit testing?

cumulus_library/studies/discovery/code_detection.sql
dogversioning: Ok, I've updated this to reflect the synthea format, for the time being.

Per our side conversation, I created a ticket to get this into regression against the synthea dataset: #135

Issue #129, opened 2023-10-12
Tech Debt: Add list of tables and confirmation to manual prefix dropping


dogversioning:
This is a little dangerous, so we should ask users to confirm before dropping all tables starting with c, for example

PR #140, opened 2023-10-23
Confirm drop by prefix, add --continue

dogversioning:
- Asks for confirmation when using the `--prefix` argument #129 
- Adds a `--continue` param for specifying a point in the manifest's list of SQL files to continue a build from (useful for debugging errors that occur at near the end of a build)
- As a result of doing this, cleaned up some places where clean wasn't quite wired into reading a prefix directly from a manifest

### Checklist

### DIFF
--- a/cumulus_library/cli_parser.py
+++ b/cumulus_library/cli_parser.py
@@ -141,6 +141,12 @@ following order of preference is used to select credentials:
     add_verbose_argument(build)
     add_aws_config(build)
 
+    build.add_argument(
+        "--continue",
+        dest="continue_from",
+        help=(argparse.SUPPRESS),
+    )
+
     export = actions.add_parser(
         "export", help="Generates files on disk from Athena views"
     )
--- a/cumulus_library/cli_parser.py
+++ b/cumulus_library/cli_parser.py
@@ -16,7 +16,7 @@ def add_table_builder_argument(parser: argparse.ArgumentParser) -> None:
     """Adds --builder arg to a subparser"""
     parser.add_argument(
         "--builder",
-        help=(argparse.SUPPRESS),
+        help=argparse.SUPPRESS,
     )
 
 
@@ -129,7 +129,7 @@ following order of preference is used to select credentials:
     clean.add_argument(
         "--prefix",
         action="store_true",
-        help=(argparse.SUPPRESS),
+        help=argparse.SUPPRESS,
     )
 
     build = actions.add_parser(
@@ -145,7 +145,7 @@ following order of preference is used to select credentials:
     build.add_argument(
         "--continue",
         dest="continue_from",
-        help=(argparse.SUPPRESS),
+        help=argparse.SUPPRESS,
     )
 
     export = actions.add_parser(
--- a/cumulus_library/cli_parser.py
+++ b/cumulus_library/cli_parser.py
@@ -123,6 +123,7 @@ following order of preference is used to select credentials:
     )
 
     add_target_argument(clean)
+    add_study_dir_argument(clean)
     add_verbose_argument(clean)
     add_aws_config(clean)
     clean.add_argument(
--- a/cumulus_library/study_parser.py
+++ b/cumulus_library/study_parser.py
@@ -94,15 +94,11 @@ class StudyManifestParser:
         sql_config = self._study_config.get("sql_config", {})
         sql_files = sql_config.get("file_names", [])
         if continue_from:
-            match_found = False
-            for file in sql_files:
-                print(continue_from.replace(".sql", ""))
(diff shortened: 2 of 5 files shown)


tests/test_study_parser.py
```diff
@@ -74,37 +75,38 @@ def test_manifest_data(manifest_key, raises):
...
 )
 @mock.patch("cumulus_library.helper.query_console_output")
-def test_clean_study(mock_output, schema, verbose, prefix, query_res, raises):
+def test_clean_study(mock_output, schema, verbose, prefix, confirm, query_res, raises):
     with raises:
-        mock_cursor = mock.MagicMock()
-        mock_cursor.__iter__.return_value = [[query_res]]
-        parser = StudyManifestParser("./tests/test_data/study_valid/")
-        tables = parser.clean_study(mock_cursor, schema, verbose, prefix=prefix)
+        with mock.patch.object(builtins, "input", lambda _: confirm):
```
mikix, line 95: fun

cumulus_library/study_parser.py
```diff
@@ -190,6 +202,13 @@ def clean_study(
             ):
                 view_table_list.remove(view_table)
         # We want to only show a progress bar if we are :not: printing SQL lines
+        if prefix:
+            print("The following views/tables were selected by prefix:")
+            for view_table in view_table_list:
+                print(f"  {view_table[0]}")
+            confirm = input("Remove these tables? (Y/N)")
```
mikix, line 209: Convention would say that this should be `y/N` since the no is default.

cumulus_library/study_parser.py
```diff
@@ -86,13 +86,25 @@ def get_study_prefix(self) -> Optional[str]:
...
         :returns: An array of sql files from the manifest, or None if not found.
         """
         sql_config = self._study_config.get("sql_config", {})
-        return sql_config.get("file_names", [])
+        sql_files = sql_config.get("file_names", [])
+        if continue_from:
+            match_found = False
+            for file in sql_files:
+                print(continue_from.replace(".sql", ""))
+                print(file.replace(".sql", ""))
```
mikix, line 100: :scissors:

cumulus_library/study_parser.py
```diff
@@ -86,13 +86,25 @@ def get_study_prefix(self) -> Optional[str]:
...
+                if continue_from.replace(".sql", "") == file.replace(".sql", ""):
+                    sql_files = sql_files[sql_files.index(file) :]
```
mikix, line 102: nit: you could save yourself an `index()` call by doing `for i, file in enumerate(sql_files):` above, but :shrug:

cumulus_library/study_parser.py
```diff
@@ -86,13 +86,25 @@ def get_study_prefix(self) -> Optional[str]:
...
+                    match_found = True
+                    break
+            if not match_found:
+                sys.exit(f"No tables matching '{continue_from}' found")
```
mikix, line 106: You ready for this Python nonsense? Just do `else:` and skip `match_found`. [For/else](https://docs.python.org/3/tutorial/controlflow.html#break-and-continue-statements-and-else-clauses-on-loops) lets you run code if the for loop never broke.

i.e.:
```
for file in sql_files:
    if continue_from.replace(".sql", "") == file.replace(".sql", ""):
        sql_files = sql_files[sql_files.index(file) :]
        break
else:
    sys.exit(f"No tables matching '{continue_from}' found")
```

Alternatively, you could also do `return sql_files[sql_files.index(file) :]` and unconditionally exit after the loop.

Or keep it! Whatever works for your sense of code cleanliness.

cumulus_library/cli_parser.py
```diff
@@ -141,6 +142,12 @@ def get_parser() -> argparse.ArgumentParser:
     add_verbose_argument(build)
     add_aws_config(build)
 
+    build.add_argument(
+        "--continue",
+        dest="continue_from",
+        help=(argparse.SUPPRESS),
```
mikix, line 148: Why the parens?

mikix (approved):

cumulus_library/cli_parser.py
dogversioning: copy pasted from elsewhere - why i did that in the first place, `¯\_(ツ)_/¯`

cumulus_library/study_parser.py
dogversioning: ooh, i dig it

cumulus_library/study_parser.py
dogversioning: taking the `for\else` approach, just in case there's other post loop stuff that might end up here in the future.

PR #144, opened 2023-11-22
feat: add new DuckDB backend for reading ndjson directly

mikix:
This adds some new arguments:
--db-type [athena,duckdb] (defaulting to athena)
--load-ndjson-dir DIR (tells DuckDB where to find source ndjsons)

A light abstraction layer has been added in databases.py to choose the correct backend based on the args.

Mostly the SQL is the same. Some light tweaks for standardization, plus some compatibility user-defined functions injected into duckdb allow both backends to work on the same SQL.

### Checklist


cumulus_library/cli_parser.py
```diff
@@ -81,6 +75,26 @@ def add_aws_config(parser: argparse.ArgumentParser) -> None:
...
+    group = parser.add_argument_group("Database config")
+    group.add_argument(
+        "--db-type",
+        help="Which database backend to use (default athena)",
+        choices=["athena", "duckdb"],
+        default="athena",
+    )
+    group.add_argument(
+        "--database",
+        # internally, we use PyAthena's terminology for this but the UX term is "database"
```
dogversioning, line 89: we should probably touch this comment since it's not 100% accurate anymore.

cumulus_library/databases.py
```diff
@@ -0,0 +1,213 @@
...
+import cumulus_fhir_support
+import duckdb
+import pandas
+import pyarrow
+import pyathena
+from pyathena.common import BaseCursor as AthenaCursor
+from pyathena.pandas.cursor import PandasCursor as AthenaPandasCursor
+
+
+# TODO: find a standard Protocol to use instead of homebrew
```
dogversioning, line 19: some cursory googling indicates there might not be one

cumulus_library/databases.py
```diff
@@ -0,0 +1,213 @@
...
+class DatabaseCursor(Protocol):
```
dogversioning, line 20: we might want to include `fetchmany` as well - i believe that is used by chunking implementations, which we should reserve the right to use.

cumulus_library/databases.py
```diff
@@ -0,0 +1,213 @@
...
+    @abc.abstractmethod
+    def cursor(self) -> DatabaseCursor:
+        """Returns a connection to the backing database"""
+
+    @abc.abstractmethod
+    def execute_as_pandas(self, sql: str) -> pandas.DataFrame:
+        """Returns a pandas.DataFrame version of the results from the provided SQL"""
+
+    # Consider: should we make the backend a context manager to close automatically?
+    #  Or do we like that an exception in the middle might not write changes out?
```
dogversioning, line 48: i think this is desirable behavior

cumulus_library/template_sql/show_tables.sql.jinja
dogversioning, line 1: for this and show views: we :could: change the template var to `database_name`?

dogversioning (commented):
Generally looks good, though I'm skipping the unit tests for now per your comments - tacitly approved but i'll look again when you're done.

cumulus_library/cli.py
```diff
@@ -297,6 +273,7 @@ def run_cli(args: Dict):
                 for target in args["target"]:
                     builder.export_study(study_dict[target], args["data_path"])
 
+        db_backend.close()
         # returning the builder for ease of unit testing
         return builder
```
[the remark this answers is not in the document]
mikix: Depends how we want to do it. In practice, this ensures that the duckdb backing file is flushed to disk.

So in a test, best approach might be to open that file back up using duckdb and do queries there to inspect the final state?

cumulus_library/template_sql/show_tables.sql.jinja
mikix: At this level of code, I think `schema_name` is fine -- at this (low) level, it _is_ called a schema. And at the user (high) level, we call it a database, which I think makes sense.

It's some of the in-between where maybe we could/should futz with it. Like, from CLI to DatabaseBackend, maybe the name "database" should be preserved, and the backend turns that into whatever is appropriate for it (schema_name or filename)? But I'm not fighting for that -- an improvement maybe, but not a necessary one.

cumulus_library/cli.py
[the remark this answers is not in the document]
dogversioning: oh i do like writing to a temp file and reading it back for validation, that feels clean - but we'd then want to get that cleaned up before each test run.

I do like the [pytest fixture](https://github.com/smart-on-fhir/cumulus-aggregator/blob/main/tests/conftest.py) approach for this over in the aggregator, but i know that's not your fave pattern - open to alternates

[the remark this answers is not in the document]
mikix: I'm only really adding one unit test in this PR (though it's a big end-to-end one that covers a lot of cases).

So I went with being unobtrusive and just did a little `with tempfile.TemporaryDirectory() as tmpdir:` and wrote a file there, which gets cleaned up automatically.

I can totes see a fixture making sense here to reduce boilerplate, but maybe if/once you use duckdb more extensively in the tests, it would be more obvious how best to write that.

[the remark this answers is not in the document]
mikix: (In my unit test I never actually manually open the file, I just export counts and make sure they match what I expect.)

cumulus_library/template_sql/utils.py
```diff
@@ -164,12 +164,12 @@ def _check_schema_if_exists(
         cursor.execute(query)
         schema_str = str(cursor.fetchone()[0])
         required_fields = [coding_element]
-        if allow_partial:
-            required_fields + ["code", "system", "display"]
+        if not allow_partial:
+            required_fields += ["code", "system", "display"]
```
mikix, line 168: This was a real bug where `allow_partial` was being ignored. I think I fixed it correctly, and then I added some unit testing for it.

tests/test_template_utils.py
```diff
@@ -16,53 +17,53 @@
...
     ).cursor()
     cursor.fetchone.side_effect = query_results
-    progress = Progress()
-    with get_progress_bar(transient=True) as progress:
-        task = progress.add_task(
-            "test_task",
-            total=3,
-        )
-        res = is_codeable_concept_populated(SCHEMA, "table", "base_col", cursor)
+    res = is_codeable_concept_array_populated(
```
mikix, line 66: It looked to me like the wrong method was being called here, so I changed it.

cumulus_library/studies/core/documentreference.sql
```diff
@@ -72,7 +72,7 @@ WITH powerset AS (
     WHERE
         d.encounter_ref = e.encounter_ref
         AND d.status = 'current'
-        AND d.docstatus IN (NULL, 'final', 'amended')
+        AND (d.docstatus IS NULL OR d.docstatus IN ('final', 'amended'))
```
mikix, line 75: This was a bug that happened to be caught by the new duckdb unit test. You apparently need to test for IS NULL, it doesn't work in a list like that. The new duckdb test has all these cases (as well as an `entered-in-error` case), so we're exercising this check.

cumulus_library/databases.py
mikix: :+1: removed this comment

cumulus_library/cli_parser.py
mikix: Edited to be more verbose about the situation.

cumulus_library/studies/core/documentreference.sql
dogversioning: oh great catch - these inline counts i haven't been paying much attention to yet.

docs/creating-studies.md
```diff
@@ -165,6 +165,57 @@ styling.
...
+where you can add edge cases.
+Cumulus Library has an optional database backend driven by local ndjson just for that!
+
+### Set up your ndjson
+
+You can grab fake
+[Synthea data](https://github.com/smart-on-fhir/sample-bulk-fhir-datasets),
+or use the result of actual bulk-export results from your EHR.
+
+Once you have that, run Cumulus ETL on your source ndjson with the
```
dogversioning, line 183: nit: this is the first time in this doc we mention the ETL, so we should probably link to the ETL section of the docsite here.

dogversioning (approved):
🦆

cumulus_library/template_sql/codeable_concept_denormalize.sql.jinja
```diff
@@ -54,6 +54,7 @@ CREATE TABLE {{ target_table }} AS (
             ROW_NUMBER()
                 OVER (
                     PARTITION BY id
+                    ORDER BY priority ASC
```
mikix, line 57: This was a bug found during testing -- without this order-by, you can get inconsistent results run-to-run

PR #150, opened 2023-11-30
PSM table generation

dogversioning:
This PR attempts to add in the research done as part of the suicidality studies around propensity score matching as a standard statistical tool in the library.

This attempts sole to deal with the PSM logistics, but not the cli/study logistics. So here's what it's trying to do:
- Provide a new table builder extension for PSM
- Provide PSM specific jinja templates
- Creates a toml input format for configuring PSM jobs
- Ancilarily provides a conftest fixture for running tests with duckdb

It does not attempt to do the following, for cognitive load reasons, which will follow on later:
- run PSM from a manifest
- metadata table peristance for managing multiple PSM tables
- cli commands around optionally running/cleaning PSM tables
- PSM date filtering on cohorts, which needs some more definition

### Checklist


cumulus_library/statistics/psm.py
```diff
@@ -0,0 +1,247 @@
...
+    """TableBuilder for creating PSM tables"""
+
+    display_text = "Building PSM tables..."
+
+    def __init__(self, toml_config_path: str):
+        """Loads PSM job details from a PSM configuration file"""
+        with open(toml_config_path, encoding="UTF-8") as file:
+            toml_config = toml.load(file)
+        self.config = PsmConfig(
+            classification_json=f"{PosixPath(toml_config_path).parent}/{toml_config['classification_json']}",
```
mikix, line 60: nit: Some fields here are directly looked up and some are optional with a default. The user experience if some of these fields are missing might be kinda obscure? They'd get some KeyError I assume. May be worth wrapping this with an `except KeyError` and give a nice little user error. But could also be a future improvement.

cumulus_library/statistics/psm.py
```diff
@@ -0,0 +1,247 @@
...
+            primary_ref=toml_config["primary_ref"],
+            dependent_variable=toml_config["dependent_variable"],
+            pos_sample_size=toml_config["pos_sample_size"],
+            neg_sample_size=toml_config["neg_sample_size"],
+            join_cols_by_table=toml_config.get("join_cols_by_table", {}),
+            count_ref=toml_config.get("count_ref", None),
+            count_table=toml_config.get("count_table", None),
+            seed=toml_config.get("seed", 123),
+        )
+        super().__init__()
```
mikix, line 73: nit: hah, a trailing super() call! That feels odd to me, emotionally. Feels like information passing from subclass to parent in a surprising way. But nothing wrong with it! Just an unusual pattern to my eyeballs.

cumulus_library/statistics/psm.py
```diff
@@ -0,0 +1,247 @@
...
+
+    def _get_symptoms_dict(self, path: str) -> dict:
+        """convenience function for loading symptoms dictionaries from a json file"""
+        with open(path) as f:
```
mikix, line 77: I'll be your pylint call and say you're missing an `encoding` arg.

cumulus_library/statistics/psm.py
```diff
@@ -0,0 +1,247 @@
...
+        :param is_positive: defines the value to be used for your filtering column
+        """
+        df = cursor.execute(query).as_pandas()
+        df = (
+            df.sort_values(by=[self.config.primary_ref])
+            .reset_index()
+            .drop("index", axis=1)
+        )
+        df = (
+            # TODO: remove replace behavior after increasing data sample size
```
mikix, line 109: Is this still valid? As a comment to a non-Matt, I'm not sure it's clear to me when that would be true / why replace is bad. I would currently guess this is about unit tests. (Same comment above the call to `knn_matched` later)

cumulus_library/statistics/psm.py
```diff
@@ -0,0 +1,247 @@
...
+            This should generally be >= 20.
+        :param dependent_variable: the name to use for your filtering column
+        :param is_positive: defines the value to be used for your filtering column
+        """
+        df = cursor.execute(query).as_pandas()
+        df = (
+            df.sort_values(by=[self.config.primary_ref])
+            .reset_index()
+            .drop("index", axis=1)
+        )
```
mikix, line 107: Why do this stuff before sampling, since we do it right after sampling?

cumulus_library/statistics/psm.py
```diff
@@ -0,0 +1,247 @@
...
+
+        df[dependent_variable] = is_positive
+        return df
+
+    def _create_covariate_table(self, cursor: DatabaseCursor, schema: str):
+        """Creates a covariate table from the loaded toml config"""
+        # checks for primary & link ref being the same
+        source_refs = list(
+            {self.config.primary_ref, self.config.count_ref} - set([None])
```
mikix, line 123: I think pylint would recommend you write that as `{None}`

cumulus_library/statistics/psm.py
```diff
@@ -0,0 +1,247 @@
...
+            schema,
+            neg_query,
+            self.config.neg_sample_size,
+            self.config.dependent_variable,
+            0,
+        )
+        cohort = pandas.concat([pos, neg])
+
+        # Replace table (if it exists)
+        # TODO - replace with timestamp prepended table
```
mikix, line 151: Just pointing this todo out to confirm it's an intentional post-PR comment, not an internal-PR comment.

cumulus_library/statistics/psm.py
```diff
@@ -0,0 +1,247 @@
...
+            neg_source_table=self.config.neg_source_table,
+            primary_ref=self.config.primary_ref,
+            dependent_variable=self.config.dependent_variable,
+            join_cols_by_table=self.config.join_cols_by_table,
+            count_ref=self.config.count_ref,
+            count_table=self.config.count_table,
+        )
+        self.queries.append(dataset_query)
+
+    def generate_psm_analysis(self, cursor: object, schema: str):
```
mikix, line 177: nit: `cursor: DatabaseCursor` unless you used `object` because you knowingly call a method not in the protocol and you didn't want to see a squiggly cursor in your IDE. In which case, let's sort that piece out (from discussion above).

cumulus_library/statistics/psm.py
```diff
@@ -0,0 +1,247 @@
...
+        columns = []
+        if self.config.join_cols_by_table is not None:
+            for table_key in self.config.join_cols_by_table:
+                for column in self.config.join_cols_by_table[table_key][
+                    "included_cols"
+                ]:
+                    if len(column) == 2:
+                        columns.append(column[1])
+                    else:
+                        columns.append(column[0])
```
mikix, line 197: I'm not :100: on what this block is doing, but could you replace this with `column[-1]`? You may be trying to be future proof, in case the user provides 3 columns, in which case maybe the first check should be `>= 2`

cumulus_library/statistics/psm.py
```diff
@@ -0,0 +1,247 @@
...
+        # instance_count present but unused for PSM if table contains a count_ref input
+        # (it's intended for manual review)
+        df = df.drop(columns="instance_count", errors="ignore")
+
+        columns = []
+        if self.config.join_cols_by_table is not None:
+            for table_key in self.config.join_cols_by_table:
+                for column in self.config.join_cols_by_table[table_key][
+                    "included_cols"
+                ]:
```
mikix, line 193: Could be slightly more readable (to my eyes) if you did the lookup during iteration:
```
            for table_config in self.config.join_cols_by_table.values():
                for column in table_config["included_cols"]:
```

cumulus_library/statistics/psm.py
```diff
@@ -0,0 +1,247 @@
...
+                    if len(column) == 2:
+                        columns.append(column[1])
+                    else:
+                        columns.append(column[0])
+
+        for column in columns:
+            encoded_df = pandas.get_dummies(df[column])
+            df = pandas.concat([df, encoded_df], axis=1)
+            df = df.drop(column, axis=1)
+        df = df.reset_index()
```
mikix, line 203: A block like this could use a comment about why it's doing what it's doing. My rough attempt: you're replacing each column with a dummy version of that column. But even after reading the pandas docs on `get_dummies`, I'm not :100: on what that means.

cumulus_library/statistics/psm.py
```diff
@@ -0,0 +1,247 @@
...
+            psm = PsmPy(
+                df,
+                treatment=self.config.dependent_variable,
+                indx=self.config.primary_ref,
+                exclude=[],
+            )
+            # This function populates the psm.predicted_data element, which is required
+            # for things like the knn_matched() function call
+            psm.logistic_ps(balance=True)
+            print(psm.predicted_data)
```
mikix, line 215: :scissors:

cumulus_library/statistics/psm.py
```diff
@@ -0,0 +1,247 @@
...
+            # This function populates the psm.df_matched element
+            # TODO: flip replacement to false after increasing sample data size
+            psm.knn_matched(
+                matcher="propensity_logit",
+                replacement=True,
+                caliper=None,
+                drop_unmatched=True,
+            )
+            print(psm.df_matched)
```
mikix, line 224: :scissors:

cumulus_library/statistics/psm.py
```diff
@@ -0,0 +1,247 @@
...
+        except ZeroDivisionError:
+            print(
+                "Encountered a divide by zero error during statistical graph generation. Try increasing your sample size."
+            )
```
mikix, line 228: Just a print and not an exit, for these two exceptions? (nit: ideally would pipe to stderr instead of stdout)

cumulus_library/statistics/psm.py
```diff
@@ -0,0 +1,247 @@
...
+    def execute_queries(
+        self,
+        cursor: object,
+        schema: str,
+        verbose: bool,
+        drop_table: bool = False,
+    ):
+        super().execute_queries(cursor, schema, verbose, drop_table)
+        self.comment_queries()
+        self.write_queries()
```
mikix, line 246: Isn't this a debugging step that writes the queries to disk? A :scissors: case?

Should `write_queries` just call `comment_queries` itself?

cumulus_library/statistics/psm.sql
```diff
@@ -0,0 +1,52 @@
...
+-- ###########################################################
+CREATE TABLE core__psm_encounter_covariate AS (
+    SELECT 
+        DISTINCT sample_cohort."encounter_ref",
+        sample_cohort."example_diagnosis",
+        (
+            SELECT COUNT( DISTINCT encounter_ref )
+            FROM "core__condition"
+            WHERE sample_cohort."subject_ref" = "core__condition"."subject_ref"
+            --AND sample_cohort.enc_end_date >= "core__condition".recordeddate
```
mikix, line 40: Is this (and the other line below) manually commented out? :scissors:?

cumulus_library/template_sql/statistics/psm_create_covariate_table.sql.jinja
mikix, line 1: This file felt fairly low on readability -- could more newlines/indents and/or a few comments help? Or separate logic into macros at top of file, which isn't a thing we normally do in these. Or maybe as simple as using a few jinja variables judiciously, like saving `join_cols_by_table[key]["join_id"]` as a variable -- that was the section that kind of broke my eyeballs.

I will say, it's more readable to me than it would have been last month - I'm glad I'm writing some of these for the quality metric study.

cumulus_mhg_dev_db
mikix, line 1: Is this supposed to be here?

cumulus_library/template_sql/statistics/psm_templates.py
```diff
@@ -0,0 +1,87 @@
...
+    It is expected that, if supplying optional parameters, all of them must be set
+
+    :param columns: a list of ids to request
+    :param source_table: the table to retrieve ids from
+    :param join_id: the id column to use for joining. Expected to exist in both source and filter tables
+    :param filter_table: a table containing ids you want to exclude
+    """
+    if (join_id is None and filter_table is not None) or (
+        join_id is not None and filter_table is None
+    ):
```
mikix, line 30: nit: Do you want to allow empty strings here? If not, this may be more readable with straight boolean logic:
```
if (join_id and not filter_table) or (not join_id and filter_table):
```
or
```
if bool(join_id) != bool(filter_table):
```

(same pattern below with count table)

tests/conftest.py
```diff
@@ -0,0 +1,44 @@
...
+                "db_type": "duckdb",
+                "schema_name": f"{tmpdir}/duck.db",
+                "load_ndjson_dir": data_dir,
+            }
+        )
+        yield db
+
+
+@pytest.fixture
+def mock_db_core():
```
mikix, line 29: Can you not re-use `mock_db` in this method, to avoid re-iterating the db setup?

mikix (approved):

cumulus_library/statistics/psm.py
dogversioning: good catch - i believe this is a holdover from when i started and that PR hadn't been merged yet.

cumulus_library/statistics/psm.py
dogversioning: yeah this is future proofing and i think i have this throw an error if the length is not 1 or 2.

cumulus_library/statistics/psm.py
dogversioning: yeah i'll add something - this is converting to a 1-hot encoding for all values of that column, basically pivoting the column values to new column headers.

cumulus_library/statistics/psm.py
dogversioning: or, jesus, raise an error or something reasonable

cumulus_library/statistics/psm.py
dogversioning: yeah this is just a debugging step, will trim.

i could go either way on making this embedded or not - i think it depends on the complexity of the queries. feel free to talk me into just doing it always.

cumulus_library/statistics/psm.sql
dogversioning: got those, probably while you were looking at this.

cumulus_library/template_sql/ctas.sql.jinja
```diff
@@ -2,21 +2,20 @@ CREATE TABLE "{{ schema_name }}"."{{ table_name }}" AS (
     SELECT * FROM (
         VALUES
         {%- for row in dataset %}
-        ((
+        (
```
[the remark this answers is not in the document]
dogversioning: i want this do be done so i can spend a day or two just actually running queries rather than testing to see if they were passed to a cursor.

cumulus_library/template_sql/statistics/psm_create_covariate_table.sql.jinja
dogversioning: lemme try moving things into macros maybe?

our sqlfluff rules are optimized for minimal whitespace padding by request - we could push back on this? it does give me a headache.

pyproject.toml
```diff
@@ -7,7 +7,8 @@ dependencies = [
     "duckdb >= 0.9",
     "fhirclient >= 4.1",
     "Jinja2 > 3",
-    "pandas <3, >=2.1.1",
+    "pandas <3, >=2.1.3",
+    "psmpy <1, >=0.3.13",
```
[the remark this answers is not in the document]
dogversioning: This is a preponderance of caution and an inability to look at the change history in github.

tests/conftest.py
dogversioning: oh probably!

tests/test_data/psm/psm_config.toml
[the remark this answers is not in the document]
dogversioning: on the one hand, i hear you.

on the other hand, there is no other place for these to live right now - there is no instance in core where this module would be used. I have been thinking about a `stats_example` study for this any any future things, but right now, for documentation purposes, other comments point here. 

When the other half of the workflow is done, maybe i can move this to a markdown doc or something? but i want to keep it here, for the time being.

cumulus_library/template_sql/statistics/psm_templates.py
dogversioning: ok, can do

cumulus_library/statistics/psm.py
mikix: Ah maybe I'm thinking of the pep8 checker `pycodestyle` then. Or we have different pylint configs. I have fixed many a missing `encoding` arg.

cumulus_library/statistics/psm.py
dogversioning: ok, moved it up to the head with no adverse impacts.

tests/conftest.py
```diff
@@ -0,0 +1,44 @@
...
+
+from cumulus_library.cli import StudyBuilder
+from cumulus_library.databases import create_db_backend
+
+
+@pytest.fixture
+def mock_db():
+    """Provides a DuckDatabaseBackend for local testing"""
+    data_dir = f"{Path(__file__).parent}/test_data/duckdb_data"
+    with tempfile.TemporaryDirectory() as tmpdir:
```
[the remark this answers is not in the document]
dogversioning: ok i tried this and the tempfiles were not behaving quite as gracefully w.r.t. yield/cleanup as tmpdirs, and it also provides a 'write out some logs while you're at it' option, so i'm electing to leave this as is.

cumulus_library/base_table_builder.py
```diff
@@ -33,7 +33,12 @@ def prepare_queries(self, cursor: object, schema: str):
         """
         raise NotImplementedError
 
-    @final
+    # 🚨🚨🚨 WARNING: in 99% of cases, subclasses should *not* re-implement 🚨🚨🚨
+    # 🚨🚨🚨 execute_queries.                                               🚨🚨🚨
+
+    # If you know what you are doing, you can attempt to override it, but it is
+    # strongly recommended you invoke this as is via a super() call, and then
+    # run code before or after that as makes sense for your use case.
```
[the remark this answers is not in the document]
dogversioning: Nah I think this is actually a pretty good idea, so i've switched this back to final and added a stub.

cumulus_library/template_sql/ctas.sql.jinja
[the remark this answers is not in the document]
dogversioning: touched insert_into as well

cumulus_library/statistics/psm.py
dogversioning: ok, debugging artifact confirmed, removed

cumulus_library/statistics/psm.py
dogversioning: updated with some explanitory text

cumulus_library/template_sql/statistics/psm_create_covariate_table.sql.jinja
dogversioning: ok, moved some stuff into macros, let me know what you think about that in this case/generally - I think? you can do centralized macros, this comma delineation nonsense feels like candidate number one for that.

tests/conftest.py
dogversioning: ok, chained these mocks together

tests/test_data/psm/psm_config.toml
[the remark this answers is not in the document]
dogversioning: decided to start stubbing out the docs, so I moved the commentary there and cleaned this up.

PR #154, opened 2023-12-07
Cutover to duckdb for unit tests

dogversioning:
This PR does the following:
- Removes athena based cursor mocking in favor of duckDB for CLI testing
  - athena is still covered by our regression environment
- Adds a simple test for the core study (and validation data)
- Some light tweaks to some of the mock studies to make more sense being run in a real db
- Clears env for the duckdb test to prevent overriding behavior

### Checklist


tests/test_cli.py
```diff
@@ -117,117 +157,112 @@ def test_count_builder_mapping(
...
-        assert mock_write.called
+def test_cli_creates_study(tmp_path):
+    cli.main(cli_args=["create", f"{tmp_path}/studydir/"])
+    with open(
+        "./cumulus_library/studies/template/manifest.toml", encoding="UTF-8"
+    ) as file:
+        source = toml.load(file)
+    with open(f"{tmp_path}/studydir/manifest.toml", encoding="UTF-8") as file:
+        target = toml.load(file)
+    print(source, target)
```
mikix, line 264: :scissors:

tests/test_core.py
```diff
@@ -0,0 +1,30 @@
...
+    ],
+)
+def test_core_tables(mock_db_core, tmp_path, table):
+    cursor = mock_db_core.cursor()
+    table_rows = cursor.execute(f"SELECT * FROM {table}").fetchall()
+    # For regenerating data if needed
+    # with open(f'./tests/test_data/core/{table}.txt','wt', encoding="UTF-8") as f:
+    #    f.write(str(table_rows))
+    with open(f"./tests/test_data/core/{table}.txt", "r", encoding="UTF-8") as f:
+        ref_table = eval(f.read())
```
mikix, line 27: nit: What about comparing the csv output? That loses some metadata like typing... But would test the export and is a more familiar human-readable-file-format. This is fine though.

mikix (approved):

tests/test_core.py
dogversioning: lemme sleep on it - my first instinct was no, because:
- this is testing more tables than the export - it might be easier to narrow down an issue by also running the base tables
- Export as a mechanism is already covered in the cli interface test

i :could: see having a test that is like 'get every table that has `__count_` in it and verify they are all in the export list of the manifest?

I went back and forth on making things more readable, and I could go back to trying that - mostly the issue is datetime serialization - but i could write out/read in one row at a time?

mikix: Hah nice :smile:

PR #158, opened 2023-12-13
Use time-elapsed for progress rather than an estimate of remaining

mikix:
### Checklist


dogversioning (approved):
i think we talked about doing this but didn't make an issue for it - but makes sense for those Big Queries

PR #163, opened 2024-01-09
Table infrastructure, condition refactor

dogversioning:
- Adds a schema parser class to databases
  - Per discussion, it's a simple first pass - it is not really database aware.
  - lots of *args **kwargs updates to support optionally passing a parser to a builder
- Moved condition table (& some prerequisite queries) to a table builder
  - Condition table should now inspect the found schema and null columns it cant find
- moved count_condition_month table to core counts builder
- some light updates to medication counts, which wasn't actually being properly tested before

### Checklist


cumulus_library/__init__.py
```diff
@@ -1,2 +1,2 @@
 """Package metadata"""
-__version__ = "1.5.0"
+__version__ = "2.0.0"
```
mikix, line 2: Exciting!

cumulus_library/databases.py
```diff
@@ -41,6 +41,51 @@ def fetchall(self) -> Optional[list[list]]:
         pass
 
 
+class DatabaseParser(abc.ABC):
+    """Parses information_schema results from a database"""
+
+    def _parse_found_schema(self, expected: dict[dict[list]], schema: list[tuple]):
+        """Checks for presence of field for each column in a table
+
+        This naieve method is a first pass, ignoring complexities of differing
```
mikix, line 50: typo: naieve

cumulus_library/databases.py
```diff
@@ -41,6 +41,51 @@ def fetchall(self) -> Optional[list[list]]:
...
+        database variable types, just iterating through looking for column names.
+
+        TODO: on a per database instance, consider a more nuanced approach
+        if needed
```
mikix, line 54: We both suspect it will be at some point yeah? I'm not hopeful we can band-aid this too much, but I think this sets us on a good path.

cumulus_library/databases.py
```diff
@@ -41,6 +41,51 @@ def fetchall(self) -> Optional[list[list]]:
...
+                        else:
+                            output[column][field] = True
+                            col_schema = col_schema[1]
+                # otherwise this field is a bare type
+                else:
+                    output[column] = True
+            else:
+                for field in expected[column]:
+                    output[column][field] = False
+        return output
```
mikix, line 76: I assume this was tested on Athena at some point yeah?

nit: It's not the prettiest little function :smile: Maybe some example input in the docstring or comments would help it be more readable - as you look at it, you can scan above to see, "ah yeah, this would parse that bit" or something.

cumulus_library/studies/core/count_core.py
```diff
@@ -1,9 +1,9 @@
 from typing import List
 from pathlib import Path
-from cumulus_library.statistics.counts import CountsBuilder
+import cumulus_library.statistics.counts as counts
```
mikix, line 3: nit: this might feel more natural as `from cumulus_library.statistics import counts`

cumulus_library/studies/core/builder_prereq_tables.py
```diff
@@ -0,0 +1,23 @@
...
+from cumulus_library.helper import get_progress_bar, query_console_output
+
+
+class CorePrereqTableBuilder(base_table_builder.BaseTableBuilder):
+    display_text = "Creating core prerequisite tables..."
+
+    def prepare_queries(self, cursor: object, schema: str, *args, **kwargs):
+        dir_path = pathlib.Path(__file__).resolve().parents[0]
+        prereq_sql = [
+            "version.sql",
```
mikix, line 15: I'm sure this makes sense, but what's the vibe behind moving these files to a prereq table builder?

cumulus_library/study_parser.py
```diff
@@ -315,7 +314,7 @@ def _execute_drop_queries(
             helper.query_console_output(verbose, drop_view_table, progress, task)
 
     def _load_and_execute_builder(
-        self, filename, cursor, schema, verbose, drop_table=False
+        self, filename, cursor, schema, verbose, drop_table=False, parser=None
```
mikix, line 317: Could add type hinting for the parser field too, since you're already importing database classes.

cumulus_library/studies/core/core_templates/core_utils.jinja
```diff
@@ -0,0 +1,38 @@
...
+        {{- comma_delineate(loop) }}
+        {%- endfor %}
+{%- endmacro %}
+
+
+-- gets a column, if present, from a basic SQL primitive col, or a null otherwise
+{% macro basic_nullable_cols(table, alias, targets, schema) %}
+        {%- for col in targets %}
+        {%- if schema[table][col] %}
+        {{ alias }}.{{ col.lower() }}
```
mikix, line 20: nit: I'm pretty sure you don't need the `.lower()` here, but it's harmless (re: us talking about case in actual SQL vs string comparison)

cumulus_library/studies/core/core_templates/condition.sql.jinja
```diff
@@ -0,0 +1,51 @@
...
+            'c',
+            ['id','category','verificationStatus','clinicalStatus'], schema) 
+        }},
+        {{- utils.depth_one_nullable_cols(
+                'condition',
+                'c',
+                [
+                    ('subject', 'reference', 'subject_ref'),
+                    ('encounter', 'reference', 'encounter_ref')
+                ],
```
mikix, line 18: nit: there is a world where you can combine basic_nullable_cols and depth_one_nullable_cols into a generic nullable_cols method.

Accept a list of str | tuple (jinja has type inspection, so your method could tell the difference). If a string, split it by `.` and use that as the depth indicators. If a tuple, take two strings, the left is the field path, the right is the new alias.

So together, these two calls could look like:

```
        {{- utils.nullable_cols(
                'condition',
                'c',
                [
                    'id', 'category', 'verificationStatus', 'clinicalStatus',
                    ('subject.reference', 'subject_ref'),
                    ('encounter.reference', 'encounter_ref'),
                ], 
          )}}
```

(I'm only proposing that because I can already foresee the need for `depth_two_nullable_cols` and it's feeling a bit wobbly.)

cumulus_library/studies/core/core_templates/condition.sql.jinja
```diff
@@ -0,0 +1,51 @@
+{% import 'core_utils.jinja' as utils %}
+
+CREATE TABLE core__condition AS
+WITH temp_condition AS (
+    SELECT
+        concat('Condition/', c.id) AS condition_ref,
+        {{- utils.basic_nullable_cols(
```
mikix, line 7: This should never fail and return null though yeah? Cumulus ETL guarantees all toplevel columns will be there, I would have thought.

cumulus_library/template_sql/statistics/count.sql.jinja
```diff
@@ -21,24 +31,38 @@ CREATE TABLE {{ table_name }} AS (
...
         {%- if fhir_resource=='document' and filter_resource %}
-        WHERE (status = 'current')
+        WHERE (s.status = 'current')
         AND (d.docStatus IS null OR d.docStatus IN ('final', 'amended'))
         {%- elif fhir_resource=='encounter' and filter_resource %}
-        WHERE status = 'finished'
+        WHERE s.status = 'finished'
         {%- elif fhir_resource=='observation' and filter_resource %}
-        WHERE (status = 'final' OR status= 'amended')
+        WHERE (s.status = 'final' OR status= 'amended')
```
mikix, line 65: Second status could use the `s.` too

cumulus_library/template_sql/statistics/count.sql.jinja
```diff
@@ -48,8 +72,14 @@ CREATE TABLE {{ table_name }} AS (
             {%- if secondary %}
             {{ secondary }}_ref,
             {%- endif -%}
+            {%- if fhir_resource=='condition' %}
+            coalesce(cast(cond_code_display AS varchar), 'missing-or-null') AS cond_code_display,
```
mikix, line 76: I made this little helper method, because I didn't like reproducing a static string all over, which might get typo'd. It doesn't save much, though, so I get if you like the cleanliness of just doing it directly. But here is my macro:

```
{% macro coalesce_missing() -%}
    {% if varargs %}
        {% set arg = varargs[0] %}
    {% else %}
        {% set arg = caller() %}
    {% endif %}

    COALESCE(
        {{ arg }},
        'missing-or-null'
    )
{%- endmacro %}
```

Can be used like so:
```
 {{ coalesce_missing('field.subfield') }} AS cond_code_display,
 {% call coalesce_missing() %}
    cast(cond_code_display AS varchar)
 {% endcall %} AS cond_code_display
```

cumulus_library/template_sql/statistics/count.sql.jinja
```diff
@@ -66,16 +96,22 @@ CREATE TABLE {{ table_name }} AS (
...
-            cube( 
+            cube(
                 {%- for col in table_cols %}
-                "{{ col }}"
+                "{{ column_or_alias(col) }}"
                 {{- comma_delineate(loop) }}
                 {%- endfor %}
+                {%- if fhir_resource=='condition' -%},
+                cond_code_display
+                {% endif %}
```
mikix, line 114: Yo there's a lot of condition specific code now :smile:  - This is stuff that was missing before and Condition is just a pain in the butt, I'm gathering, rather than the result of all your schema work.

cumulus_library/template_sql/utils.py
```diff
@@ -160,7 +160,6 @@ def _check_schema_if_exists(
         if cursor.fetchone() is None:
             return False
 
-        query = get_column_datatype_query(schema, table, base_col)
         cursor.execute(query)
```
mikix, line 163: nit: you could just save the result of the previous query rather than re-querying now, yeah?

pyproject.toml
```diff
@@ -14,6 +14,7 @@ dependencies = [
     "requests >= 2.28",
     "rich >= 13.2",
     "sqlfluff >= 2.3.4",
+    "sqlparse >0.4",
```
mikix, line 17: Where is this being used? I must have missed it and GitHub UI is making it annoying to re-open every file.

tests/test_counts_templates.py
```diff
@@ -147,4 +169,6 @@
 )
 def test_count_query(expected, kwargs):
     query = get_count_query("test_table", "test_source", ["age", "sex"], **kwargs)
+    with open("output.sql", "w") as f:
+        f.write(query)
```
mikix, line 173: :scissors:?

mikix (approved):
Nice - looks much more pleasant to work with, once you have the schema in hand.

cumulus_library/databases.py
dogversioning: french loan words, man

cumulus_library/studies/core/core_templates/core_utils.jinja
dogversioning: i'd rather make any information_schema lookups downstream of this default to all lowercase, as that's best practice when you're not trying to maintain an external data model's semantics.

cumulus_library/studies/core/core_templates/condition.sql.jinja
dogversioning: so i had already planned for a two depth call in the future, thinking that separation would be easier maintenance wise - but if you're feeling that type inference is ok at this complexity, you can talk me out of it.

cumulus_library/studies/core/core_templates/condition.sql.jinja
dogversioning: this is probably paranoid, but might be useful if we hit a bug or something - the implicit goal here is to make library runs less failure prone, so some 'lets verify even though we're sure' is in spirit.

Although - and I just created #164 to remember to track this - we should track cases where we hit one of these and report it back in some fashion.

cumulus_library/template_sql/statistics/count.sql.jinja
dogversioning: I think the macro idea is smart, especially as we're about to change this based on some of the system discussion I was mentioning earlier - would be nice to have one point of entry for this.

That said - I might elect to do that in the next PR so I have a better idea of the expected scopes - lemme sleep on this one and make a call one way or another.

cumulus_library/template_sql/statistics/count.sql.jinja
dogversioning: I :think: that it still makes sense to keep all these running through the same template since 80% is boilerplate, but i think there's going to be more of this kind of thing as we get the other count types out of static sql and into builders, though at this point i think it should mostly be just elif statements against these blocks.

cumulus_library/template_sql/utils.py
dogversioning: i think this is a debugging cursor state artifact

pyproject.toml
dogversioning: this is being used in the prereq builder. specifically, since we have more than one query per file in some of those cases, this is a nice convenience for splitting those into individual queries for athena queueing rather than writing bespoke logic from scratch.

tests/test_counts_templates.py
dogversioning: 'how to identify which thing i was debugging right before i opened a pr' for 100, alex

cumulus_library/studies/core/core_templates/condition.sql.jinja
mikix: I don't feel strongly about this - I just have a personal bias towards high-level API vs low-level API. :smile: but the tradeoffs around that are squarely on you, so whatever you like.

Speaking of, I might want to use these schema concepts from the quality study. In my head, using stuff from Python feels OK, but not jinja utilities (so I'd have to recreate these macros myself). Does that vibe with you? Or am I not even supposed to be calling into Python... - a future task could be to clarify which API is public and which isn't (maybe using my `__init__.py` vibes of putting public API in there).

cumulus_library/template_sql/utils.py
mikix: With the second `execute()` call, the second `fetchone()` is getting the _second_ line of results. I assume that's not what you want? You might want to save the results of the first `fetchone()` and re-use it.

cumulus_library/studies/core/builder_prereq_tables.py
```diff
@@ -0,0 +1,23 @@
...
+            "setup.sql",
+            "fhir_lookup_tables.sql",
+            "fhir_mapping_tables.sql",
+        ]
+        for sql_file in prereq_sql:
+            with open(dir_path / sql_file) as file:
+                queries = sqlparse.split(file.read())
```
mikix, line 22: Ah OK so this is the `sqlparse` use. This logic of "read multiple queries from a file" is something I wouldn't mind being added as syntactic sugar for jinja templates too. I had to split some files up in the quality study to accommodate the one-query-one-file requirement

cumulus_library/template_sql/utils.py
dogversioning: i think this was actually a mistake to have removed the second query in the first place - after digging in the tests for this file are still using pyathena mocks and i think that masked this issue. Should have thought about this one a bit more when i saw it the first time around - this was more of a rabbit hole picking it up fresh in the morning than i'd thought.

cumulus_library/studies/core/core_templates/condition.sql.jinja
dogversioning: so i'm not quite there yet (especially not in one of these big PRs), but I like the idea of some sort of global location for jinja macros that could be imported in other studies. I think the trick is deciding what is study specific, and what is truly global. 

In my head, this was core specific, since I'm not expecting study authors to routinely need to engage with the nuances of nested FHIR elements. I think the quality study is an exception to the rule since you're interacting with base FHIR there - so, maybe these could be global, but in like a `fhir_utils` type bucket?

mikix: I don't want to burden these early macros with future thinking - I'm happy to re-implement / copy some of these schema macros.

The only thing I'd ask is that moving some of the schema python stuff around would be considered a breaking change. Is that a fair burden?

dogversioning: Yeah I think so - I want to leave this comment open specifically because i might make that as part of the breaking changes this is incurring once the other tables are cut over.

PR #166, opened 2024-01-23
Refactored remaining core tables

dogversioning:
- Updates to core templates for handling date fields
- Moves all base core tables to builders
- Moves core-related templates from base templates to core_templates
- moves non-template reading functions from template_sql/templates to template_sql/utils

### Checklist


tests/test_data/core/core__count_documentreference_month.txt
```diff
@@ -1,18 +1,18 @@
-(22, datetime.date(2018, 6, 1), 'ambulatory', None)
```
dogversioning, line 1: Summary of changes here:
- Going through single entrypoint for counts switched datetime to str
- Reordered to accomodate template structure due to all the secondary/tertiary conditions

tests/test_data/core/core__count_encounter_month.txt
```diff
@@ -1,57 +1,57 @@
-(10, '2018-07-01', 'ambulatory', None, 'male', None, 'not hispanic or latino')
```
dogversioning, line 1: This was regenned more than once during development, but should be exactly the same, just sorted differently. I should probably fix this quirk of the reference data export mechanism

tests/test_data/core/core__count_encounter_month_missing_data.txt
```diff
@@ -1,64 +1,64 @@
+(10, '2018-07-01', 'cumulus__missing-or-null', None, 'male', None, 'not hispanic or latino')
```
dogversioning, line 1: this is just a change of the null representation string.

tests/test_data/core/core__count_medicationrequest_month.txt
```diff
@@ -1,12 +1,12 @@
-(11, 'stopped', None, datetime.date(2018, 6, 1), None)
```
dogversioning, line 1: this is only datetime->str conversion.

tests/test_data/core/core__count_observation_lab_month.txt
```diff
@@ -1,14 +1,14 @@
 (10, None, '34533-0', 'Urine smell ammoniacal (finding)', None)
-(10, None, None, 'Brown color (qualifier value)', 'AMB')
-(10, None, '5778-6', 'Brown color (qualifier value)', 'AMB')
+(10, None, None, 'Brown color (qualifier value)', 'ambulatory')
+(10, None, '5778-6', 'Brown color (qualifier value)', 'ambulatory')
```
dogversioning, line 3: This was consolodating on the code display value versus the actual code value - this i don't think is being looked at often, but this was one of the older vintage of count table we had.

tests/test_data/core/core__observation.txt
```diff
@@ -1,20 +1,20 @@
-([{'id': None, 'coding': [{'id': None, 'code': 'laboratory', 'display': 'Laboratory', 'system': 'http://terminology.hl7.org/CodeSystem/observation-category', 'userSelected': None, 'version': None}], 'text': None}], None, 'final', {'id': None, 'code': '34533-0', 'display': 'Odor of Urine', 'system': 'http://loinc.org', 'userSelected': None, 'version': None}, None, None, None, {'id': None, 'coding': [{'id': None, 'code': '167248002', 'display': 'Urine smell ammoniacal (finding)', 'system': 'http://snomed.info/sct', 'userSelected': None, 'version': None}], 'text': 'Urine smell ammoniacal (finding)'}, datetime.date(2018, 6, 2), datetime.date(2018, 5, 28), datetime.date(2018, 6, 1), datetime.date(2018, 1, 1), 'Patient/8877ef1f-7cd7-3242-d7f0-73cf3f7165f4', 'Encounter/299b6495-3fe7-8db3-c494-6e1ce8b7986d', '05e6184c-f3ef-017b-ce21-599d375020b3', 'Observation/05e6184c-f3ef-017b-ce21-599d375020b3')
```
dogversioning, line 1: Since I did a lot of unnesting on this one, i also took this opporitunity to move the ID to the first column of the table. But there's enough going on here that a detailed comparison is probably not warranted

tests/test_data/core/core__patient.txt
```diff
@@ -1,50 +1,50 @@
-('female', datetime.date(2002, 1, 1), 22, '672', '0734762a-9db6-22fc-b595-c3e472bb2a9a', 'Patient/0734762a-9db6-22fc-b595-c3e472bb2a9a', 'white', 'not hispanic or latino')
```
dogversioning, line 1: This was a case where all the patient's birthdates were getting set to january first! Yeesh!

tests/test_data/core/core__count_patient.txt
```diff
@@ -1,12 +1,12 @@
-(17, None, 'male', 'white', 'not hispanic or latino')
```
dogversioning, line 1: this is removing the patient age field (which looks like it didn't work well with our synthetic data anyway). This is now only calculated in encounters relative to the age at the time of visit.

tests/test_data/core/core__encounter.txt
```diff
@@ -1,50 +1,50 @@
-({'id': None, 'code': 'AMB', 'display': None, 'system': 'http://terminology.hl7.org/CodeSystem/v3-ActCode', 'userSelected': None, 'version': None}, 'AMB', 'ambulatory', 'finished', None, None, None, None, None, None, None, None, 15, datetime.date(2018, 6, 3), datetime.date(2018, 6, 3), datetime.date(2018, 5, 28), datetime.date(2018, 6, 1), datetime.date(2018, 1, 1), 'Patient/e80dda2c-a260-dbf7-3167-bf0945f3a91d', 'Encounter/029d1814-d7bf-0624-524d-7ccda5f320f6', '029d1814-d7bf-0624-524d-7ccda5f320f6', 'female', 'black or african american', 'not hispanic or latino', '672')
```
dogversioning, line 1: This changes were we pull the end of an encounter from, which is null in our test data set. I'll verify this out of thread, but maybe we should coalesce to start date if end date is null?

tests/conftest.py
```diff
@@ -41,7 +41,9 @@ class ResourceTableIdPos(IntEnum):
     CONDITION = 7
     DOCUMENTREFERENCE = 11
     ENCOUNTER = 20
-    OBSERVATION = 14
+    MEDICATION = 1
+    MEDICATIONREQUEST = 1
+    OBSERVATION = 1
```
dogversioning, line 46: So, I knocked observation ID to the first col since I was already doing a bunch of changes to get everything unnested correctly. 

Either at the end of the PR review, or as a followon, I am going to move the primary ID of each core table to the first column, and then remove this check. I just didn't want to pollute the test data changes too much while there were other structural changes going on, just to make it more obvious what was actually changing as part of this.

cumulus_library/studies/core/builder_medication.py
```diff
@@ -106,6 +107,7 @@ def prepare_queries(self, cursor: object, schema: str, *args, **kwargs) -> dict:
             or medication_datasources["by_external_ref"]
         ):
             self.queries.append(
+                # core_templates.get_core_template()
                 get_core_medication_query(medication_datasources, has_userselected)
```
dogversioning, line 111: I need to remove this, and the entry from templates, and wire it up to the core specific template loader - i don't think this is blocking review, though

cumulus_library/studies/core/manifest.toml
```diff
@@ -4,22 +4,18 @@ study_prefix = "core"
...
+    "builder_patient.py",
+    "builder_encounter.py",
+    "builder_documentreference.py",
+    "builder_medication.py",
+    "builder_medicationrequest.py",
+    "builder_observation.py"
 ]
 
 [sql_config]
 file_names = [
```
dogversioning, line 16: I left three bare sql files in here since they didn't touch the FHIR tables at all, but I might? move them to be loaded from the various builders, the way the prereq builder works.

cumulus_library/studies/core/core_templates/core_utils.jinja
```diff
@@ -1,38 +1,138 @@
+{#- A collection of macros for constructing common safe column selects
```
mikix, line 1: This line feels like it relates to some of these util methods but not others - do you want a `schema.jinja` file? (I'm fine with them still living in `utils` but maybe this comment would more naturally live above the block with the safe-select macros)

cumulus_library/studies/core/core_templates/core_utils.jinja
```diff
@@ -1,38 +1,138 @@
...
 
--- gets a column, if present, from a basic SQL primitive col, or a null otherwise
-{% macro basic_nullable_cols(table, alias, targets, schema) %}
+-#}
+{% macro nullable_cols(table, alias, targets, schema) %}
         {%- for col in targets %}
-        {%- if schema[table][col] %}
+        {#- depth one nested column-#}
+        {%- if col is not string and col|length ==3%}
+        {%- if schema[table][col[0]][col[1]] %}
```
mikix, line 34: nit: would it look too weird to indent the logic tree statements, but not the output stuff? Like:
```
    {%- if col is not string and col|length ==3%}
        {%- if schema[table][col[0]][col[1]] %}
        {{ alias }}.{{ col[0].lower() }}.{{ col[1].lower() }} AS {{col[2]}}
        {%- endif -%}
    {%- endif -%}
```

or
```
        {%- if col is not string and col|length ==3%}
            {%- if schema[table][col[0]][col[1]] %}
        {{ alias }}.{{ col[0].lower() }}.{{ col[1].lower() }} AS {{col[2]}}
            {%- endif -%}
        {%- endif -%}
```

I dunno - I wish jinja code was more readable in general.

cumulus_library/studies/core/core_templates/core_utils.jinja
```diff
@@ -1,38 +1,138 @@
...
+        {{ alias }}.{{ col[0].lower() }}.{{ col[1].lower() }} AS {{col[2]}}
+        {%- else %}
+        NULL AS {{col[2]}}
+        {%- endif %}
+        {#- depth two nested column -#}
+        {%- elif col is not string and col|length ==4%}
+        {%- if schema[table][col[0]][col[2]] %}
+        {{ alias }}.{{ col[0].lower() }}.{{ col[1].lower() }}.{{ col[2].lower() }} AS {{col[3]}}
```
mikix, line 42: So I was inspired by this line to look up whether SQL does provide guarantees about case insensivity for column names. Looks like (as always with SQL) [everything depends](https://learnsql.com/blog/sql-case-sensitive/). Mostly things are case insensitive, but the default collation on some SQL servers will use the system default (so be case sensitive on Linux and insensitive on Windows - yikes). But collation can be changed.

Anyway, I'm skeptical on our need to cater to oddball databases that we don't even support right now, but I'm fine with it. Our two database targets both are insensitive for column names. It might be interesting to find out if Azure is (I would have to assume it is, but worth validating).

cumulus_library/studies/core/core_templates/core_utils.jinja
```diff
@@ -1,38 +1,138 @@
...
+        {%- endif %}        
+        {{- comma_delineate(loop) }}
+        {%- endfor %}
+{%- endmacro %}
+
+{#- creates a date from a from an ISO formatted date str 
+targets is an array expecting a data type of the following:
+    string: refers to a vanilla SQL primitive column
+    tuple(col, object, col_alias): refers to a one level nested SQL object
+    tuple(col, object, , child, col_alias): refers to a two level nested SQL obj
```
mikix, line 60: typo with extra comma?

cumulus_library/studies/core/core_templates/core_utils.jinja
```diff
@@ -1,38 +1,138 @@
...
+-#}
+{% macro date_cols_from_str(table, alias, targets, schema) %}
+        {%- for col in targets %}
+        {#- depth one nested column-#}
+        {%- if col is not string and col|length ==3%}
```
mikix, line 65: super nit: why do you never leave a space after the number on these?

cumulus_library/studies/core/core_templates/core_utils.jinja
```diff
@@ -1,38 +1,138 @@
...
+        cast(NULL AS date) AS {{ col.lower() }}
+        {%- endif %}    
         {{- comma_delineate(loop) }}
         {%- endfor %}
 {%- endmacro %}
 
--- gets a column, if present, from a value of an object in a col
-{% macro depth_one_nullable_cols(table, alias, targets, schema) %}
+{#- creates a date from a list of columns that may contain a partial date str -#}
+{% macro partial_date_cols(table, alias, targets, schema) %}
```
mikix, line 89: Why does this need to exist? (i.e. why not just call `date_cols_from_str`)  I believe this function's logic is already the behavior of `from_iso8601_timestamp` - and I coded this logic into the duckdb version of that function too.

cumulus_library/studies/core/core_templates/documentreference.sql.jinja
```diff
@@ -0,0 +1,64 @@
+{% import 'core_utils.jinja' as utils %}
+CREATE TABLE core__documentreference AS
+WITH temp_documentreference AS (
+    SELECT DISTINCT
+        {{- utils.basic_cols(
```
mikix, line 5: How useful is `basic_cols` vs just using `nullable_cols`? Early detection of oddities, I guess. But like, what if in the data there **is** a null `id` field (that would be messed up, but bear with me). How should our pipeline treat that? I believe ETL requires the `id` field to exist, but doesn't verify that the value itself isn't null. And then this would error out when they tried to build core?

Just curious if we couldn't just treat everything as nullable and simplify some templating. With our growing collection of data-sanity studies (`quality` and the other one you made I don't remember the name), I wonder if `core` is the right place to enforce basic stuff like this, rather than being graceful about data it receives, which the user probably couldn't easily fix anyway.

cumulus_library/studies/core/core_templates/documentreference.sql.jinja
```diff
@@ -0,0 +1,64 @@
...
+            ) 
+        }},
+        {{- utils.nullable_cols(
+                'documentreference',
+                'dr',
+                [
+                    'type',
+                    'status',
+                    'docstatus',
+                    'context',
```
mikix, line 20: super nit: trailing whitespace here that I happened to note because of a nervous highlighting tic :smile:

cumulus_library/studies/core/core_templates/documentreference.sql.jinja
```diff
@@ -0,0 +1,64 @@
...
+                ],
+                schema
+            ) 
+        }},
+    FROM documentreference AS dr
+)
+
+SELECT DISTINCT
+    type_coding.type_row AS doc_type,
+    coalesce(type_coding.type_row.code, 'None') AS doc_type_code,
```
mikix, line 44: Not `cumulus__missing-or-null`?

cumulus_library/studies/core/core_templates/documentreference.sql.jinja
```diff
@@ -0,0 +1,64 @@
...
+    CASE
+        WHEN
+            type_coding.type_row.display IS NOT NULL
+            THEN replace(type_coding.type_row.display, ',', '')
```
mikix, line 48: We can quote our CSV now, yeah? Is this still needed?

If it is still needed, I'd expect this to live in some `get_user_string` macro or something that would be called everywhere we use a display/text value from FHIR.

cumulus_library/studies/core/core_templates/encounter.sql.jinja
```diff
@@ -0,0 +1,117 @@
...
+        {{- utils.date_cols_from_str(
+                'encounter',
+                'e',
+                [
+                    ('period', 'start', 'period_start'),
+                ],
+                schema
+            ) 
+        }},
+        --{{ schema['encounter'] }} }}
```
mikix, line 40: :scissors:

cumulus_library/studies/core/core_templates/encounter.sql.jinja
```diff
@@ -0,0 +1,117 @@
...
+    e.status,
+    e.type_code,
+    e.type_code_system,
+    e.sevicetype_code,
+    e.sevicetype_code_system,
+    e.priority_code,
+    e.priority_code_system,
+    e.reasoncode_code,
+    e.reasoncode_code_system,
+    date_diff('year', date(p.birthdate), e.period_start_day) AS age_at_visit,
```
mikix, line 98: This is probs fine since the ETL always gives you year-only birthDates. But if we ever want to handle cases where the day might actually fall anywhere in the year, this is the logic I used in `quality` (works in both athena and duckdb, unlike `date_diff`):

```
                year(targetDateTime)
                - year(birthDate)
                - (
                    CASE WHEN
                        extract(doy FROM birthDate) > extract(doy FROM targetDateTime)
                    THEN 1
                    ELSE 0
                    END
                )
```

mikix (commented):
Not done yet, but I'm going to break for lunch and wanted to dump these on ya first  :smile:

cumulus_library/studies/core/builder_observation.py
```diff
@@ -0,0 +1,69 @@
+""" Module for extracting US core extensions from patient records"""
+from cumulus_library import base_table_builder
+from cumulus_library.template_sql import templates, utils
+from cumulus_library import databases
+from cumulus_library.studies.core.core_templates import core_templates
+
+CCC = utils.CodeableConceptConfig
```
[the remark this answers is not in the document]
dogversioning: there is a `ConceptMap` thing in FHIR - I dunno, normally I'm like 'be explicit in the class name' but this one just rubbed me the wrong way being used 5 times

cumulus_library/studies/core/builder_observation.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+# TODO: upgrade to 3.10+, use kw_only flag to subclass a dataclass for generating source/target
+code_sources = [
+    CCC(column_name="category", is_array=True, filter_priority=False),
+    CCC(column_name="code", is_array=False, filter_priority=False),
+    CCC(column_name="interpretation", is_array=True, filter_priority=False),
+    CCC(column_name="valuecodeableconcept", is_array=False, filter_priority=False),
+]
+for source in code_sources:
```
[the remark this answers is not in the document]
dogversioning: this will go away on the version bump, so i'm inclined to not worry about it -  #167 for breadcrumbs

cumulus_library/studies/core/core_templates/core_utils.jinja
```diff
@@ -1,38 +1,138 @@
+{#- A collection of macros for constructing common safe column selects
+
+    For layout, is is generally assumed that these are being invoked from within
+    a WITH TABLE AS(...) statement, so they will always assume two levels of
+    indentation.
```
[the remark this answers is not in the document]
dogversioning: i will die on this hill

cumulus_library/studies/core/core_templates/core_utils.jinja
dogversioning: can you be more explicit about this? i don't get the vibe.

cumulus_library/studies/core/core_templates/core_utils.jinja
dogversioning: so - this is configurable as a jinja config param. i was explicitly asked to minimize the amount of indentation, but i agree that i would prefer to have if indentation. 

if we do, i want to do it as a seperate PR. this is already too big for my poor brain to reason over.

cumulus_library/studies/core/core_templates/core_utils.jinja
dogversioning: so - this is actually not about catering to a specific DB. the intent here is to make information_schema.cols-style calls easier for downstream users, since we hide it behind some templates they can elect to use. the goal is to be able to use the results of that table as a means to look into other sources without having to do casts. 

if that's sensible, i can document the reason for this rats nest for future readers? i agree its ugly as hell, but i like the end product.

cumulus_library/studies/core/core_templates/core_utils.jinja
dogversioning: because sqlfluff will complain and force me to fix it/fix it for me. 

i'm glad we don't do pair programming often, because the amount of stuff like this i do without thinking about it, throwing it off to a linter to handle later, would make you %100 insane. like mixing single quotes and double quotes in the same line! i'm a monster, but at least i know enough to set traps for myself.

cumulus_library/studies/core/core_templates/core_utils.jinja
dogversioning: mostly for the schema validation step - but the case block could be replaced with `date_cols_from string`.

cumulus_library/studies/core/core_templates/documentreference.sql.jinja
dogversioning: This is a field that could reasonably be expected to be null, so we explictly coalesce it to avoid the error state reporting.

cumulus_library/studies/core/core_templates/documentreference.sql.jinja
dogversioning: i think the only advantage is query speed, and i don't know how big that is.

cumulus_library/studies/core/core_templates/documentreference.sql.jinja
dogversioning: part of the sqlfluff pass - though i should check my local settings, i would have expected this to be caught pre-commit.

cumulus_library/studies/core/core_templates/documentreference.sql.jinja
dogversioning: it isn't - and i also realized that this probably shouldn't be here at all, this coding should be in a denormalized table, so some of this may just go away.

cumulus_library/studies/core/core_templates/encounter.sql.jinja
dogversioning: frankly i just copied this over verbatim from the raw sql without thinking about it.

but - can we talk about this? i assume there's a reason, which is why i didn't touch this, but that feels like an odd behavior.

cumulus_library/studies/core/core_templates/core_utils.jinja
mikix: Oh just that `collection of macros for constructing common safe column` looks to be a file-level description but this file has other general utils stuff like loop comma handling, etc. I was just suggesting moving this comment above the schema-related methods.

cumulus_library/studies/core/core_templates/core_utils.jinja
mikix: I didn't follow your intent comment - maybe let's chat on slack or something for higher bandwidth clarity

cumulus_library/studies/core/core_templates/core_utils.jinja
mikix: :smile:

Are you saying that sqlfluff will fix this for you so you don't worry about it (in which case, this is going to be fixed later as you work on linting errors?), or are you saying that sqlfluff *wants* there to be no space?

cumulus_library/studies/core/core_templates/encounter.sql.jinja
mikix: `date_diff` works like you want I believe, in Athena. (i.e. it takes day of year into account, is my memory)

But not duckdb - in duckdb, it is naive and just does year-minus-year.

So I used the logic above to get both to work the same. I'm guessing the author of that sql just didn't have duckdb in their brains because why would they. But you might hid oddities during unit testing if you ever feed this a non `-01-01` date.

cumulus_library/studies/core/core_templates/medicationrequest.sql.jinja
```diff
@@ -0,0 +1,59 @@
...
+CREATE TABLE core__medicationrequest AS
+WITH temp_mr AS (
+    SELECT 
+        {{- utils.basic_cols(
+                'medicationrequest',
+                'mr',
+                ['id', 'status', 'intent']
+            ) 
+        }},
+        {{- utils.date_cols_from_str(
```
mikix, line 11: tiny nit: this method and `truncate_date_cols` is starting to feel redundant - is there an opportunity to merge these two into the same - like allow a non-period-tuple argument that just uses the raw date? But absolutely not necessary - I'm just trying to make your `utils.jinja` file even longer :smile:

cumulus_library/studies/core/core_templates/medicationrequest.sql.jinja
```diff
@@ -0,0 +1,59 @@
...
+            ) 
+        }},
+        {{- utils.nullable_cols(
+                'medicationrequest',
+                'mr',
+                [
+                    'display',
+                    ('subject', 'reference', 'subject_ref')
+                ],
+                schema
```
mikix, line 34: nit; do you need to pass this down? Can the method just grab it from the global variables itself? Don't know if your brain would like or hate that, but might save some lines of code.

cumulus_library/studies/core/core_templates/observation.sql.jinja
```diff
@@ -0,0 +1,72 @@
+{% import 'core_utils.jinja' as utils %}
+
+CREATE TABLE core__observation AS
+WITH temp_observation AS (
+    SELECT
+        {{- utils.basic_cols(
+                'observation',
+                'o',
+                ['id','status', 'effectivedatetime']
```
mikix, line 9: `effectiveDateTime` is an optional field - it should be in a nullable cols call, right?

cumulus_library/studies/core/core_templates/observation.sql.jinja
```diff
@@ -0,0 +1,72 @@
...
+            ) 
+        }},
+        --{{ schema }}
```
mikix, line 12: :scissors:

cumulus_library/studies/core/core_templates/patient.sql.jinja
```diff
@@ -1,17 +1,23 @@
...
+{% import 'core_utils.jinja' as utils %}
 CREATE TABLE core__patient AS
 WITH temp_patient AS (
     SELECT DISTINCT
-        p.gender,
+        {{- utils.nullable_cols(
+            'patient',
+            'p',
+            [
+                'id', 'gender', 'address'
```
mikix, line 9: nit: you've been putting `id` in a basic_cols call

cumulus_library/studies/core/core_templates/core_utils.jinja
mikix: `date_cols_from_string` also checks the schema, yeah?

cumulus_library/studies/core/core_templates/encounter.sql.jinja
dogversioning: so the bit I actually wanted to talk about was truncating the birthdates at the ETL - just wanted to understand more of why that's happening.

cumulus_library/template_sql/templates.py
```diff
@@ -144,7 +80,9 @@ def get_column_datatype_query(schema_name: str, table_name: str, column_names: L
 def get_core_medication_query(
     medication_datasources: dict, has_userselected: Optional[bool] = False
 ):
-    with open(f"{PATH}/core_medication.sql.jinja") as core_medication:
+    with open(
+        f"/Users/mgarber/code/cumulus-library/cumulus_library/studies/core/core_templates/medication.sql.jinja"
```
mikix: ^

mikix (commented):
Dropping another group of comments, before I dig into the tests

cumulus_library/studies/core/core_templates/core_utils.jinja
dogversioning: the former.

cumulus_library/studies/core/core_templates/encounter.sql.jinja
mikix: Oh that was because of reducing PHI vibes - we keep most datetimes in place, for clinical reasons. But I guess they (Andy & Jamie) felt like birthdates were more sensitive and we should keep truncating them.

For reference, that happens at the MS-deid-tool layer and we _could_ change the config to not do that.

cumulus_library/studies/core/core_templates/core_utils.jinja
mikix: From that discussion: Matt was talking about new (alias) column names, I meant to be talking about the column name _lookups_.

We both agreed a single standard made sense for aliases.

cumulus_library/studies/core/study_period.sql
```diff
@@ -17,11 +17,11 @@ WITH documented_encounter AS (
...
         ce.status,
         cd.doc_ref,
-        date_diff('day', ce.start_date, cd.author_date) AS diff_enc_note_days,
+        date_diff('day', ce.start_date, date(cd.author_date)) AS diff_enc_note_days,
         coalesce(ce.enc_class_code, 'None') AS enc_class_code,
         coalesce(ce.enc_class_display, 'None') AS enc_class_display,
-        coalesce(cd.doc_type.code, 'None') AS doc_type_code,
-        coalesce(cd.doc_type.display, cd.doc_type.code) AS doc_type_display
+        coalesce(cd.doc_type_code, 'None') AS doc_type_code,
+        coalesce(cd.doc_type_display, 'None') AS doc_type_display
```
mikix, line 24: Ah a repeat of my missing-or-null comments.

mikix (approved):
OK finished looking at this :face_exhaling: and looks good! Thanks for all this refactoring.

Comments can be resolved as you see fit, but nothing deal breaking.

cumulus_library/studies/core/core_templates/core_utils.jinja
dogversioning: Did the spirit rather than the letter - i marked the loop comma thing as needing to move to a tbd global macro location.

cumulus_library/studies/core/core_templates/core_utils.jinja
dogversioning: yep - removed

cumulus_library/studies/core/core_templates/medicationrequest.sql.jinja
mikix: I was testing a similar thing in the `quality` study and I think they cannot grab it from the global variables after all.  :-/

cumulus_library/studies/core/core_templates/encounter.sql.jinja
dogversioning: OK - I created #169 - we're using that now in a way that's slightly inaccurate and i want to, at some point, just discuss the implications once more.

cumulus_library/studies/core/core_templates/medicationrequest.sql.jinja
dogversioning: I'm going to pass on this for now - these are enough different in context to me that i want them to be explicit.

PR #170, opened 2024-01-30
CLI sql generation

dogversioning:
- Added a cli param, `generate-sql`, which will parse a study directory and dump the output of builders into a `/reference_sql` subdir
  - Included output for core
- Light refactors while I was in here:
  - Renamed 'StudyBuilder' to 'StudyRunner', since building happens elsewhere and it's a better namespace seperation
  - cleaned up class imports in cli.py to be module level imports.

### Checklist


cumulus_library/base_table_builder.py
```diff
@@ -102,19 +103,29 @@ def post_execution(
...
         self.queries = commented_queries
 
-    def write_queries(self, filename: str = "output.sql"):
+    def write_queries(
+        self, filename: str = "output.sql", dir_path: pathlib.Path = None
+    ):
         """writes all queries constructed by prepare_queries to disk"""
+        if dir_path:
+            dir_path.mkdir(exist_ok=True)
+            filename = dir_path / filename
```
mikix, line 128: nit: out of curiousity, does `path /= filename` work? Feels like I wouldn't want that to work. Just wondering if Python lets you handle the operator `/` but disallow `/=`.

cumulus_library/cli.py
```diff
@@ -280,18 +297,19 @@ def get_studies_by_manifest_path(path: PosixPath) -> dict:
 
 def run_cli(args: Dict):
     """Controls which library tasks are run based on CLI arguments"""
+    print(args)
```
mikix, line 300: :scissors:

cumulus_library/cli.py
```diff
@@ -280,18 +297,19 @@ def get_studies_by_manifest_path(path: PosixPath) -> dict:
...
 
     # all other actions require connecting to the database
     else:
-        db_backend = create_db_backend(args)
+        db_backend = databases.create_db_backend(args)
         try:
-            builder = StudyBuilder(db_backend, data_path=args.get("data_path"))
-            if args["verbose"]:
+            builder = StudyRunner(db_backend, data_path=args.get("data_path"))
+            if "verbose" in args and args["verbose"]:
```
mikix, line 312: nit: `if args.get("verbose")` is maybe easier

cumulus_library/cli.py
```diff
@@ -336,14 +354,22 @@ def run_cli(args: Dict):
...
                     for target in args["target"]:
                         builder.export_study(study_dict[target], args["data_path"])
+
+            elif args["action"] == "generate-sql":
+                if "all" in args["target"]:
+                    for target in study_dict.keys():
+                        builder.generate_all_sql(study_dict[target])
+                else:
+                    for target in args["target"]:
+                        builder.generate_study_sql(study_dict[target])
```
mikix, line 364: nit: this is a noiser change, but `builder` as a variable name might be confusing with the class rename.

cumulus_library/study_parser.py
```diff
@@ -369,9 +371,17 @@ def _load_and_execute_builder(
...
         table_builder_class = table_builder_subclasses[0]
         table_builder = table_builder_class()
-        table_builder.execute_queries(
-            cursor, schema, verbose=verbose, drop_table=drop_table, parser=parser
-        )
+        if generate_sql:
+            table_builder.prepare_queries(cursor, schema, parser=parser)
+            table_builder.comment_queries(doc_str=doc_str)
+            table_builder.write_queries(
+                filename=f"{filename.split('.')[0]}.sql",
```
mikix, line 378: nit: is there a more robust / pathlib-y way to get the basename? Like `basename` (didn't check if that exists) or something? Just thinking about filenames with periods in them like `hello.kitty.txt`

cumulus_library/study_parser.py
```diff
@@ -503,8 +515,37 @@ def run_single_table_builder(
...
+        )
+
+    def run_generate_sql(
+        self,
+        cursor: databases.DatabaseCursor,
+        schema: str,
+        parser: databases.DatabaseParser = None,
+        **kwargs,
+    ) -> None:
+        """Loads modules from a manifest and executes code via BaseTableBuilder
```
mikix, line 528: Is that accurate? Feels like it converts jinja templates to sql more than run code.

cumulus_library/study_parser.py
```diff
@@ -503,8 +515,37 @@ def run_single_table_builder(
...
+
+        :param cursor: A DatabaseCursor object
+        :param schema: The name of the schema to write tables to
+        :param verbose: toggle from progress bar to query output
+        """
+        all_generators = (
+            self.get_table_builder_file_list()
+            + self.get_counts_builder_file_list()
+            + self.get_statistics_file_list()
         )
```
mikix, line 538: This feels like something that could get stale. Should there be a utility method that grabs all these and used in multiple places? Or does that just not make sense for this setup.

cumulus_library/study_parser.py
```diff
@@ -319,9 +319,11 @@ def _load_and_execute_builder(
         filename: str,
         cursor: databases.DatabaseCursor,
         schema: str,
-        verbose: bool,
+        verbose: bool = False,
         drop_table: bool = False,
         parser: databases.DatabaseParser = None,
+        generate_sql: bool = False,
```
mikix, line 325: nit: is there a sensible way to reword this variable for clarity? Like even when this is False, SQL is still generated, right (I think)? It's actually whether that SQL is executed that is different between the two flows.

mikix (approved):
I did not examine the generated reference sql at all - I trust it looks like expected

cumulus_library/base_table_builder.py
dogversioning: Looks like it doesn't.

cumulus_library/study_parser.py
dogversioning: Oh there is -  `Path.stem`, which, i get the name, but odd choice.

I could? rewrite the write_queries method to natively take a Path as a single arg and do that coercion here.

cumulus_library/study_parser.py
dogversioning: Reworded, lemme know what you think.

cumulus_library/study_parser.py
dogversioning: This is currently the only place that needs to know about all the table builder types in one go. If there was another place, I'd agree. I'm inclined to wait until that happens?

cumulus_library/study_parser.py
dogversioning: `write_reference_sql` is more verbose but it's explicit.

PR #171, opened 2024-02-01
Counts generation peformance improvements

dogversioning:
- Primarily, it restructures how count queries work
  - Frustratingly, it is faster to run multiple queries with a single DISTINCT statement, cube them, and then join them together afterwards, rather than doing all the DISTINCT statements in one category
- Tech debt:
  - Split out CLI verbose output from progress bar rendering, since the former should happen before a query is run, and the latter should happen after.
  - In files where this required adjusting imports, if they were not already doing module level versus class level imports, converted the file to the former.
  - Improved sorting for core testing data export in the hope that re-running it will be less noisy in git history.
  - cutover to `cumulus__none` as the blessed representation of null data from the source system.

### Checklist


cumulus_library/template_sql/statistics/count.sql.jinja
```diff
@@ -5,23 +5,45 @@
...
 {%- endif -%}
 {%- endmacro -%}
 
+{#- Moved extant queries to be secondary resource only, but leaving tertiary table
+infrastructure for future/moving resource specific logic to template generator -#}
+{%- macro tertiary_resource(fhir_resource) -%}
+{#- implicit null -#}
+{%- endmacro -%}
+
+{%- macro cols_delenated_list(table_cols, fhir_resource)-%}
```
mikix, line 25: typo: delineated

cumulus_library/template_sql/statistics/count.sql.jinja
```diff
@@ -5,23 +5,45 @@
...
+            {% for col in table_cols %}
+            "{{ column_or_alias(col) }}"
+            {{- comma_delineate(loop) }}
+            {%- endfor %}
+            {%- if fhir_resource in ('documentreference', 'observation' ) -%},
+            enc_class_display
+            {% endif %}
```
mikix, line 32: This felt like an odd addition to this otherwise business-logic-neutral util method. As someone coming fresh to this code, this would surprise me. (here and in `coalesced_list`)

There are a couple places in this file where you iterate `table_cols` but don't include `enc_class_display` -- what is the different business-logic context there? Whatever it is, could you put that in the method name here, to clarify that this method also includes contextual logic stuff? (like `cols_for_cubing_as_a_delineated_list` or something)

Or is there a way to use `enc_class_display` in those other places, and then you could just have it be in the `table_cols` list and this method looks nice and lean?

(Also, I see in the final select, you don't use this method, because you need to prefix with `p.` -- maybe an optional `prefix` arg to the macro could help there.

And in `null_replacement` you don't use the coalesced version of this macro, because you are changing aliases at same time... hmm maybe that's harder to fix, unless you allow a `call` version of the macro)

dogversioning: so i have this on my list of followon tasks - similar to how i moved `cond_display_code` out, i agree this would be good to move to the functions upstream that call the template generator method.

however - that's a larger structural change, because i then need to create a way to define a table join upstream (condition didn't have this problem since the table refactor included the foreign columns being referenced). I'm on the fence of how to do this atm (like, do i store some artifact that says which tables are joinable, and then leverage that, enforcing a two table only join? or do i provide a syntax like we do in the psm config, where you can specify arbitrary joins)?

Anyway, in the interest of not letting the perfect getting in the way of the good, I'm electing to defer this, since it doesn't structurally impact the final table output, just the way that you get there, but I'll handle this at some point soon (as referenced in the eaten comment).

cumulus_library/databases.py
```diff
@@ -216,6 +216,21 @@ def __init__(self, db_file: str):
             None,
             duckdb.typing.TIMESTAMP,
         )
+        self.connection.create_function(
+            # This is in support of the md5 function - duckdb's implementation of
+            # md5 just takes a string (very reasonable!), but trino's implementation
+            # takes a varbinary type (which duckdb does not support), so we have to
+            # run a string through a UTF8 conversion in production environments.
```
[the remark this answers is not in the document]
mikix: :+1:

cumulus_library/template_sql/statistics/count.sql.jinja
mikix: That's fine as a medium term goal. I guess my other request was "in the absence of refactoring the code, is there a name that makes more sense for this util method" -- something to indicate to a newbie to this code why this method includes a custom resource/column check.

Like, this is named like a generic utility method that will spit out column names. But you have lots of places where you don't use it because this method has additional business logic. So... some intern coming to this is going to be like "it's a bug that we don't use this utility method in these other looping cases" or something. I dunno what a good name is because it's not 100% to me what the semantic meaning of this business logic is compared to all the places you loop over `table_cols` _without_ including this logic. Which is part of my problem - the code flow is confusing and some comments or more descriptive names would help (absent any refactoring)

PR #172, opened 2024-02-02
Cleanup on primary key position

dogversioning:
- moves primary key to first position in tables, renames it so that it is always `id`
- removes test infrastructure to work around random primary key placement
- adds a few missing derived tables to core unit testing.

### Checklist


cumulus_library/template_sql/statistics/count.sql.jinja
```diff
@@ -4,6 +4,8 @@
 {%- endif -%}
 {%- endmacro -%}
 
+{#- TODO: move secondary/tertiary resource declaration upstream to
```
dogversioning, line 7: these are my attempt to address the dangling comment from the prior PR i pulled the trigger on early

cumulus_library/study_parser.py
```diff
@@ -311,9 +311,7 @@ def _execute_drop_queries(
             drop_view_table = templates.get_drop_view_table(
                 name=view_table[0], view_or_table=view_table[1]
             )
-            with helper.query_console_output(
-                verbose, drop_view_table, progress, task
-            ) as v:
+            with helper.query_console_output(verbose, drop_view_table, progress, task):
```
dogversioning, line 314: dangling debug from the last PR

cumulus_library/template_sql/statistics/count.sql.jinja
mikix: You didn’t pull early really, I was just being verbose in the comments 😄

tests/conftest.py
```diff
@@ -35,18 +35,6 @@
 }
 
 
-class ResourceTableIdPos(IntEnum):
```
mikix, line 38: Nice

mikix (approved):

Issue #167, opened 2024-01-24
Tech debt: upgrade min python version


dogversioning:
There's some dataclass subclassing wonkiness that is addressed in 3.10 that would be nice to have. We should go to 3.10 (or maybe leapfrog to 3.11).

PR #173, opened 2024-02-05
Uprev to 3.10, namespace clarification

dogversioning:
- Min python is now 3.10 #167 
  - Dataclass inheritence cleanup of some code in observation 
- Made the namespace of the base utils/template wrapper and utils a little clearer

### Checklist


.github/workflows/ci.yaml
```diff
@@ -52,19 +56,23 @@ jobs:
...
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
           pip install .
-      # See https://github.com/aws-actions/configure-aws-credentials for configuring
+      # See https://github.com/Moulick/configure-multiple-aws-roles for configuring
       # the aws side of this - the below action is just a light wrapper
       - name: Configure AWS Credentials
-        uses: mcblair/configure-aws-profile-action@v0.1.1
+        uses: Moulick/configure-multiple-aws-roles@v4
```
dogversioning, line 71: So, dusting this off - the original action is using an AWS API which is about to be deprecated. I've been watching [this PR](https://github.com/mcblair/configure-aws-profile-action/pull/1#event-11695267238) waiting for a fix to go in. This ended up being resolved via a fork.

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "cumulus-library"
-requires-python = ">= 3.9"
+requires-python = ">= 3.11"
```
mikix, line 3: Let me start with this change and we can discuss rest after.

Things this will make more difficult:
- Running in BCH's E2 environment. 3.9 is the only Python3 available via biogrids in that environment. There may be an alternate way to get a different python3 but I don't know how friendly that is. It would need research. So this change would make cumulus-library un-runnable in E2. -- That is probably OK for now? I don't have an immediate need for that, but it's something to note.
- Running on Mike's laptop. I'm still using the LTS release of 22.04 (until this April when 24.04 comes out and I'll probably upgrade). Python 3.10 is the default python here. I can install python 3.11 and with some work, can probably figure out how to use it for cumulus-library, but that's investigation I can do. Not a blocker per se, but possibly annoying. If you don't really need 3.11 features, is it alright to drop this down to 3.10 until 24.04 comes out and I upgrade?

dogversioning: Per an offline discussion, we're going to hold to 3.10 until at least the next Ubuntu release as a baseline for 'what's reasonable to ask of others' guideline.

mikix: Thank you :pray: for letting me be lazy for a little longer

.github/workflows/ci.yaml
```diff
@@ -18,12 +18,12 @@ jobs:
...
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
 
       - name: Set up Python 
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v5
         with:
-          python-version: 3.9
+          python-version: '3.10'
```
mikix, line 26: small point here: we're only testing on 3.10, but as a command line CLI that doesn't ship with its own python in a docker image (i.e. unlike the ETL) - we should probably test on all supported pythons

Doesn't need to happen on this PR - I filed #174 for it.

mikix (approved):

PR #175, opened 2024-02-06
Ruff

dogversioning:
- Swapped out black & manual linting for ruff
- Global linting/formatting pass
  - About %50 of this is just backing out the pre-3.10 Typing module

### Checklist


cumulus_library/statistics/psm.py
dogversioning, line 1: All the noqa statements in this file are just 'PsmPy is doing something under the hood I don't want to debug unless I can open a PR against it to fix it'. Still waiting on that.

.github/workflows/ci.yaml
```diff
@@ -47,10 +47,10 @@ jobs:
       - name: Run sqlfluff on jinja templates
         run: |
           sqlfluff lint
-      - name: Run black
+      - name: Run ruff
         if: success() || failure() # still run black if above checks fails
         run: |
-          black --check --verbose .
+          ruff
```
mikix: Ah... sensible defaults.

I don't think this will verify that the code is formatted as you expect, yeah? (i.e. that `ruff format` was run already -- which it might not be if the dev didn't install commit hooks, which drive-by contributors probably will forget to do). So it doesn't really replace the previous `black` command.

You might want to also add `ruff format --check` here

.pre-commit-config.yaml
```diff
@@ -1,15 +1,10 @@
...
     hooks:
-      - id: black
-        entry: bash -c 'black "$@"; git add -u' --
-        # It is recommended to specify the latest version of Python
-        # supported by your project here, or alternatively use
-        # pre-commit's default_language_version, see
-        # https://pre-commit.com/#top_level-default_language_version
-        language_version: python3.9
+      - id: ruff
+        args: [--fix]
```
mikix, line 6: Ah here's the philosophical rub. We've talked about how appropriate it is to automatically fix bad code for the developer in a way that they aren't brought into the loop so they can learn to be better.

This does bring them into the loop a little bit, by not auto-committing the results at least. So I'd like to keep it that way, if/when we make the auto-formatting auto-committed.

I'd personally prefer to keep such checks out of the commit hooks altogether because of my WIP-commit-heavy workflow. I assume your workflow is different.

Anyway, we don't need to agree on this kind of thing. Just raising the point again.

.pre-commit-config.yaml
```diff
@@ -1,15 +1,10 @@
...
+      - id: ruff-format
```
mikix, line 7: On the other hand, formatting I find it awkward to push in front of the dev - they didn't do anything wrong really. So this I'd like to see return to auto-commit status at some point, once we trust it (which off-PR we talked about as the reason for this change).

cumulus_library/base_table_builder.py
```diff
@@ -122,7 +121,9 @@ def comment_queries(self, doc_str=None):
         commented_queries.pop()
         self.queries = commented_queries
 
-    def write_queries(self, path: pathlib.Path = pathlib.Path.cwd() / "output.sql"):
+    def write_queries(self, path: pathlib.Path | None = None):
+        if path is None:
+            path = pathlib.Path.cwd() / "output.sql"
         """writes all queries constructed by prepare_queries to disk"""
```
mikix, line 127: This docstring got downgraded to just a loose string.

cumulus_library/cli.py
```diff
@@ -264,11 +261,12 @@ def get_study_dict(alt_dir_paths: List) -> Optional[Dict[str, pathlib.Path]]:
...
     # first, we'll get any installed public studies
     with open(
-        pathlib.Path(cli_path, "./module_allowlist.json"), "r", encoding="utf-8"
+        pathlib.Path(cli_path, "./module_allowlist.json"), encoding="utf-8"
     ) as study_allowlist_json:
         study_allowlist = json.load(study_allowlist_json)["allowlist"]
-    site_packages_dir = sysconfig.get_path("purelib")
+    site_packages_dir = "".join(sysconfig.get_path("purelib"))
     for study, subdir in study_allowlist.items():
+        print(site_packages_dir)
```
mikix, line 269: :scissors:

cumulus_library/cli.py
```diff
@@ -264,11 +261,12 @@ def get_study_dict(alt_dir_paths: List) -> Optional[Dict[str, pathlib.Path]]:
 
     # first, we'll get any installed public studies
     with open(
-        pathlib.Path(cli_path, "./module_allowlist.json"), "r", encoding="utf-8"
+        pathlib.Path(cli_path, "./module_allowlist.json"), encoding="utf-8"
     ) as study_allowlist_json:
         study_allowlist = json.load(study_allowlist_json)["allowlist"]
-    site_packages_dir = sysconfig.get_path("purelib")
+    site_packages_dir = "".join(sysconfig.get_path("purelib"))
```
mikix, line 267: This is odd... that method returns a string as far as I can tell. And this join is pointless then. Does this method return an array sometimes?

cumulus_library/databases.py
```diff
@@ -238,22 +242,24 @@ def __init__(self, db_file: str):
...
+        This is often the output folder of Cumulus ETL"""
         for name, table in tables.items():
             self.connection.register(name, table)
 
     @staticmethod
     def _compat_array_join(
-        value: Optional[list[Optional[str]]], delimiter: str
-    ) -> Optional[str]:
+        value: list[str | None] | None, delimiter: str
+    ) -> str | None:
```
mikix, line 254: It can be `None` alright?! Geeze...

cumulus_library/studies/discovery/code_detection.py
```diff
@@ -1,39 +1,40 @@
...
             task = progress.add_task(
                 "Discovering available coding systems...",
                 total=len(code_sources),
             )
             for code_source in code_sources:
                 if code_source["is_array"]:
-                    code_source["has_data"] = utils.is_codeable_concept_array_populated(
+                    code_source[
+                        "has_data"
+                    ] = sql_utils.is_codeable_concept_array_populated(
```
mikix, line 23: This construct doesn't feel very readable to me.

Maybe a temporary variable would be better? Like `has_data = sql_utils...(` and then `code_source["has_data"] = has_data`

But small nit - it's fine as is.

cumulus_library/study_parser.py
```diff
@@ -336,7 +332,7 @@ def _load_and_execute_builder(
 
         As with eating an ortolan, you may wish to cover your head with a cloth.
         Per an article on the subject: "Tradition dictates that this is to shield
-        – from God’s eyes – the shame of such a decadent and disgraceful act."
+        - from God's eyes - the shame of such a decadent and disgraceful act."
```
mikix, line 335: **wat.** If `ruff` made you take out the PROPER unicode apostrophe and replace it with this tragically vertical approximation of an apostrophe, it's not fit for purpose!

\<small\>it's fine...\</small\>

cumulus_library/template_sql/base_templates.py
```diff
@@ -212,6 +211,8 @@ def get_insert_into_query(
     :param table_cols: Comma deleniated column names, i.e. ['first','second']
     :param dataset: Array of data arrays to insert, i.e. [['1','3'],['2','4']]
     """
+    if type_casts is None:
+        type_casts = {}
```
mikix, line 215: nit: might be a matter of taste, but I think `type_casts = type_casts or {}` is more readable / common. But this works too. (here and elsewhere)

pyproject.toml
```diff
@@ -55,3 +54,24 @@ build-backend = "flit_core.buildapi"
...
+    "E",  # pycodestyle
+    "F",  # pyflakes
+    "I",  # isort
+    "ISC",  # implicit string concatenation
+    "PLE",  # pylint errors
+    "RUF",  # the ruff developer's own rules
+    "UP",  # alert you when better syntax is available in your python version
+]
+lint.ignore = [
+    "ISC001"  # Recommended ingore from `ruff format`
```
mikix, line 73: Let's keep the gore out plz

But also - why is it recommended to ignore this? Inline concatenation is pretty rare / odd to see. I'm not fighting for this rule, just curious.

pyproject.toml
```diff
@@ -55,3 +54,24 @@ build-backend = "flit_core.buildapi"
...
+]
+
+[tool.ruff.lint.per-file-ignores]
+"cumulus_library/schema/valueset.py" = ["E501"]
```
mikix, line 77: Why is github highlighting this stuff red?

(Also, maybe a comment about why this ignore / when if ever it could be removed)

tests/conftest.py
```diff
@@ -121,7 +119,7 @@ def update_nested_obj(id_path, obj, i):
                             # panda's deep copy is not recursive, so we have to do it
                             # again for nested objects
                             df[id_path[0]] = df[id_path[0]].map(
-                                lambda x: update_nested_obj(
+                                lambda x, i=i, id_path=id_path: update_nested_obj(
```
mikix, line 122: :throws hands up: fine

tests/regression/run_regression.py
```diff
@@ -32,8 +31,8 @@
     export_missing = exports - references
     sys.exit(
         "❌ Found differences in files present: ❌\n"
-        f"Files present in reference not in export: {str(ref_missing)}\n"
-        f"Files present in export not in reference: {str(export_missing)}"
+        f"Files present in reference not in export: {ref_missing!s}\n"
+        f"Files present in export not in reference: {export_missing!s}"
```
mikix, line 35: Is `str()` not the default here?

tests/regression/run_regression.py
```diff
@@ -59,7 +58,7 @@
             diffs.append(
                 [
                     filename,
-                    ("Rows differ between reference and export:\n" f"{compared}"),
+                    ("Rows differ between reference and export:\n", f"{compared}"),
```
mikix, line 61: This change doesn't seem right - I think the original intention of this parens and code was to have two lines of string concatenation, like:

```
                    (
                        "Rows differ between reference and export:\n"
                        f"{compared}")
                    )
```

Like the block above has. But at some point it must have got moved to the same line.

With the change in this PR, you'd be printing a tuple instead of a string.

tests/test_cli.py
```diff
@@ -29,7 +27,8 @@ def open(self, *args, **kwargs):
         if str(args[0]).endswith(".bsv"):
             args = (
                 Path(
-                    f"./tests/test_data/mock_bsvs/{str(args[0]).rsplit('/', maxsplit=1)[-1]}"
+                    "./tests/test_data/mock_bsvs/",
+                    f"{str(args[0]).rsplit('/', maxsplit=1)[-1]}",
```
mikix, line 31: nit: I think you can drop the surrounding f-string if you're breaking it off to its own Path() arg

tests/test_cli.py
```diff
@@ -101,11 +100,12 @@ def test_cli_early_exit(args):
         ),
     ],
 )
-def test_cli_path_mapping(
-    mock_load_json, mock_path, tmp_path, args, raises, expected
-):  # pylint: disable=unused-argument
+def test_cli_path_mapping(mock_load_json, mock_path, tmp_path, args, raises, expected):  # pylint: disable=unused-argument
```
mikix, line 103: This feels like a long line for `ruff` - curious why it allows it here.

tests/test_cli.py
```diff
@@ -101,11 +100,12 @@ def test_cli_early_exit(args):
...
     with raises:
-        mock_path.return_value = f"{Path(__file__).resolve().parents[0]}" "/test_data/"
+        mock_path.return_value = (
+            f"{Path(__file__).resolve().parents[0]}",
+            "/test_data/",
+        )
```
mikix, line 108: This is changing the return type to a tuple instead of a string. Here and below.

tests/test_conftest.py
```diff
@@ -15,17 +14,17 @@ def test_ndjson_data_generator(tmp_path):
         for filepath in [f for f in Path(target / key).iterdir()]:
             with open(filepath) as f:
                 first_new = json.loads(next(f))
-                for line in f:
-                    pass
-                last_new = json.loads(line)
+                *_, last_new = f
+                last_new = json.loads(last_new)
```
mikix, line 18: :thinking: OK...

tests/test_conftest.py
```diff
@@ -15,17 +14,17 @@ def test_ndjson_data_generator(tmp_path):
...
                 # handling patient file of length 1:
-                line = first_line
-                for line in f:
-                    pass
-                last_ref = json.loads(line)
+                try:
+                    *_, last_ref = f
+                    last_ref = json.loads(last_ref)
+                except Exception:
+                    last_ref = first_ref
```
mikix, line 27: What's with this exception here? That's new behavior yeah? Is there a more specific exception class to reference?

.github/workflows/ci.yaml
mikix: Ah, this is obsolete by the time I submitted. Marking resolved.

tests/test_cli.py
mikix: Is it interesting / expected that this change didn't break the test?

tests/conftest.py
dogversioning: [another candidate](https://docs.astral.sh/ruff/rules/function-uses-loop-variable/#flake8-bugbear-b) for exclusion, but this is bugbear so maybe not:

tests/regression/run_regression.py
dogversioning: str() inside fstring calls is a thing that was a constant source of complaints, i'll get the raw code back

tests/regression/run_regression.py
dogversioning: moving to a single string

this happened a lot - this is the thing that has me nervous atm. I don't have a test to test if the regression test runner is working correctly, which is why i didn't catch this one. feels a little like turtles all the way down, but can add a tech debt ticket for it.

tests/test_cli.py
dogversioning: #noqa/pylint codes seem to be allowed to break line length restricitions. i dont hate it.

tests/test_cli.py
dogversioning: OH that's the source of the `site_packages_dir = "".join(sysconfig.get_path("purelib"))` issue, i'll fix that here and update there.

tests/test_conftest.py
dogversioning: if you try to unpack an iterator and its empty, you get an error - the docs didn't specify which. I can do some research.

pyproject.toml
dogversioning: will add, lemme check the formatting, might need to move the x.y into [tool.ruff.x] y or something

cumulus_library/databases.py
mikix: That wouldn't help here exactly, because these were explicit Optionals not implicit.

But I can _see_ the logic for disallowing implicit optional - like if change the default value to not be `None` anymore, it's a little surprising for that to also change the type of the arg itself. But that's the kind of argument for a library with a strict API and not just devs making code.

I'm fine either way. Disabling individual rules is a little annoying, config wise, but maybe it's a win overall.

pyproject.toml
mikix: Ah so it's about conflicts between the formatter and the linter. Sure - maybe expand the comment to link that ticket or say something along those lines. Because this would hopefully be a temporary ignore, until `ruff` fixes itself.

tests/regression/run_regression.py
mikix: Yeah I'm just wondering why python even has `!s`... like without it, is there any behavior change on some edge cases or something? I'd assumed that f-string formatting would apply str() on non-strings for us, but maybe not.

tests/test_conftest.py
mikix: Speaking of, it's interesting to me that `ruff` doesn't complain about a loose `Exception` catch. Which... I don't mind. There are plenty of valid reasons for it.

Can you check if it _would_ complain about a bare `except`? Like if you did `except:` just alone -- I would want it to treat that as a flaggable error (and probably not an auto-fix one).

tests/regression/run_regression.py
dogversioning: ah, ok here we go: [RUF010](https://docs.astral.sh/ruff/rules/explicit-f-string-type-conversion/)

tests/test_conftest.py
dogversioning: it did, that's how we ended up here in the first place.

mikix: There wasn't a bare-except here before. But maybe during PR work you had one?

.pre-commit-config.yaml
dogversioning: ok, per this and our chat off PR about this last night, I did some digging and found that pre-commit supports hooks at basically every lifecycle stage, so i've moved this check to pre-push. It didn't look like it is installed normally, so you have to specify the stage with `pre-commit install --hook-type pre-push`, but it then does the linting only on push to remote.

cumulus_library/studies/discovery/code_detection.py
dogversioning: related to my 'mccabe complexity is a headache' comment, i just went ahead and moved this logic to a function so the line length wasn't an issue.

.pre-commit-config.yaml
mikix: Interesting... OK - we have the `pre-commit` instructions in our CONTRIBUTING.md docs usually (I think that's what we named them). - maybe those need updating.

It's annoying that pre-commit isn't more turnkey, but fine :smile:

pyproject.toml
dogversioning: Expanded this comment.

pyproject.toml
dogversioning: fixed formatting, added comment.

tests/test_conftest.py
dogversioning: yeah, it happened midstream

.pre-commit-config.yaml
dogversioning: ah yep, meant to do that and spaced out - bumped the docs.

mikix (approved):

PR #176, opened 2024-02-07
WIP: Be looser about case when checking schemas

mikix:
This commit allows the following:
- Case-insensitive schema lookups (i.e. table name and column name arguments to get_column_datatype_query() can be any case)
- Case-insensitive schema validation (i.e. the expected column names passed to validate_table_schema() can be any case)

### Checklist
- [ ] Consider if documentation (like in `docs/`) needs to be updated
- [ ] Update template repo if there are changes to study configuration


tests/test_duckdb.py
```diff
@@ -71,4 +73,54 @@ def test_duckdb_from_iso8601_timestamp(timestamp, expected):
         .execute(f"select from_iso8601_timestamp('{timestamp}')")
         .fetchone()[0]
     )
-    assert parsed, expected
+    assert parsed == expected
```
mikix, line 76: AHEM.  :sob: 

Fixing this bad assert (point in favor of assertEqual I suppose...) revealed that duckdb's version that I wrote is returning naive timestamps - which I think I actually did manually change from aware timestamps at one point... but Athena looks like it uses aware ones too....  Needs more investigation. I just fixed the tests to work here, but didn't change any business logic for now. 

Added a TODO in the code though.

tests/test_data/core/core__encounter.txt
```diff
@@ -21,7 +21,7 @@
...
-('83d0d564-3bbf-48eb-7445-bd2b81130671', {'id': None, 'code': 'EMER', 'display': None, 'system': 'http://terminology.hl7.org/CodeSystem/v3-ActCode', 'userSelected': None, 'version': None}, 'EMER', 'emergency', 'finished', None, None, None, None, None, None, '55680006', 'http://snomed.info/sct', 69, datetime.date(2018, 6, 16), datetime.date(2018, 6, 16), datetime.date(2018, 6, 11), datetime.date(2018, 6, 1), datetime.date(2018, 1, 1), 'Patient/16be855b-ece2-8b96-1ef9-a4d93adf3289', 'Encounter/83d0d564-3bbf-48eb-7445-bd2b81130671', 'male', 'white', 'not hispanic or latino', '660')
+('83d0d564-3bbf-48eb-7445-bd2b81130671', {'id': None, 'code': 'EMER', 'display': None, 'system': 'http://terminology.hl7.org/CodeSystem/v3-ActCode', 'userSelected': None, 'version': None}, 'EMER', 'emergency', 'finished', None, None, None, None, None, None, '55680006', 'http://snomed.info/sct', 69, datetime.date(2018, 6, 16), datetime.date(2018, 6, 17), datetime.date(2018, 6, 11), datetime.date(2018, 6, 1), datetime.date(2018, 1, 1), 'Patient/16be855b-ece2-8b96-1ef9-a4d93adf3289', 'Encounter/83d0d564-3bbf-48eb-7445-bd2b81130671', 'male', 'white', 'not hispanic or latino', '660')
```
mikix, line 24: These changes in the encounter tables were because of the bug where after the old code `split` the schema string, future searches were only in the second half of the split.

If you gave an expected field list with a different order than DuckDB gives it back (in this case, the encounter builder was feeding (start, end) but the schema was giving back (end, start)), then the `end` field would never be detected as existing in the schema. So for DuckDB *and* Athena, the `end` field would always fall back to the `start` value.

cumulus_library/databases.py
```diff
@@ -64,36 +66,33 @@ def _parse_found_schema(self, expected: dict[dict[list]], schema: list[list]):
...
-                # is this an object column?
-                if len(expected[column]) > 0:
-                    for field in expected[column]:
-                        col_schema = col_schema.split(field, 1)
-                        if len(col_schema) != 2:
-                            output[column][field] = False
-                            col_schema = col_schema[0]
-                        else:
-                            output[column][field] = True
-                            col_schema = col_schema[1]
```
mikix, line 84: This here was the source of one of the bugs - after splitting on each field, future scans would only look at the part *after* the split. So order of fields matters in a way that is hard to predict.

I've switched from splitting entirely and just search in the entire column schema for each field.

cumulus_library/template_sql/column_datatype.sql.jinja
```diff
@@ -4,5 +4,5 @@ SELECT
 FROM information_schema.columns
 WHERE
     table_schema = '{{ schema_name }}'
-    AND table_name = '{{ table_name }}'
-    AND LOWER(column_name) IN ('{{ column_names|join("', '") }}') --noqa: LT05
+    AND table_name = '{{ table_name|lower }}'
```
mikix, line 7: I tried this same line with `LOWER(table_name)` and Athena took a minute and then errored out about `DELTA_LAKE_INVALID_SCHEMA` and `Metadata not found in transaction log...` etc.  :shrug: - removing the lower fixes that oddity and doesn't super matter - we don't make mixed case tables today.

cumulus_library/studies/core/core_templates/core_utils.jinja
```diff
@@ -90,22 +90,22 @@ targets is assumed to be a list of tuples of one of the following format:
         {%- for col in targets %}
         {%- if col is not string and col|length ==4%}
         {%- if schema[table][col[0]][col[1]] %}
-        date_trunc('{{ col[3] }}', date(from_iso8601_timestamp({{ alias }}.{{ col[0] }}.{{ col[1] }}))) 
+        date_trunc('{{ col[3] }}', date(from_iso8601_timestamp({{ alias }}."{{ col[0] }}"."{{ col[1] }}")))
```
mikix, line 93: This came up because now that the regression test actually sees the `period.end` field, it tries to access it. But `end` is a keyword that can't just be injected like that.

So I took the safest but ugliest approach - quoting each arbitrary field access.
- A less safe but prettier option is just quoting the last field access. That's where trickier words like `end` and `date` are likely to show up.
- The least safe but prettiest option is to have the caller of this quote it. Like `('period', '"end"', 'period_end_day', 'day'),` when calling `truncate_date_cols`

cumulus_library/databases.py
dogversioning: So, this does create a different kind of bug, though maybe note as critical as the found one: for columns where a object parameter occurs more than once, this may over-identify data as being present. specifically, id shows up in a lot of places. 

if this :does: hit, it will prevent a table from being run, which is the spirit of the previous version - take a few false negatives for more true positives.

If this is blocking you now, that's fine, but we might need to more robustly solve this in the near future.

dogversioning (approved):

cumulus_library/databases.py
mikix: We chatted about this in person - I think we agree it is maybe not a new bug but just another instance of the existing flakiness of "searching in a big string for a field". We need to move to per-db-backend schema parsing to fix this and to allow deeper-depth schema searching.

PR #177, opened 2024-02-07
Centralized jinja template entrypoint

dogversioning:
- template_sql/base_templates now has a single entrypoint for rendering templates
  - This also supports loading templates from designated folders
  - As a result, everything going through this now gets access to macros
- reorged all template renderers to go through this entrypoint
- moved statistics templates to be a subfolder of statistics for pattern standardization

### Checklist
- [ ] Consider if documentation (like in `docs/`) needs to be updated - They do, I'm going to do that in one go.


dogversioning:
I decided to not go crazy on moving additional macro snippets into shared locations, but it should be simple to do that going forward now.

mikix (approved):

PR #178, opened 2024-02-08
Expanded code system preference lists

dogversioning:
This PR makes the following change:
- Adds code systems IDed by discovery runs to places where we walk down system lists

### Checklist


cumulus_library/studies/discovery/code_detection.py
```diff
@@ -8,7 +8,7 @@
 class CodeDetectionBuilder(base_table_builder.BaseTableBuilder):
     display_text = "Selecting unique code systems..."
 
-    def _check_coding_against_db(code_source, schema, cursor):
+    def _check_coding_against_db(self, code_source, schema, cursor):
```
dogversioning, line 11: Need to have tests for this study - #179

cumulus_library/studies/core/builder_condition.py
```diff
@@ -54,6 +54,9 @@ def denormalize_codes(self):
                 "http://snomed.info/sct",
                 "http://hl7.org/fhir/sid/icd-10-cm",
                 "http://hl7.org/fhir/sid/icd-9-cm",
+                "urn:oid:1.2.840.114350.1.13.71.2.7.2",
+                "urn:oid:1.2.840.114350.1.13.71.2.7.4",
```
mikix, line 58: All these private/arbitrary systems need comments.

cumulus_library/template_sql/codeable_concept_denormalize.sql.jinja
```diff
@@ -21,7 +21,7 @@ CREATE TABLE {{ target_table }} AS (
         {%- endif %}
         {%- if filter_priority %}
         WHERE
-            u.codeable_concept.system = '{{ system }}'
+            u.codeable_concept.system LIKE '{{ system }}%'
```
mikix, line 24: Oh thats.. interesting. As a developer passing a system around, I would be very surprised to see it used as a prefix rather than a black-box string. `http://example.org/animals/` being different from `http://example.org/animals/lions` for example. What's the use case for this - private urn namespaces? Is there maybe a way to opt-into the wildcard nature? But even then, I'm leery that we get a promise about all possible meanings / systems in the namespace?

dogversioning: there's a lot of noise in the numeric ones specifically, and without the context, I couldn't determine what they corresponded to and how comprehensive they were, so rather than putting in 15 codes i tried to cheat it - i can flip this back and be more explicit.

cumulus_library/studies/core/builder_condition.py
```diff
@@ -54,6 +54,12 @@ def denormalize_codes(self):
                 "http://snomed.info/sct",
                 "http://hl7.org/fhir/sid/icd-10-cm",
                 "http://hl7.org/fhir/sid/icd-9-cm",
+                "http://hl7.org/fhir/sid/icd-9-cm/diagnosis",
+                # Cerner specific systems
```
mikix, line 58: That doesn't seem entirely true - it's a spec standard way of indicating why data is gone. Epic doesn't use it though, it sounds like.

cumulus_library/studies/core/builder_condition.py
```diff
@@ -54,6 +54,12 @@ def denormalize_codes(self):
...
+                "http://terminology.hl7.org/CodeSystem/data-absent-reason",
+                # EPIC specific systems
```
mikix, line 60: I'm guessing we can't say anything else about them? But it'd be nice if we could distinguish what these two mean even a little bit.

cumulus_library/studies/core/builder_encounter.py
```diff
@@ -71,6 +86,12 @@ def denormalize_codes(self, schema, cursor):
                 "code_systems": [
                     "http://terminology.hl7.org/CodeSystem/v3-ActPriority",
                     "http://snomed.info/sct",
+                    "http://hl7.org/fhir/sid/icd-10-cm",
+                    "http://hl7.org/fhir/sid/icd-9-cm",
+                    # Cerner specific systems
+                    "https://fhir.cerner.com/%/nomenclature",
```
mikix, line 92: :thinking:

mikix (approved):

cumulus_library/template_sql/codeable_concept_denormalize.sql.jinja
dogversioning: ok, per our conversation offline, i've updated this to the explicit IDs, and removed the installation-specific guids from the cerner IDs.

I created #180 to track the need to think about some optimizations as these lists get larger.

cumulus_library/studies/core/builder_condition.py
dogversioning: yeah, hard agree!

cumulus_library/studies/core/builder_condition.py
dogversioning: oh good catch

Issue #182, opened 2024-02-09
tech debt: reference sql validation


dogversioning:
CI should do the following so that our reference SQL is always up to date as a documentation reference:
- run `generate-sql` and ensure there are no changes
- execute sql against a database and ensure it's valid

Issue #122, opened 2023-09-12
Tech Debt: Study creation cleanup


dogversioning:
Currently, there are two ways a user can create a study:

- By cloning a template from github
- Via the CLI's create command

It may be better to support just one way of doing this - internally we're currently only using the first method, but the second is more in line with the spirit of having all the commands available in one place.

We should consider doing one of the following:
- Remove the CLI create method
- Deprecate the github template repo, and have the CLI be a bit more guided about study creation and create git infrastructure/manifest formatting dynamically (see AWS SAM/vite for examples of good guided CLI flows
- If keeping both, consider adding a task in the PR template to remind to cross-update on changes.

dogversioning:
So - this is now covered by the [Cumulus docsite](https://docs.smarthealthit.org/cumulus/library/creating-studies.html) - we may elect to make this the 'official' way to make a study, in which case we'll deprecate the other creation mechanisms.

PR #183, opened 2024-02-09
Added archival mode

dogversioning:
- Adds an `--archive` flag to the export mode, with a warning about how this is potentially sensitive:
image
- Removes the create argument per the docs being a 'good enough' source #122
- Removes a nested field from encounter we only caught because dictionaries are complained about when sorting dataframes
- Regenned reference SQL after the yesterday's commits
  - #182 to remember to automate checks for this

### Checklist
- [ ] Consider if documentation (like in `docs/`) needs to be updated - They do, will do in one go


cumulus_library/cli.py
```diff
@@ -344,11 +347,28 @@ def run_cli(args: dict):
                             )
 
             elif args["action"] == "export":
+                if args["archive"]:
```
mikix, line 350: I initially had a thought that maybe this should be a toplevel action instead of a flag on `export` but after thinking about it more, decided you had the right of it. Just talking out loud - in case you had thoughts on that too. But I think this makes sense.

cumulus_library/studies/core/reference_sql/builder_condition.sql
```diff
@@ -34,7 +34,7 @@ CREATE TABLE core__condition_codable_concepts_display AS (
             condition AS s,
             UNNEST(s.code.coding) AS u (codeable_concept)
         WHERE
-            u.codeable_concept.system = 'http://hl7.org/fhir/sid/icd-10-cm'
+            u.codeable_concept.system LIKE 'http://hl7.org/fhir/sid/icd-10-cm'
```
mikix, line 37: I'm lightly curious about the performance cost of this. Correctness is more important obvi, but if this is noticeably slower, I imagine the jinja could switch to like if it detects a wildcard.

cumulus_library/study_parser.py
```diff
@@ -665,29 +668,62 @@ def _execute_build_queries(
...
+        table_list = []
+        path = pathlib.Path(f"{data_path}/{self.get_study_prefix()}/")
         self.reset_counts_exports()
+        if archive:
+            table_query = base_templates.get_show_tables(
+                schema_name, f"{self.get_study_prefix()}__"
+            )
+            result = db.cursor().execute(table_query).fetchall()
+            for row in result:
+                table_list.append(row[0])
```
mikix, line 693: nit: would feel more Pythonic to me to do `table_list = [row[0] for row in result]` and drop the `table_list = []` bit at the top. But I have a known bias for comprehensions :smile:

cumulus_library/study_parser.py
```diff
@@ -665,29 +668,62 @@ def _execute_build_queries(
...
             dataframe = dataframe.sort_values(
                 by=list(dataframe.columns), ascending=False, na_position="first"
             )
             dataframe.to_csv(
                 f"{path}/{table}.csv", index=False, quoting=csv.QUOTE_MINIMAL
             )
-            dataframe.to_parquet(f"{path}/{table}.parquet", index=False)
-            queries.append(query)
+            if not archive:
+                dataframe.to_parquet(f"{path}/{table}.parquet", index=False)
```
mikix, line 711: Why not parquet too in this case? Just space reasons?

This method is starting to feel a little awkward. You start with forked logic (which tables) and end with forked logic (how to handle the results). Only the middle is shared. Should the middle be moved to a helper method and you have two different methods here, like:
```
if archive:
    do_archive()
else:
    do_export()

def do_archive():
   get tables
   do_inner_export()
   zip results

def do_export():
    get tables
    do_inner_export
    parquet results
```

Or maybe instead move some of the "gross" specialized code (like "getting all tables" or "zipping up all tables") into helpers, so that this method can focus just on the if/else-ing.

I dunno. No action needed per se, just started feeling like a lot of very different code based on some `if` conditions.

mikix (approved):

cumulus_library/studies/core/reference_sql/builder_condition.sql
dogversioning: there is a half baked idea in my head of moving the system extraction logic into here in some way, or in some other way doing db introspection to help build these queries out in a more nuanced fashion.

but - only doing LIKE when required seems like low hanging fruit in the interim.

cumulus_library/study_parser.py
dogversioning: something something jack of all trades, master of none/javascript poisoning something something - agree, will change

cumulus_library/study_parser.py
dogversioning: re: parquet, my original thought was 'this is for parking data ahead of a paper', since that where this request originated, and since that's all csvs in our current use case, i deferred to that. but there's no reason :not: to parquet, and i guess that enables you to upload an older set of data to the aggregator, so maybe it's worth doing - and that would decrease some of the branching logic. so lemme do that and then i'll see how that feels w.r.t. breaking things out.

I have a nagging doubt about some of the  infrastructure in this file -  specifically, since its starting to get large, so I think that biases me towards trying to get everything in one place rather than making more functions. 

This might be a bad idea. I feel like at :some: point this might need to get broken out into a set of functions per arg inheriting from a base, but i'm reluctant to go that far at the moment, which might be driving some other less sustainable architecture choices.

dogversioning: moved zipping out to utils - that and the parquet cleanup help.

PR #191, opened 2024-03-06
US core required/must support review

dogversioning:
- Adds remaining fields in US core suppported profile to core tables
- Moves encounter builder CodeableConcepts to dataclass based instead of dict based
- `--builder` cli arg now looks for a substring
- added `--builder` to sql generator subparser

### Checklist


.pre-commit-config.yaml
```diff
@@ -12,4 +12,6 @@ repos:
   - repo: https://github.com/sqlfluff/sqlfluff
     rev: 2.3.4
     hooks:
-      - id: sqlfluff-lint
\ No newline at end of file
+      - id: sqlfluff-lint
+        types: []
+        types_or: [sql,jinja]
```
dogversioning, line 17: this is very silly - please reinstall the pre-commit hook after this goes in

cumulus_library/studies/core/builder_documentreference.py
```diff
@@ -35,7 +36,31 @@ def prepare_queries(
...
                     is_array=False,
                     target_table="core__documentreference_dn_type",
-                )
+                ),
+                # TODO: The spec allows any value for category, but it's unclear what
+                # the value for the systems would be. For the time being, we select the
+                # spec valid code (which is a set of one value).
+                # It may be worth confirming the values in this field with either the
+                # quality or discovery studies on an ongoing basis to find other valid
+                # values
```
mikix, line 45: I might reword this slight for clarity:
- instead of spec, maybe specifically mention the US Core v4 profile (vs the base FHIR spec, which doesn't care at all about the category system)
- maybe instead of "allows any" mention "is extensible" (or if that's too nerdy, something about "allows any, but prefers this one")
- for "spec valid code" maybe say something like "the recommended code system"
- when I hear "is a set of one value", I just wanna know the one value :smile: - maybe mention `one value: "clinical-note"`

cumulus_library/studies/core/builder_documentreference.py
```diff
@@ -35,7 +36,31 @@ def prepare_queries(
...
+                    target_table="core__documentreference_dn_category",
+                    code_systems=[
+                        "http://hl7.org/fhir/us/core/ValueSet/us-core-documentreference-category"
+                    ],
+                ),
+                sql_utils.CodeableConceptConfig(
+                    source_table="documentreference",
+                    source_id="id",
+                    column_name="content.format",
+                    is_array=False,
```
mikix, line 61: `content` is an array though - does the generator automatically assume parents are arrays?

cumulus_library/studies/core/core_templates/condition.sql.jinja
```diff
@@ -74,9 +74,14 @@ SELECT
     tc.recordeddate,
     tc.recordeddate_week AS recorded_week,
     tc.recordeddate_month AS recorded_month,
-    tc.recordeddate_year AS recorded_year
+    tc.recordeddate_year AS recorded_year,
+    t_clinicalstatus_coding.clinicalstatus_row.code AS clinicalstatus_code,
```
mikix, line 78: I hate that allergies and conditions do a complicated codeable concept status field. Just use a static code field like all the other resources. You think you're too good for a static code, Condition? You're not.

cumulus_library/study_parser.py
```diff
@@ -142,6 +142,14 @@ def get_export_table_list(self) -> list[str] | None:
                 )
         return export_table_list
 
+    def get_all_generators(self) -> list:
```
mikix, line 145: nit: "list[xxx]" would be helpful

cumulus_library/study_parser.py
```diff
@@ -505,46 +513,49 @@ def run_statistics_builders(
...
-        if not name.endswith(".py"):
-            name = f"{name}.py"
-        self._load_and_execute_builder(
-            name, cursor, schema, verbose=verbose, drop_table=True, parser=parser
-        )
+        """targets all table builders matching a target string for running"""
+        all_generators = self.get_all_generators()
+        for file in all_generators:
+            if builder and file.find(builder) == -1:
+                continue
```
mikix, line 528: nit: both places you ask for all generators, you immediately filter by `builder` - should that maybe instead be an argument to the get-generators method?

docs/core-study-details.md
```diff
@@ -232,32 +256,54 @@ for more information.
...
-|gender                |varchar    |           |
-|race_display          |varchar    |           |
-|ethnicity_display     |varchar    |           |
-|postalcode3           |varchar    |           |
+|             Column             |   Type    |Description|
+|--------------------------------|-----------|-----------|
+|id                              |varchar    |           |
+|status                          |varchar    |           |
+|enc_class_code                  |varchar(6) |           |
+|enc_class_display               |varchar(21)|           |
```
mikix, line 264: What causes these columns to be size-limited? We probably don't want them to be, yeah?

mikix (approved):

cumulus_library/studies/core/builder_documentreference.py
dogversioning: yep, all great notes.

cumulus_library/studies/core/builder_documentreference.py
dogversioning: good catch - will update.

dogversioning: for posterity - I think this actually requires a larger chunk of effort, since this is a coding rather than a codeable concept and I think they's kind of smashed together at the moment in a way that suggests reusability but doesn't actually provide it.

Since this PR is already huge, I want to take this as is, where it will always just generate a null table, and work on the other thing seperately. #192 to track.

PR #193, opened 2024-03-14
Medicationrequest cleanup

dogversioning:
This PR makes the following updates:
- Adds missing R4 US Core fields to medication reference
- Prioritizes the medicationcodeableconcept in place of cross joining from medicationreference
- renames rx_* to medication_* to more closely align with the FHIR spec
- Upgrades sqlfluff to version 3, released this week
- reference sql regen

### Checklist


cumulus_library/.sqlfluff
```diff
@@ -49,7 +49,7 @@ prefix = Test
...
...ss': {'code': True, 'system': True, 'display': False, 'userSelected': True, 'version': True}, 'subject': {'reference': True, 'display': False, 'type': True}, 'id': True}, 'medicationrequest': {'id': True, 'status': True, 'intent': True, 'authoredon': True, 'category': {'code': True, 'system': True, 'display': False}, 'subject': {'reference': True}, 'encounter': {'reference': True}}, 'observation': {'id': True, 'category': {'coding': True, 'code': True, 'display': True, 'system': True, 'text': True}, 'status': True, 'code': {'coding': True, 'code': True, 'display': True, 'system': True, 'text': True}, 'interpretation': {'coding': True, 'code': True, 'display': True, 'system': True, 'text': True}, 'referencerange': {'low': False, 'high': False, 'normalvalue': False, 'type': False, 'appliesto': False, 'age': False, 'text': True}, 'effectivedatetime': True, 'valuequantity': {'value': True, 'comparator': False, 'unit': False, 'system': False, 'code': False}, 'valuecodeableconcept': {'coding': True, 'code': True, 'display': True, 'system': True}, 'subject': {'reference': True}, 'encounter': {'reference': True}}, 'patient': {'id': True, 'gender': True, 'address': True, 'birthdate': True}}
```
mikix, line 52: This is... a bad file. Not saying it needs changing. Just whinging.

mikix (approved):

cumulus_library/.sqlfluff
dogversioning: ok, as mentioned briefly offline, i went down the rabbit hole on this - sqlfluff uses the core configparser lib, which has some allowance for [multiline configs](https://docs.python.org/3/library/configparser.html#supported-ini-file-structure), which is not quite the way i want it to behave but is good enough to make this more readable than it has been previously.

PR #200, opened 2024-03-27
A couple miscellaneous fixes

mikix:
I found these while working on tests for a different upcoming PR.

### Checklist

### DIFF
--- a/cumulus_library/studies/core/core_templates/documentreference.sql.jinja
+++ b/cumulus_library/studies/core/core_templates/documentreference.sql.jinja
@@ -22,7 +22,6 @@ WITH temp_documentreference AS (
                     'docstatus',
                     'context',                
                     ('subject', 'reference', 'subject_ref'),
-                    ('context', 'encounter', 'encounter_ref'),
                     ('context', 'period', 'start', 'author_date')
                 ], 
                 schema
@@ -49,7 +48,23 @@ WITH temp_documentreference AS (
     LEFT JOIN core__documentreference_dn_type AS cdrt ON dr.id = cdrt.id
     LEFT JOIN core__documentreference_dn_category AS cdrc ON dr.id = cdrc.id
     LEFT JOIN core__documentreference_dn_format AS cdrf ON dr.id = cdrf.id
+),
 
+temp_encounters AS (
+    SELECT
+        tdr.id,
+
+{% if schema["documentreference"]["context"]["encounter"] %}
+        context_encounter.encounter.reference AS encounter_ref
+    FROM temp_documentreference AS tdr,
+         unnest(context.encounter) AS context_encounter (encounter) --noqa
+
+{% else %}
+        '' AS encounter_ref
+    FROM temp_documentreference AS tdr
+    WHERE 1=0 -- forces an empty table
+
+{% endif %}
 )
 
 SELECT DISTINCT
@@ -67,8 +82,8 @@ SELECT DISTINCT
     tdr.author_year,
     tdr.doc_format_code,
     tdr.subject_ref,
-    coalesce(tdr.encounter_ref, context_encounter.encounter.reference) as encounter_ref,
+    te.encounter_ref,
     concat('DocumentReference/', tdr.id) AS doc_ref
-FROM temp_documentreference AS tdr,
-    unnest(context.encounter) AS context_encounter (encounter) --noqa
+FROM temp_documentreference AS tdr
+LEFT JOIN temp_encounters AS te ON tdr.id = te.id
 WHERE date(tdr.author_day) BETWEEN date('2016-06-01') AND current_date;
--- a/cumulus_library/studies/core/core_templates/patient.sql.jinja
+++ b/cumulus_library/studies/core/core_templates/patient.sql.jinja
@@ -34,24 +34,33 @@ WITH temp_patient AS (
         patient AS p
     LEFT JOIN core__patient_ext_race AS er ON p.id = er.id
     LEFT JOIN core__patient_ext_ethnicity AS ee ON p.id = ee.id
+),
+
+-- Grab a single address to use for this patient, if present
+temp_address AS (
+    SELECT
+        tp.id,
+        {# TODO: consider 'use' field to prefer a home address #}
+        {# complications: no postalcode on home address or only other 'use' values present #}
+        {# for now, just grab one arbitrarily (but reproducibly) #}
+        substr(max(t_address.addr_row.postalcode), 1, 3) AS postalcode3
+    FROM
+        temp_patient AS tp,
+        unnest(tp.address) AS t_address (addr_row)
+    GROUP BY tp.id
 )
 
 SELECT DISTINCT
     tp.id,
     tp.gender,
     tp.birthdate,
-    CASE
-        WHEN
-            t_address.addr_row.postalcode IS NOT NULL
-            THEN substr(t_address.addr_row.postalcode, 1, 3)
-        ELSE 'None'
-    END AS postalcode3,
+    coalesce(ta.postalcode3, 'cumulus__none') AS postalcode3,
     concat('Patient/', tp.id) AS subject_ref,
     coalesce(tp.race_display, 'unknown') AS race_display,
     coalesce(tp.ethnicity_display, 'unknown') AS ethnicity_display
 FROM
-    temp_patient AS tp,
-    unnest(tp.address) AS t_address (addr_row)
+    temp_patient AS tp
+LEFT JOIN temp_address AS ta ON tp.id = ta.id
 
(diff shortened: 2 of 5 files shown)


cumulus_library/studies/core/core_templates/patient.sql.jinja
```diff
@@ -34,24 +34,32 @@ WITH temp_patient AS (
...
-    END AS postalcode3,
+    coalesce(ta.postalcode3, 'None') AS postalcode3,
     concat('Patient/', tp.id) AS subject_ref,
     coalesce(tp.race_display, 'unknown') AS race_display,
     coalesce(tp.ethnicity_display, 'unknown') AS ethnicity_display
 FROM
-    temp_patient AS tp,
-    unnest(tp.address) AS t_address (addr_row)
+    temp_patient AS tp
+LEFT JOIN temp_address AS ta ON tp.id = ta.id
```
mikix, line 62: Left joining rather than unnesting means that we don't drop patients without addresses on the floor.

cumulus_library/studies/core/core_templates/patient.sql.jinja
```diff
@@ -34,24 +34,32 @@ WITH temp_patient AS (
...
     LEFT JOIN core__patient_ext_ethnicity AS ee ON p.id = ee.id
+),
+
+-- Grab a single address to use for this patient, if present
+temp_address AS (
+    SELECT
+        tp.id,
+        {# FIXME: use 'rank' to prefer an address #}
+        {# complications: no postalcode on best one or no ranks specified #}
+        substr(arbitrary(t_address.addr_row.postalcode), 1, 3) AS postalcode3
```
mikix, line 45: That felt more complicated than I wanted to chew off right now. But this change at least gets us out of duplicate patient rows. Which I assume is worse than an arbitrary address being used? But open to push back there, that this should be fixed differently.

cumulus_library/studies/core/core_templates/documentreference.sql.jinja
```diff
@@ -22,7 +22,6 @@ WITH temp_documentreference AS (
                     'docstatus',
                     'context',                
                     ('subject', 'reference', 'subject_ref'),
-                    ('context', 'encounter', 'encounter_ref'),
```
mikix, line 25: Because context.encounter was not a requested field from the schema, this would always result in NULL. And it's an array anyway, so I don't think this is the best access approach. I switched to an unnest below.

cumulus_library/studies/core/core_templates/documentreference.sql.jinja
```diff
@@ -67,8 +66,14 @@ SELECT DISTINCT
     tdr.author_year,
     tdr.doc_format_code,
     tdr.subject_ref,
-    coalesce(tdr.encounter_ref, context_encounter.encounter.reference) as encounter_ref,
+    {% if schema["documentreference"]["context"]["encounter"] %}
+    context_encounter.encounter.reference AS encounter_ref,
+    {% else %}
+    CAST(NULL AS VARCHAR) AS encounter_ref,
+    {% endif %}
```
mikix, line 73: This approach has the benefit of not blowing up for a database without context.encounter (thus, is helpful for light-weight unit tests).

But... if the field exists in the schema, the unnest means that we don't include any DocRefs without linked encounters at all (which might be a reasonable position). But if it's not in the schema, you _do_ get NULL encounter_refs.

So this fix is a bit inconsistent. And we should either unnest as a temp table then `LEFT JOIN` in order to allow loose DocRefs or change the no-schema approach to drop unlinked rows instead of using NULL.

cumulus_library/studies/core/core_templates/documentreference.sql.jinja
```diff
@@ -67,8 +66,14 @@ SELECT DISTINCT
...
     concat('DocumentReference/', tdr.id) AS doc_ref
-FROM temp_documentreference AS tdr,
-    unnest(context.encounter) AS context_encounter (encounter) --noqa
+FROM temp_documentreference AS tdr
+{% if schema["documentreference"]["context"]["encounter"] %}
+    , unnest(context.encounter) AS context_encounter (encounter) --noqa
+{% endif %}
```
mikix, line 78: Do we want an unnest here? This will create duplicate DocRef rows if there are multiple encounters in that array, yeah?

cumulus_library/studies/core/core_templates/patient.sql.jinja
dogversioning: in the cases where you were seeing this, is `use` present and populated? we could default to `home`? and then walk down?

regardless, unless someone is commuting by plane, on average i would think that these would be pretty close (especially on the reduction to 3 digits, unless someone is commuting from new hampshire to boston, so this is certainly better than duplicates.

cumulus_library/studies/core/core_templates/patient.sql.jinja
dogversioning: good catch, and one that i'm now very worried about.

cumulus_library/studies/core/core_templates/documentreference.sql.jinja
dogversioning: i think this has to be ok so you can traverse backwards, right? generally we're doing distinct ID checks on the primary resource when we're doing thinks like counts

this happens a lot with the core tables, for things like coding systems, for example - you've got to handle this in some way in your study queries.

cumulus_library/studies/core/core_templates/documentreference.sql.jinja
dogversioning: I think i'm in favor of the first one.

cumulus_library/studies/core/core_templates/patient.sql.jinja
mikix: Hmm, and even if we do keep `arbitrary()`, we should probably sort first.

dogversioning (approved):
Can we modify/add some data to the duckdb json that will exercise these changes?

cumulus_library/studies/core/core_templates/patient.sql.jinja
mikix: Oh yeah, all my talk of `rank` was me looking at the wrong struct 🤦 (ContactPoint) - yeah, Address looks like `use` would be what you'd want to cue off, if available. But yeah, I think this is a good first step, and we can optimize in future.

mikix:
> Can we modify/add some data to the duckdb json that will exercise these changes?

Yes, let me add my dumb little test that triggered these. Though it is a bit of a testing departure from the rest of the code - but I think it will be useful. Let me push up and you can comment.

cumulus_library/studies/core/core_templates/documentreference.sql.jinja
dogversioning: i've got to touch the core tables docs as part of my PR, i'll add a note at least there. I also have a WIP 'notes for study authors' doc this could get into.

mikix:
OK updated to resolve your comments, and added a little test helper class I was working on, for more dynamic ndjson rather than ndjson that lives on disk in a folder. Felt easier to create little edge cases for. But it's a bit of a departure from the norm in this repo - I can adjust it to your taste.

cumulus_library/studies/core/core_templates/patient.sql.jinja
```diff
@@ -34,24 +34,32 @@ WITH temp_patient AS (
...
     tp.id,
     tp.gender,
     tp.birthdate,
-    CASE
-        WHEN
-            t_address.addr_row.postalcode IS NOT NULL
-            THEN substr(t_address.addr_row.postalcode, 1, 3)
-        ELSE 'None'
-    END AS postalcode3,
+    coalesce(ta.postalcode3, 'None') AS postalcode3,
```
mikix, line 56: Should this be `cumulus__none`?

dogversioning (commented):
I do like that approach for edge cases, it feels less risky than mucking with the source data in terms of spiraling dependencies.

tests/testbed_utils.py
```diff
@@ -0,0 +1,121 @@
...
+from pathlib import Path
+
+import duckdb
+
+from cumulus_library import cli
+from cumulus_library.databases import create_db_backend
+
+
+class LocalTestbed:
+    def __init__(self, path: Path, with_patient: bool = True):
```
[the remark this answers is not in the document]
mikix: I wasn't sure how, tbh

cumulus_library/studies/core/core_templates/patient.sql.jinja
mikix: Switched

PR #203, opened 2024-04-01
core: add completion tracking support

mikix:
This will ignore any encounter rows that haven't yet loaded all the resources we care about.

This completion tracking is opt-in and won't affect legacy data. We include any encounters which don't have a completion group registration.

In addition:
- Add --load-ndjson-dir to the generate-sql command.
- The core study now builds core__incomplete_encounters holding just the encounter IDs of ignored encounters that were deemed incomplete. This is not referenced by other tables - it's just a debugging tool.

This PR does not need to land until completion tracking in the ETL is done. (Though it's fully backwards-compatible and can land now. Any non-tracked encounters will be allowed through.)

I'm not adding any user docs in this PR - I feel like largely, the user instructions will be for the ETL side of things - how to load data in a way that will work for completion tracking. Seem right?

This PR is the main Library-side piece of https://github.com/smart-on-fhir/cumulus-etl/issues/296 (ETL-side [here](https://github.com/smart-on-fhir/cumulus-etl/pull/304))

### Checklist


cumulus_library/studies/core/core_templates/encounter.sql.jinja
```diff
@@ -51,6 +104,12 @@ WITH temp_encounter_nullable AS (
             ) 
         }}
     FROM encounter AS e
+    LEFT JOIN temp_encounter_completion AS tec USING(id)
+    WHERE (
+        -- NULL completion just means it's a row that isn't completion-tracked
+        -- (likely a legacy row), so allow it in.
+        tec.is_complete IS NULL OR tec.is_complete
+    )
```
dogversioning, line 112: i think there's a use case for 'how many of my encounters are complete' that this is filtering out - not sure the best way to handle, just observing.

mikix: We talked about this off-PR - it's a good point that ignored encounters have no one to speak for them with this current state.

So I should probably add a `core__incomplete_encounters` table or some such that can just hold id values - so at least if there is a mistake and someone forgot to load Conditions for a group, you'll know you've got a mistake.

When you're done with your ingestion process, and there are ids in that table, something went wrong.

mikix: This is done now :+1:

cumulus_library/studies/core/core_templates/documentreference.sql.jinja
```diff
@@ -60,7 +60,7 @@ temp_encounters AS (
          unnest(context.encounter) AS context_encounter (encounter) --noqa
 
 {% else %}
-        '' AS encounter_ref
+        tdr.id AS encounter_ref
```
mikix, line 63: I found that athena will error out if you try to make a zero-length varchar? :shrug: So I switched to grabbing the type from the base table

.github/workflows/ci.yaml
```diff
@@ -69,7 +69,7 @@ jobs:
       # See https://github.com/aws-actions/configure-aws-credentials for configuring
       # the aws side of this - the below action is just a light wrapper
       - name: Configure AWS Credentials
-        uses: mcblair/configure-aws-profile-action@v0.1.1
+        uses: mcblair/configure-aws-profile-action@v1.0.0
```
dogversioning, line 72: oh sweet, nice catch

cumulus_library/studies/core/core_templates/documentreference.sql.jinja
dogversioning: could you get to the same behavior with an explicit cast? it might be a little more explicit that way.

cumulus_library/studies/core/reference_sql/builder_encounter.sql
```diff
@@ -531,8 +583,8 @@ temp_encounter AS (
     LEFT JOIN core__encounter_dn_priority AS edt ON e.id = edt.id
     LEFT JOIN core__encounter_dn_servicetype AS eds ON e.id = eds.id
     LEFT JOIN core__encounter_dn_priority AS edp ON e.id = edp.id
-    LEFT JOIN core__encounter_dn_reasonCode AS edr ON e.id = edr.id
-    LEFT JOIN core__encounter_dn_dischargeDisposition AS edd ON e.id = edd.id
+    LEFT JOIN core__encounter_dn_reasoncode AS edr ON e.id = edr.id
+    LEFT JOIN core__encounter_dn_dischargedisposition AS edd ON e.id = edd.id
```
dogversioning, line 587: whew, thank you

cumulus_library/studies/core/reference_sql/builder_patient.sql
```diff
@@ -266,24 +266,33 @@ WITH temp_patient AS (
...
+
+-- Grab a single address to use for this patient, if present
+temp_address AS (
+    SELECT
+        tp.id,
+        
+        
+        
+        substr(max(t_address.addr_row.postalcode), 1, 3) AS postalcode3
+    FROM
```
dogversioning, line 279: making a note to fix this the next time i'm opening a PR

dogversioning (approved):
Looks good - Can you:
- Add a note to the core study docs about what this field represents
- Run `generate-md` on the core study and update the affected tables

cumulus_library/studies/core/reference_sql/builder_patient.sql
mikix: Ahem, that's my fault, sorry

.github/workflows/ci.yaml
mikix: Just happened to see the deprecation warnings in the logs

cumulus_library/studies/core/core_templates/documentreference.sql.jinja
mikix: Yup, that works - will switch.

PR #206, opened 2024-04-16
core: fix typo in core__encounter.serviceType_code*

mikix:
This renames two new fields to core__encounter:
- seviceType_code -> serviceType_code
- seviceType_code_system -> serviceType_code_system

### Checklist


dogversioning (commented):
So - i am mentally compiling a bugfix release. I would be fine taking this as part of that and just make the breaking change now, rather than keep the wrong thing around.

So my request, if you agree:
- set the version to 2.0.1
- remove the typo fields

And then I would probably stop scope creep on bugfix release and just fix the other issue (cnt subject vs cnt_subject_ref) and just cut a release on that.

mikix:
I'm fine with whatever you wanna do here, but that feels like a pretty big semantic versioning cheat.

I'd personally either cut a 3.0 release with the fields cut, or leave the fields in. And I lean toward the latter, the typo fields being left in don't hurt anyone. :shrug:

mikix:
OK, I updated this to match a paper-bag-release style (v2.0.1 with the typo fields removed).

dogversioning:
> OK, I updated this to match a paper-bag-release style (v2.0.1 with the typo fields removed).

i love shame! it's my favorite emotion

PR #210, opened 2024-04-16
core: tighten up core_observation_lab and _vital_signs inclusion

mikix:
Require both these observation sub-type tables to use the proper category code and system.

### Checklist


cumulus_library/studies/core/observation_type.sql
mikix, line 1: A thing I didn't do in this PR, but seems appropriate, is switching `core__observation_vital_signs` from including `valueCodeableConcept` fields to `valueQuantity` fields. I believe quantities are usually used for vital signs, not concepts.

dogversioning (approved):
saying this out loud so it's not just in my head - i think that at some point these two tables should move to a builder, i just didn't get to it for aforementioned release deadlines.

mikix:
> saying this out loud so it's not just in my head - i think that at some point these two tables should move to a builder, i just didn't get to it due to aforementioned release deadlines.

When I add a `component` table (probably in that same file?), I'll add a comment to that effect.

PR #213, opened 2024-04-18
Improved upload error reporting

dogversioning:
- Switches to the builtin response.raise_for_status() function for error generation on rejected uploads for nicer formatting
- Catches the RequestException gracefully rather than dumping a stack trace

### Checklist

### DIFF
--- a/cumulus_library/cli.py
+++ b/cumulus_library/cli.py
@@ -323,7 +323,7 @@ def run_cli(args: dict):
             upload.upload_files(args)
         except requests.RequestException as e:
             print(str(e))
-            exit()
+            sys.exit()
 
     # all other actions require connecting to the database
     else:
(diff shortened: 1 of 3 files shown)


cumulus_library/cli.py
```diff
@@ -318,7 +319,11 @@ def run_cli(args: dict):
         create_template(args["create_dir"])
 
     elif args["action"] == "upload":
-        upload.upload_files(args)
+        try:
+            upload.upload_files(args)
+        except requests.RequestException as e:
+            print(str(e))
+            exit()
```
mikix, line 326: Should be sys.exit, yeah?

mikix (approved):

cumulus_library/cli.py
dogversioning: yep, was just getting to that

PR #220, opened 2024-04-23
feat: support input database without patient extensions

mikix:
Features:
- The core__patient no longer requires the presence of race/ethnicity extensions to build (it now uses the deep-schema checking code)

Bug Fixes:
- If there are both detailed and ombCategory values for an extension, no longer report both - just prefer the ombCategory version.

### Checklist


cumulus_library/studies/core/builder_patient.py
```diff
@@ -18,9 +18,52 @@
...
+            },
+        )
+        if has_extensions:
+            config = sql_utils.ExtensionConfig(
+                source_table="patient",
+                source_id="id",
+                target_table=f"core__patient_ext_{name}",
+                target_col_prefix=name,
+                fhir_extension=url,
+                ext_systems=["ombCategory", "detailed"],
```
mikix, line 53: Note: I dropped `text` from this list, because that uses `valueString`, not `valueCoding` so it would never yield anything anyway, since the jinja code doesn't check `valueString` -- open question as to whether we want to add support for that (`text` is the only actually required component of the extension after all, and all we care about is a display string... but that's for another PR).

cumulus_library/template_sql/extension_denormalize.sql.jinja
```diff
@@ -72,7 +72,7 @@ CREATE TABLE {{ target_table }} AS (
             {%- endif %}
             ROW_NUMBER()
                 OVER (
-                    PARTITION BY id, system
+                    PARTITION BY id
```
mikix, line 75: This is a bug fix, afaict - when you have both ombCategory and detailed, this was partitioning in such a way that both would have available_priority = 1 and both show up in the extension table. Then when you joined with patients, you'd get duplicate patient rows.

tests/regression/reference/core__count_medicationrequest_month.csv
mikix, line 1: Afaict, this changed due to a re-run of the regression ETL on updated sample-bulk-fhir-datasets data (we inlined some medication codes in Aug '23)

dogversioning (approved):

cumulus_library/studies/core/builder_patient.py
mikix: https://github.com/smart-on-fhir/cumulus-library/issues/222

cumulus_library/template_sql/sql_utils.py
```diff
@@ -256,14 +256,14 @@ def is_field_populated(
     return True
 
 
-def _check_schema_if_exists(
+def is_field_present(
```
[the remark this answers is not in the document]
mikix: Sure, will do.

That opens the question of "should any study author be calling any of this python code?" - I had kind of assumed it was all semi-private.

cumulus_library/studies/core/reference_sql/builder_documentreference.sql
```diff
@@ -153,7 +153,6 @@ CREATE TABLE core__documentreference AS
 WITH temp_documentreference AS (
     SELECT DISTINCT
         dr.id,
-        dr.type,
```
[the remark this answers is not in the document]
mikix: Yeah, looks like I forgot to run `generate-sql` at the tail end of my deep-schema-support PR, after review changes.

cumulus_library/template_sql/sql_utils.py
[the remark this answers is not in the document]
dogversioning: I :think: the answer is no, but want to leave at least a breadcrumb trail of 'ok if you're doing something really unusual, here's a subset of functions as reasonable entrypoints'

cumulus_library/template_sql/extension_denormalize.sql.jinja
mikix: We talked about this on slack - we think it _is_ a bug, and will leave the code change in. But we have lots of questions about best approaches for race/ethnicity. Should `race_detailed` be a separate column? What to do with `text`? Problems for another day.

cumulus_library/template_sql/sql_utils.py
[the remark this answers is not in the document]
mikix: Docstring updated to basically say "hey you probably don't need this if you're not core"

PR #223, opened 2024-04-25
UMLS API

dogversioning:
- Adds an API module for talking to UMLS endpoints
- Cuts over to the responses library for mocking (due to requests_mock not well supporting streams)

Out of scope of this PR is integration with the CLI, which will come in a future PR once there is a builder that uses it.

### Checklist


cumulus_library/apis/umls.py
```diff
@@ -0,0 +1,152 @@
...
+        if oid:
+            url = f"{url}/{oid}"
+
+        # If we're inspecting url references in a VSAC response, they come back
+        # specifying a url that does not align with the actual implemented rest
+        # APIs, so we do some massaging
+        if "http:" in url:
+            url = url.replace("http:", "https:")
+        if "/res/" not in url:
+            url = url.replace("/fhir/", "/fhir/res/")
```
mikix, line 87: That's rude

cumulus_library/apis/umls.py
```diff
@@ -0,0 +1,152 @@
...
+        res_json = response.json()
+        if "compose" in res_json and "include" in res_json["compose"]:
+            nested_valuesets = []
+            for record in res_json["compose"]["include"]:
+                if "valueSet" in record:
+                    nested_get = self.get_vsac_valuesets(url=record["valueSet"][0])
+                    nested_valuesets.append(nested_get[0])
+
+            return [response.json(), *nested_valuesets]
+        return response.json()
```
mikix, line 100: I think you want `[response.json()]` here?

Or just skip the forking logic and do something like:
```python
all_valuesets = [response.json()]
records = res_json.get("compose", {}).get("include", [])
for record in records:
    ...
    all_valuesets.append()
return all_valuesets
```

cumulus_library/apis/umls.py
```diff
@@ -0,0 +1,152 @@
...
+                    f.write(chunk)
+                    chunks_read += 1
+                    progress.update(
+                        task,
+                        description=(
+                            f"Downloading {file_meta['fileName']}: "
+                            f"{chunks_read/1000} MB"
+                        ),
+                    )
+        base_utils.unzip_file(path / file_meta["fileName"], path)
```
mikix, line 152: Does this process leave the zip file on disk? Maybe delete it after this? Or unzip as you stream, dunno how trivial the library makes that.

tests/test_data/apis/umls/include_valueset.json
```diff
@@ -0,0 +1,62 @@
+{
+    "--comment--": "this file is an example of a VSAC valueset respsonse which includes references to other value sets. It is part of the public COVID data set, and is otherwise unmodified except for this comment, and changing the included valueset references to point at the single_valueset.json",
```
mikix, line 2: :shakes fist at json:

tests/test_umls_api.py
```diff
@@ -0,0 +1,172 @@
...
+            assert data[i]["id"] == expected_oids[i]
+
+
+@mock.patch.dict(
+    os.environ,
+    clear=True,
+)
+@responses.activate
+def test_download_umls(tmp_path):
+    with open("./tests/test_data/apis/umls/umls.zip", "rb") as download_zip:
```
mikix, line 131: Can you add a comment here or somewhere about what the origin of this zip file is / how to re-generate it if needed or whatever.

mikix (approved):

.gitignore
```diff
@@ -7,6 +7,8 @@ data_export/
 cumulus_library_columns.json
 output.sql
 *generated.md
+MRCONSO.RRF
```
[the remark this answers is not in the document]
dogversioning: [this](https://www.ncbi.nlm.nih.gov/books/NBK9685/#ch03.sec3.3.4) is the licensed 1.7GB file containing all the concept names that is the main driver behind this, which i'm assuming will be very easy to accidentally commit at some point, probably by me.

cumulus_library/apis/umls.py
```diff
@@ -0,0 +1,152 @@
...
+                raise errors.ApiError("No UMLS API key provided")
+        self.api_key = api_key
+        if validator_key is None:
+            validator_key = api_key
+        self.validator_key = validator_key
+
+        auth_payload = {"validatorApiKey": self.validator_key, "apiKey": self.api_key}
+        self.session = requests.Session()
+        response = self.session.get(
+            "https://utslogin.nlm.nih.gov/validateUser", params=auth_payload
```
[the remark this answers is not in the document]
dogversioning: basically my idea here was to deal with the bad key once, rather than having to catch it each time, so we don't have to do a bunch of different flavors of catching elsewhere.

I think this api set is maintained by multiple teams - some places ask for it as a basic auth token, where some ask for it as a request param of varying different names.

i could move this to a function call if you felt strongly about it, but i think this is the DRYest version.

cumulus_library/apis/umls.py
dogversioning: it does leave the zip file on disk.

i could clean this up - i went back and forth on doing that, versus keeping the file around and then checking for it in the dir if it already existed and then skipping the download. this is clearly the worst of both worlds.

tests/test_umls_api.py
dogversioning: yeah, can do.

cumulus_library/apis/umls.py
dogversioning: the data model is sacred and must never be modified

cumulus_library/apis/umls.py
dogversioning: i had [response.json()] here at one point and was getting an extra layer of list at some point. lemme go back and check.

cumulus_library/apis/umls.py
```diff
@@ -0,0 +1,152 @@
...
+        """
+
+        if api_key is None:
+            api_key = os.environ.get("UMLS_API_KEY", None)
+            if api_key is None:
+                raise errors.ApiError("No UMLS API key provided")
+        self.api_key = api_key
+        if validator_key is None:
+            validator_key = api_key
+        self.validator_key = validator_key
```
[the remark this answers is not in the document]
dogversioning: no strong opinion, happy to change

cumulus_library/apis/umls.py
dogversioning: ok, now this removes the zip file at the end.

PR #230, opened 2024-04-30
Add table creation from files

dogversioning:
- Adds a function to databases to upload parquets to S3
- Adds templates for creating tables from external files
- converts vocab study to use database upload

### Checklist


cumulus_library/studies/vocab/.gitignore
```diff
@@ -0,0 +1 @@
+icd/*.parquet
```
dogversioning, line 1: i put this in the study folder just in case we ever decide to break it off into its own repo

cumulus_library/studies/vocab/vocab_icd_builder.py
```diff
@@ -30,49 +31,31 @@ def prepare_queries(self, cursor: object, schema: str, *args, **kwargs):
...
-                        base_templates.get_insert_into_query(
-                            table_name=table_name, table_cols=headers, dataset=dataset
-                        )
-                    )
+        header_types = ["STRING", "STRING", "STRING", "STRING", "STRING"]
+        for file in icd_files:
+            parquet_path = path / f"icd/{file.stem}.parquet"
+            if not parquet_path.is_file():
+                df = pandas.read_csv(file, delimiter="|", names=headers)
+                df.to_parquet(parquet_path)
```
dogversioning, line 42: so i opted to do this step, rather than converting the files to parquet up front and putting those in source control, to make it easier for third parties to understand what was in these files - kind of making the parquet bit an invisible transport mechanism. Could be talked out of it.

tests/regression/run_regression.py
```diff
@@ -20,62 +20,85 @@
 
 from pandas import read_parquet
 
-ref_path = f"{Path(__file__).resolve().parent}/reference"
-export_path = f"{Path(__file__).resolve().parent}/data_export/core"
+from tests import conftest
 
-references = set(os.listdir(ref_path))
```
dogversioning, line 26: The bulk of the changes in this file are just indentation into a function. No changes were made to the logic of the previous regression.

.github/workflows/ci.yaml
```diff
@@ -80,8 +80,8 @@ jobs:
           WG: cumulus
           DB: cumulus_library_regression_db
         run: |
-          cumulus-library build -t core --profile $PROFILE --workgroup $WG --database $DB
-          cumulus-library export -t core ./tests/regression/data_export/ --profile $PROFILE --workgroup $WG --database $DB
+          cumulus-library build -t vocab -t core --profile $PROFILE --workgroup $WG --database $DB
+          cumulus-library export -t vocab -t core ./tests/regression/data_export/ --profile $PROFILE --workgroup $WG --database $DB
```
mikix, line 84: Discovery too? Doesn't need to be this PR, but might as well yeah?

dogversioning:
also noting: file upload means that we're going to need to allow file deletion, in cases where a table needs to be recreated with a new dataset. I don't think it's critical right this second - I'm currently deferring that pending a discussion on the mrconso scope, where i might need to do this. but i'll create a issue regardless if we're ok with ignoring for now.

cumulus_library/databases.py
```diff
@@ -497,3 +499,47 @@ def create_db_backend(args: dict[str, str]) -> DatabaseBackend:
...
+        profile = cursor.connection.profile_name
+
+        if not remote_filename:
+            remote_filename = file
+        session = boto3.Session(profile_name=profile)
+        s3_client = session.client("s3")
+        with open(file, "rb") as b_file:
+            s3_client.put_object(
+                Bucket=bucket,
+                Key=f"user_uploads/{study}/{topic}/{remote_filename}",
```
mikix, line 538: This makes me lightly nervous - we're attempting to own `/user_uploads` on the bucket - that might conflict with whatever bucket setup they have and requires granting upload rights to a whole new space. And prevents admin-level namespacing.

For example, we at BCH have several different stacks set up (cerner, i2b2, test stacks) - this would currently share the same folder for all those.

Can we put `user_uploads` as a subfolder of `OutputLocation`? That way the user permission grant is very close as it was before (just add a little S3 permission on top, but no path changes), and admins can namespace as they like.

cumulus_library/studies/vocab/vocab_icd_builder.py
```diff
@@ -30,49 +31,31 @@ def prepare_queries(self, cursor: object, schema: str, *args, **kwargs):
...
-                    self.queries.append(
-                        base_templates.get_insert_into_query(
-                            table_name=table_name, table_cols=headers, dataset=dataset
-                        )
-                    )
+        header_types = ["STRING", "STRING", "STRING", "STRING", "STRING"]
+        for file in icd_files:
+            parquet_path = path / f"icd/{file.stem}.parquet"
+            if not parquet_path.is_file():
+                df = pandas.read_csv(file, delimiter="|", names=headers)
```
mikix, line 41: Is that really all a bsv file is? I don't even know what it stands for or why it's common in this space. I think it also uses `#` as a header marker? Or at least, that's how `ctakesclient` parses a bsv file.

Is this even a real format?

cumulus_library/studies/vocab/vocab_icd_builder.py
```diff
@@ -30,49 +31,31 @@ def prepare_queries(self, cursor: object, schema: str, *args, **kwargs):
...
-                if rows_processed > 0:
-                    self.queries.append(
-                        base_templates.get_insert_into_query(
-                            table_name=table_name, table_cols=headers, dataset=dataset
-                        )
-                    )
+        header_types = ["STRING", "STRING", "STRING", "STRING", "STRING"]
+        for file in icd_files:
+            parquet_path = path / f"icd/{file.stem}.parquet"
+            if not parquet_path.is_file():
```
mikix, line 40: If you wanted to be slightly fancier, also compare the timestamp on the bsv file against the parquet file.

cumulus_library/studies/vocab/vocab_icd_builder.py
mikix: I buy it

cumulus_library/databases.py
```diff
@@ -497,3 +499,47 @@ def create_db_backend(args: dict[str, str]) -> DatabaseBackend:
...
 
     return backend
+
+
+def upload_file(
+    *,
+    cursor: DatabaseCursor,
+    file: pathlib.Path,
+    study: str,
+    topic: str,
```
mikix, line 509: Will `topic` always be a sensible thing to provide? Maybe allow this to be null or just skip it and prefix the filename for the `vocab` study? But it's also fine to leave here - just felt like a level of organization not all studies will need.

cumulus_library/template_sql/ctas_from_parquet.sql.jinja
```diff
@@ -0,0 +1,24 @@
...
+)
+{#- TODO: we may want to consider an optional partition parameter for
+large tables, though we would need to also run a MSCK REPAIR TABLE query 
+after this table is created to make the data available.
+
+See https://docs.aws.amazon.com/athena/latest/ug/parquet-serde.html
+for more info #}
+STORED AS PARQUET
+LOCATION '{{ remote_location }}'
+tblproperties ("parquet.compression"="SNAPPY");
```
mikix, line 21: Do we need to specify compression? What happens if a researcher gives us a differently-compressed parquet file?

tests/test_vocab.py
```diff
@@ -0,0 +1,55 @@
...
+        )
+    )
+    db = databases.DuckDatabaseBackend(f"{tmp_path}/duck.db")
+    cursor = db.cursor()
+    table_rows, cols = conftest.get_sorted_table_data(cursor, "vocab__icd")
+    expected_cols = ["CUI", "TTY", "CODE", "SAB", "STR"]
+    for col in expected_cols:
+        assert any(col_schema[0] == col for col_schema in cols)
+    for col in cols:
+        assert col[0] in expected_cols
```
mikix, line 49: This might read easier like:
```python
    expected_cols = {"CUI", "TTY", "CODE", "SAB", "STR"}
    found_cols = {col_schema[0] for col_schema in cols}
    assert expected_cols = found_cols
```

tests/test_vocab.py
```diff
@@ -0,0 +1,55 @@
...
+    icd_dir = pathlib.Path("./cumulus_library/studies/vocab/icd/")
+    bsv_row_count = 0
+    for file in icd_dir.glob("*.bsv"):
+        with open(file) as f:
+            bsv_row_count += len(f.readlines())
```
mikix, line 54: You never check this `bsv_row_count` against anything.

tests/test_vocab.py
```diff
@@ -0,0 +1,55 @@
...
+    assert len(table_rows) == conftest.VOCAB_ICD_ROW_COUNT
```
mikix, line 55: Can we also get maybe one spot check line comparison? Like the first & last line or something, just for a sanity check on the actual content.

tests/regression/regression-iam.json
mikix, line 1: :+1:

cumulus_library/databases.py
dogversioning: i was lightly nervous about putting it in the `results` folder, since that's technically owned by athena and i suppose there's nothing to stop athena from wiping that folder out? if we're sure that would be safe, that would be fine.

cumulus_library/studies/vocab/manifest.toml
```diff
@@ -9,3 +9,9 @@ file_names = [
 file_names = [
     "icd_legend.sql",
 ]
+
+# This export is only intended for use as part of unit testing
+[export_config]
+export_list = [
+    "vocab__icd"
+]
```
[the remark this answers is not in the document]
dogversioning: regression does, that's where it's being used.

cumulus_library/studies/vocab/vocab_icd_builder.py
dogversioning: per our side convo, 'bar seperated value' is the format, which originates with that metathesaurus.

cumulus_library/studies/vocab/vocab_icd_builder.py
dogversioning: i think, since we just learned we're deprecating this study in the next PR I'm filing, I'm going to skip this.

cumulus_library/template_sql/ctas_from_parquet.sql.jinja
dogversioning: we do, and it specifically needs to be this one and only this one. see [supported serdes](https://docs.aws.amazon.com/athena/latest/ug/supported-serdes.html). I think the deal is that the underlying serde may support multiple compressions, but AWS supports one. But this is about serializing the data from the parquet to the compressed table, not an artifact of the provided parquet itself.

tests/regression/regression-iam.json
dogversioning: This had actually gotten :way: out of date, even prior to the thing breaking regression.

tests/test_vocab.py
dogversioning: ah, i was, before i replaced it with the constant below

tests/test_vocab.py
dogversioning: yes, meant to, will add.

cumulus_library/databases.py
dogversioning: ah what a good point!

cumulus_library/studies/vocab/manifest.toml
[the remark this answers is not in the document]
mikix: Ah gotcha - then maybe just update the comment from "unit testing" to "the CI regression test" or something

cumulus_library/databases.py
mikix: I think this issue was the only thing that made me suss in this PR - I'll approve to get out of you way if we're namespacing this.

mikix (approved):

cumulus_library/databases.py
mikix: For extra future proofing, should we name the subfolder `cumulus` or something? That way we could throw things that aren't user uploads in there at some point too, if we find a need. Don't feel strongly about this, and we could always take a different name in there at that point.

dogversioning: yeah, i'm going to change it to `cumulus_user_uploads`, and I am also going to add the db name to this path.

PR #234, opened 2024-05-10
Small CLI fixes

mikix:
A couple args were broken in the 2.1 release:
--umls-key (env var would work, but not CLI arg)
--builder

### Checklist

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -126,8 +126,6 @@ def test_count_builder_mapping(mock_path, tmp_path):
                 "study_python_counts_valid",
                 "-s",
                 "./tests/test_data",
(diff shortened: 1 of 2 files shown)


tests/test_cli.py
```diff
@@ -126,8 +126,6 @@ def test_count_builder_mapping(mock_path, tmp_path):
                 "study_python_counts_valid",
                 "-s",
                 "./tests/test_data",
-                "--database",
-                "test",
```
mikix, line 130: This and similar changes is because duckdb_args provides --database already (and in a file that's properly in tmp_path), and this extra --database was confusing me until I figured that out.

PR #235, opened 2024-05-10
Make it easier to run studies from git checkouts

mikix:
- When loading a builder, we add the study dir's parent folder into sys.path, so that builders can import helper code. This is basically equivalent to doing PYTHONPATH=. from a checkout.
- When searching for the study dir, and we find a manifest.toml, actually read the manifest to find the study_prefix instead of using the directory name for the study name. This allows more natural python package layouts (rather than `module_name/study_name`, you can just do `module_name`)

This also bumps ruff's version to 0.4.4 and has the commit hook just auto-commit any formatting changes.

### Checklist

### DIFF
--- a/cumulus_library/cli.py
+++ b/cumulus_library/cli.py
@@ -318,14 +318,18 @@ def get_study_dict(alt_dir_paths: list) -> dict[str, pathlib.Path] | None:
     return manifest_studies
 
 
-def get_studies_by_manifest_path(path: pathlib.Path) -> dict:
+def get_studies_by_manifest_path(path: pathlib.Path) -> dict[str, pathlib.Path]:
     """Recursively search for manifest.toml files from a given path"""
     manifest_paths = {}
     for child_path in path.iterdir():
         if child_path.is_dir():
             manifest_paths.update(get_studies_by_manifest_path(child_path))
         elif child_path.name == "manifest.toml":
-            manifest_paths[path.name] = path
+            try:
+                manifest = study_parser.StudyManifestParser(path)
+                manifest_paths[manifest.get_study_prefix()] = path
+            except errors.StudyManifestParsingError as exc:
+                rich.print(f"[bold red] Ignoring study in '{path}': {exc}")
     return manifest_paths
 
 
--- a/cumulus_library/template_sql/sql_utils.py
+++ b/cumulus_library/template_sql/sql_utils.py
@@ -8,7 +8,7 @@ simply. This includes, but is not limited, to:
     - Data with deep missing elements
     - Data which may or may not be in an array depending on context
 """
-import abc
+
 from dataclasses import dataclass, field
 
 from cumulus_library import base_utils, databases
@@ -26,8 +26,8 @@ REFERENCE = ["reference"]
 
 
 @dataclass(kw_only=True)
-class BaseConfig(abc.ABC):
-    """Abstract ase class for handling table detection/denormalization"""
+class BaseConfig:
+    """Base class for handling table detection/denormalization"""
 
     source_table: str = None
     source_id: str = "id"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,7 +31,8 @@ classifiers = [
 dynamic=["version"]
 [project.optional-dependencies]
 dev = [
-    "ruff == 0.2.1",
+    # if you update the ruff version, also update .pre-commit-config.yaml
+    "ruff == 0.4.4",
     "pre-commit",
 ]
 test = [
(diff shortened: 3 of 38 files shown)


cumulus_library/template_sql/sql_utils.py
```diff
@@ -26,8 +26,8 @@
 
 
 @dataclass(kw_only=True)
-class BaseConfig(abc.ABC):
-    """Abstract ase class for handling table detection/denormalization"""
+class BaseConfig:
+    """Base class for handling table detection/denormalization"""
```
mikix, line 30: ruff complained about this - there are no abstract methods or properties now that this is a simple dataclass.

pyproject.toml
```diff
@@ -31,7 +31,7 @@ classifiers = [
 dynamic=["version"]
 [project.optional-dependencies]
 dev = [
-    "ruff == 0.2.1",
+    "ruff",
```
mikix, line 34: I chose to unpin this here, maybe I shouldn't have.

My thinking was:
- The commit hook has its own checkout of ruff, so the developer's version **could** diverge, though they may see different errors if they manually run ruff.
- But I liked that oddity more than I liked developers having to manage different packages wanting different ruff versions.
- Thoughts?

pyproject.toml
```diff
@@ -62,9 +62,6 @@ testpaths = [
     "tests",
 ]
 
-[tool.ruff]
-target-version = "py310"
```
mikix, line 66: Looking at the docs for this, ruff will look at `requires-python` and actually recommends just setting that instead of this.

pyproject.toml
mikix: Hmm, though then what version should the CI install...? Should that be hardcoded in the CI to match the hook? Or should we hardcode the version here in the pyproject.toml... hmmm.... Maybe I should leave this alone. But I don't love it.

cumulus_library/cli.py
```diff
@@ -318,14 +318,18 @@ def get_study_dict(alt_dir_paths: list) -> dict[str, pathlib.Path] | None:
...
     for child_path in path.iterdir():
         if child_path.is_dir():
             manifest_paths.update(get_studies_by_manifest_path(child_path))
         elif child_path.name == "manifest.toml":
-            manifest_paths[path.name] = path
+            try:
+                manifest = study_parser.StudyManifestParser(path)
+                manifest_paths[manifest.get_study_prefix()] = path
+            except errors.StudyManifestParsingError as exc:
+                rich.print(f"[bold red] Ignoring study in '{path}': {exc}")
```
mikix, line 332: I did not offer any kind of compatibility / migration for the old way of doing things. Like I _could have_ also added the directory name into `manifest_paths`.

But... I figured any existing studies (outside of some of our test data studies) would have realistically needed to have the `study_prefix` match the directory already. If you didn't have that setup, that meant that your `--target` name didn't match your `study_prefix`, and that seems like something we don't really want to encourage / support?

tests/test_data/study_python_local_template/module1.py
```diff
@@ -1,5 +1,6 @@
+from study_python_local_template import local_template
+
 from cumulus_library.base_table_builder import BaseTableBuilder
-from tests.test_data.study_python_local_template import local_template
```
mikix, line 2: This was done to test the new `sys.path` injection

mikix:
I didn't see any changes needed in the [template repo](https://github.com/smart-on-fhir/cumulus-library-template/), mostly because (1) it doesn't include any builders that try to import anything, (2) it doesn't use the two-level directory setup that some released studies like covid have had to use, and (3) its instructions talk about renaming the study folder already to match a python module name and to also rename study_prefix.

pyproject.toml
dogversioning: I started doing explicit pins on these in response to a SQLFluff linting error that caused malformed code in some cases. I think the astral folks are less likely to make the same mistake, but i sort of like 'a human has decided to increment the version on the autoformatting tools'. could be talked out of this.

dogversioning:
> I didn't see any changes needed in the [template repo](https://github.com/smart-on-fhir/cumulus-library-template/), mostly because (1) it doesn't include any builders that try to import anything, (2) it doesn't use the two-level directory setup that some released studies like covid have had to use, and (3) its instructions talk about renaming the study folder already to match a python module name and to also rename study_prefix.
> 
> So I think it should be fine both before and after this change, but your naming can be more flexible after this PR

We should maybe chat about whether maintaining the template repo even makes sense - The example config file in the docs might be the 'right' way to telegraph how to make studies at this point.

dogversioning (approved):

PR #236, opened 2024-05-13
fix: if a table does not exist, use an empty schema

mikix:
(Instead of erroring out.)

This helps in cases where some tables get optionally created (like the ETL completion tables), because Athena generates an exception for that.

### Checklist

### DIFF
--- a/cumulus_library/template_sql/sql_utils.py
+++ b/cumulus_library/template_sql/sql_utils.py
@@ -116,7 +116,7 @@ def _check_data_in_fields(
 
     with base_utils.get_progress_bar(transient=True) as progress:
         task = progress.add_task(
-            "Detecting available encounter codeableConcepts...",
+            "Detecting available codeableConcepts...",
             # Each column in code_sources requires at most 3 queries to
             # detect valid data is in the DB
             total=len(code_sources),
@@ -209,7 +209,12 @@ def validate_schema(
     validated_schema = {}
     for table, cols in expected_table_cols.items():
         query = base_templates.get_column_datatype_query(schema, table, cols.keys())
-        table_schema = cursor.execute(query).fetchall()
+        try:
+            table_schema = cursor.execute(query).fetchall()
+        except Exception:
+            # The database might reasonably raise an exception in cases like
+            # the table not existing. Here we treat error tables as no-column tables.
+            table_schema = []
         validated_schema[table] = parser.validate_table_schema(cols, table_schema)
     return validated_schema
 
--- a/cumulus_library/template_sql/sql_utils.py
+++ b/cumulus_library/template_sql/sql_utils.py
@@ -90,9 +90,7 @@ class ExtensionConfig(BaseConfig):
 
 
 def _check_data_in_fields(
-    schema: str,
-    cursor: databases.DatabaseCursor,
-    parser: databases.DatabaseParser,
+    database: databases.DatabaseBackend,
     code_sources: list[CodeableConceptConfig],
 ) -> dict:
     """checks if CodeableConcept fields actually have data available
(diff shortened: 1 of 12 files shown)


cumulus_library/template_sql/sql_utils.py
```diff
@@ -116,7 +116,7 @@ def _check_data_in_fields(
 
     with base_utils.get_progress_bar(transient=True) as progress:
         task = progress.add_task(
-            "Detecting available encounter codeableConcepts...",
+            "Detecting available codeableConcepts...",
```
mikix, line 119: Unrelated, but just noticed this - seems like a copy and paste from encounter-specific checks to generic checks, but kept the encounter wording.

cumulus_library/template_sql/sql_utils.py
```diff
@@ -209,7 +209,12 @@ def validate_schema(
     validated_schema = {}
     for table, cols in expected_table_cols.items():
         query = base_templates.get_column_datatype_query(schema, table, cols.keys())
-        table_schema = cursor.execute(query).fetchall()
+        try:
+            table_schema = cursor.execute(query).fetchall()
+        except Exception:
```
dogversioning, line 214: i know this is a pain, but: can we catch a more specific error here? Two cases that should error out here are 'something happened to your network connection' and 'the query you sent was malformed in some way'.

mikix: OK I can look into which specific error it is - probably a generic Athena error that I need to inspect the code of.

However - that's injecting some database-specific knowledge into this generic code. I could maybe add a method like "tell me if this exception is SomeKindOfError" to DatabaseBackend? Or just live with the boundary bleeding.

dogversioning: i the the 'put it in the databasebackend' approach is probably the right one, but if it turns into A Whole Thing we could live with it for now and revisit.

mikix: I'll try that approach and see

mikix: OK added a new commit for easier reading.

I did three things:
- Unrelated syntax fix for an excess comma that I noticed
- Changed all the `core` builders to just grab `config` in `prepare_queries` so they could get direct access to the database backend.
- Added a DatabaseBackend.operational_errors() method to surface which exception classes are "operational errors" -- this is the abstraction layer I happened to choose, I'm not rooting for it hard and am open to push back (i.e. I could imagine a very specific method like "is_table_not_found_error()" or a generic method like "exception_to_error_enum()" - but this just an easy & quick path for now). In Athena, network errors and syntax errors would both be `DatabaseError`. There's also stuff like `ProgrammingError` or `NotSupportedError`. `OperationalError` seems to be used in both pyAthena and DuckDB for stuff like "out of memory" or "table already exists" or "table doesn't exist" kind of stuff.

dogversioning (approved):
Thanks for doing the fussier version of this fix - i think it also starts getting us ahead on 'getting rid of the schema/cursor direct passing', which is nice.

cumulus_library/template_sql/sql_utils.py
```diff
@@ -201,24 +195,31 @@ def denormalize_complex_objects(
 
 
 def validate_schema(
-    cursor: databases.DatabaseCursor,
-    schema: str,
+    database: databases.DatabaseBackend,
     expected_table_cols: dict,
-    parser: databases.DatabaseParser,
```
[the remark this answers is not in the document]
mikix: The data metrics study currently does some of this itself:
```
        for table, cols in fields_to_check.items():
            query = base_templates.get_column_datatype_query(schema, table.lower(), cols.keys())
            cursor.execute(query)
            table_schema = cursor.fetchall()
            self.schemas[table] = parser.validate_table_schema(cols, table_schema)
```

- But I can switch it to the sql_utils methods, if we think those are more stable/public. (and it would help get fixes like this empty table thing)

Issue #184, opened 2024-02-13
tech debt: replace LIKE with regexp_like


dogversioning:
For performance reasons, this will be faster. We should also only use like if we have to (currently, just for matching cerner UID strings).

duckdb equivalent is regexp_matches.

PR #237, opened 2024-05-14
Replaced wildcard match with regex

dogversioning:
This PR addresses #184 by adding a macro for wildcard conversion and switching template queries over to using regexp_like.

### Checklist

### DIFF
--- a/cumulus_library/databases.py
+++ b/cumulus_library/databases.py
@@ -13,6 +13,7 @@ import datetime
 import json
 import os
 import pathlib
+import re
 import sys
 from pathlib import Path
 from typing import Any, Protocol
@@ -342,6 +343,13 @@ class DuckDatabaseBackend(DatabaseBackend):
             duckdb.typing.VARCHAR,
         )
         self.connection.create_function(
+            # DuckDB's version is regexp_matches.
+            "regexp_like",
+            self._compat_regexp_like,
+            None,
+            duckdb.typing.BOOLEAN,
+        )
+        self.connection.create_function(
             # We frequently use Athena's date() function because it's easier than
             # the more widely-supported way of CAST(x AS DATE).
             # Rather than convert all of our SQL to the longer version,
@@ -392,6 +400,15 @@ class DuckDatabaseBackend(DatabaseBackend):
         if value is None:
             return None
         return delimiter.join(v for v in value if v is not None)
+
+    @staticmethod
+    def _compat_regexp_like(string: str | None, pattern: str | None) -> bool:
+        if string is None or pattern is None:
+            return False
+        match = re.search(pattern, string)
+        if match is None:
+            return False
+        return True
 
     @staticmethod
     def _compat_date(
--- a/cumulus_library/studies/core/core_templates/medication.sql.jinja
+++ b/cumulus_library/studies/core/core_templates/medication.sql.jinja
@@ -1,4 +1,5 @@
 {% import 'core_utils.jinja' as utils -%}
+{% import 'syntax.sql.jinja' as syntax -%}
 
 CREATE TABLE core__medication AS (
     WITH
@@ -32,7 +33,9 @@ CREATE TABLE core__medication AS (
             mr.id,
             substring(mr.med_ref, 2) AS medication_id
         FROM mr_basics AS mr
-        WHERE mr.med_ref IS NOT NULL AND mr.med_ref LIKE '#%'
+        WHERE mr.med_ref IS NOT NULL AND regexp_like(
+            mr.med_ref, '{{ syntax.wildcard_to_regex('#%') }}'
+        )
     ),
 
     external_refs AS (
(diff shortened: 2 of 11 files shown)


cumulus_library/template_sql/codeable_concept_denormalize.sql.jinja
```diff
@@ -79,7 +79,7 @@ CREATE TABLE {{ target_table }} AS (
             UNNEST(s.{{ field_alias }}.coding) AS u (coding)
         {%- if filter_priority %}
         WHERE
-            u.coding.system LIKE '{{ system }}'
+            REGEXP_LIKE(u.coding.system, '{{ syntax.wildcard_to_regex(system) }}')
```
dogversioning, line 82: This instance and the one in coding are all caps due to sqlfluff enforcement picking up style from the unnest. Elsewhere it complained about it being lowercase. This is probably an issue with the athena sqlfluff dialect.

cumulus_library/template_sql/shared_macros/syntax.sql.jinja
```diff
@@ -17,4 +17,16 @@
 UNION
 {%- endif -%}
 {%- endmacro -%}
+
+{#- The intent of this function is to allow an end user to work in SQL for partial
+string searches, rather than regex. So for example, a value of macro_string like:
```
mikix, line 22: This docstring could use a link to the regex syntax being used. Like the [trino docs](https://trino.io/docs/current/functions/regexp.html) I guess (and/or the java docs it then links to).

cumulus_library/template_sql/shared_macros/syntax.sql.jinja
```diff
@@ -17,4 +17,16 @@
...
+
+https://fhir.cerner.com/%/codeSet/71
+
+will be converted into a regex-escaped pattern with a regex match anything pattern:
+
+https:\/\/fhir\.cerner\.com\/(.*)\/codeSet\/71 -#}
+{%- macro wildcard_to_regex(match_string) -%}
+{{ match_string.replace("/", "\\/").replace(".", "\\.").replace("%", "(.*)") }}
+{%- endmacro -%}
```
mikix, line 31: - You must add `^` and `$` at ends of the resulting string to match wildcard semantics (`LIKE` is a direct match, `regexp_like()` is a search).
- Why are we escaping forward slashes? Those don't seem to be part of the regex syntax.
- Though we should probably be escaping backslashes, dollar signs, and a whole bunch of stuff if you're paranoid. But since we control the source patterns (so far), we can get away without it. It's just annoying that trino doesn't provide a "make this string a regex literal kthx" method.
- Why the parens around `.*`? - I'm used to that being used for grouping but you're not referring back to the group here.
- Totes optional, but you could return an actual string (i.e. add `'` around the result) so every caller doesn't have to do that.

tests/test_base_templates.py
```diff
@@ -95,7 +96,7 @@ def test_codeable_concept_denormalize_filter_creation():
             source AS s,
             UNNEST(s.code_col.coding) AS u (coding)
         WHERE
-            u.coding.system LIKE 'http://snomed.info/sct'
+            REGEXP_LIKE(u.coding.system, 'http:\\/\\/snomed\\.info\\/sct')
```
mikix, line 99: nit: maybe make your eyeballs bleed less with a raw string.

mikix (approved):

cumulus_library/studies/core/core_templates/medication.sql.jinja
```diff
@@ -32,15 +33,19 @@ CREATE TABLE core__medication AS (
             mr.id,
             substring(mr.med_ref, 2) AS medication_id
         FROM mr_basics AS mr
-        WHERE mr.med_ref IS NOT NULL AND mr.med_ref LIKE '#%'
+        WHERE mr.med_ref IS NOT NULL AND regexp_like(
+            mr.med_ref, '{{ syntax.wildcard_to_regex('#%') }}'
```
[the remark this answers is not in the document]
dogversioning: yeah - my goal was to make regex an implementation detail, rather than something a study author needs.

I dig a util method - i'll update for that.

tests/test_base_templates.py
dogversioning: i started there - jinja chokes on it, but for this test it could be ok, it just makes updating this for changes a bit fussy.

cumulus_library/template_sql/shared_macros/syntax.sql.jinja
dogversioning: 1,4,5 - makes sense, will do.
2 - well i was looking at this in a regex debugger and apparently had PHP syntax on, which was the default, which... anyway, it seems to work both ways, but i'll change it.
3 - We could go crazy on this one, but yeah, since this is code systems right now, i scoped to fit, but if there's something you're feeling paranoid about, now's as good a time as any.

cumulus_library/databases.py
```diff
@@ -341,6 +342,13 @@ def __init__(self, db_file: str):
             None,
             duckdb.typing.VARCHAR,
         )
+        self.connection.create_function(
+            # DuckDB's version is regexp_matches.
```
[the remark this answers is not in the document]
dogversioning: SQL is short for 'Syntax Questionably Levereaged'

cumulus_library/template_sql/shared_macros/syntax.sql.jinja
mikix: Naw I don't feel strongly on 3 - though....  we should almost certainly escape the single quote character `'`. I think as long as you do that, we should avoid truly disastrous outcomes and merely have the wrong matches if a string goes sideways.

mikix: Or mabye jinja auto-escapes quotes like that for us, I forget.

PR #238, opened 2024-05-14
Add data-metrics to list of studies

mikix:
(And add umls/data_metrics to study allowlist.)

### Checklist

### DIFF
--- a/cumulus_library/module_allowlist.json
+++ b/cumulus_library/module_allowlist.json
@@ -2,6 +2,8 @@
     "desc" : 
         "this dict maps pip installable studies to expected paths inside site-packages",
     "allowlist" : {
-        "covid_symptom" : "cumulus_library_covid/covid_symptom"        
+        "covid_symptom" : "cumulus_library_covid/covid_symptom",
+        "data_metrics" : "cumulus_library_data_metrics",
+        "umls" : "cumulus_library_umls/umls"
     }
 }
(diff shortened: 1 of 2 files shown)


cumulus_library/module_allowlist.json
```diff
@@ -2,6 +2,8 @@
     "desc" : 
         "this dict maps pip installable studies to expected paths inside site-packages",
     "allowlist" : {
-        "covid_symptom" : "cumulus_library_covid/covid_symptom"        
+        "covid_symptom" : "cumulus_library_covid/covid_symptom",
+        "data_metrics" : "cumulus_library_data_metrics",
+        "umls" : "cumulus_library_umls/umls"
```
mikix, line 7: Data Metrics isn't available yet on pip, but it will be soon. I'm using the no-subdir pathing for it, which will require a library release to make happen (but that's planned I believe).

I figured I'd also add umls here (with the subdir, since that's what it uses now).

No other studies are in pypi, yeah?

docs/study-list.md
```diff
@@ -18,6 +18,7 @@ for more details and any associated publications.
 - [Suicidality length of stay](https://github.com/smart-on-fhir/cumulus-library-suicidality-los)
 
 ### 2024
+- [Data Metrics](https://github.com/smart-on-fhir/cumulus-library-data-metrics)
```
mikix, line 21: Alphabetic inside the year, or still always append?

dogversioning: oh god. i had been doing chronological (which just happened to be the same) but maybe alpha is better?

dogversioning (approved):

docs/study-list.md
mikix: I guess it matters how important the chronology is?

mikix: Switched to pure chronological to keep with your intent

PR #244, opened 2024-05-29
Support a flat ndjson dir layout as well as a hierarchical one

mikix:
Previously, --load-ndjson-dir would only look for an ETL output-style format like this:

`dir/condition/*.ndjson`
`dir/patient/*.ndjson`

But now it will also look for flat files as well (i.e. ETL input-style format) like this:

dir/1.Patient.ndjson
dir/Patient.october.ndjson
dir/Patient.ndjson

This will make it nicer to use the --load-ndjson-dir flow when you are working on ndjson files directly, without going through Cumulus ETL first.

### Checklist

### DIFF
--- a/cumulus_library/databases.py
+++ b/cumulus_library/databases.py
@@ -515,21 +515,37 @@ class DuckDbParser(DatabaseParser):
         return parsed
 
 
-def _read_rows_from_table_dir(path: str) -> list[dict]:
-    # Grab filenames to load (ignoring .meta files and handling missing folders)
-    folder = Path(path)
-    filenames = []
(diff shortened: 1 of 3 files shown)


dogversioning (approved):
I think this needs some documentation tweaks for the public site (mostly here: https://docs.smarthealthit.org/cumulus/library/creating-studies.html#testing-studies), but otherwise seems totally reasonable.

mikix:
> I think this needs some documentation tweaks for the public site

Added a new commit to update that page (the only doc page that mentioned --load-ndjson-dir) to basically drop the ETL step entirely.

PR #245, opened 2024-05-31
Seperated CLI actions from study_parser

dogversioning:
- Moves the logic of the various CLI actions into their own files
  - Light refactoring to make actions functional rather than class based
- renamed a lot of vars to more explictly indicate what things are parsing a manifest specifically

### Checklist


cumulus_library/parsers/fhir_valueset.py
dogversioning, line 1: I deleted this file as it has been supplanted by a combination of the TableBuilder and the VSAC API module.

I :could: move various parsers into this location.

dogversioning (commented):
i called all these thing 'actions' but could be talked into a different name for them.

cumulus_library/study_parser.py
```diff
@@ -1,51 +1,14 @@
...
-    orig_path = list(sys.path)
-    try:
-        sys.path.insert(0, str(add))
-        yield
-    finally:
-        sys.path = orig_path
+from cumulus_library import errors
 
 
 class StudyManifestParser:
```
mikix, line 10: tiny nit: it's a little verbose to me to see this `Parser` terminology all around the place. Like, yes this parses a manifest as an implementation detail, but the object itself represents the result - you're really passing around a `StudyManifest` object from function to function. When I see an object or method with a parser name, I kind of expect it to be stage 1 that then spits out a Manifest object result (or just have the parsing be a quiet detail of the class - like `.from_file(path)` or `__init__(path)`

cumulus_library/actions/builder.py
mikix, line 1: I'm assuming all these code moves don't include any functional changes, that this is just moving code around?

cumulus_library/cli.py
```diff
@@ -19,13 +19,13 @@
...
     study_parser,
-    upload,
 )
+from cumulus_library.actions import builder, cleaner, exporter, file_generator, uploader
 from cumulus_library.template_sql import base_templates
 
 
 class StudyRunner:
-    """Class for managing cursors and executing Cumulus queries"""
+    """Class for managing cursors and invoking actions"""
```
mikix, line 28: nit: I think there's opportunity here for even further isolation of the actions - like this file still holds a lot of info about what each action needs, which feels like it's breaking the abstraction a bit. In theory, the top level could just be "routing to a single action method which does the further CLI parsing" and "shared utility methods for CLI parsing". But that's not an important refactor. This is already quite a bit better.

mikix (approved):
> i called all these thing 'actions' but could be talked into a different name for them.

Actions is fine! In an argparse sense, they are "subcommands" - in the ETL I usually use the word "command" for this kind of thing.

cumulus_library/actions/builder.py
dogversioning: yeah there are no functional changes, just reorg + var name changes

dogversioning:
> Actions is fine! In an argparse sense, they are "subcommands" - in the ETL I usually use the word "command" for this kind of thing.

I thought about reusing subcommands, but since the CLI subcommands -> actions mapping is not exactly 1-1 (some clean is invoked before a build), i liked a little seperation.

cumulus_library/cli.py
dogversioning: yeah, i don't disagree - i drew a line in the sand here for a couple reasons:
- i like two medium changes over one big one for something of this scope/keeping context in my brain
- the `lib_transactions` stuff currently is kind of inline here in a way that doesn't decouple well - there probably should be a dedicated module for managing that

But i generally think this is a good thing to do at some point in the future.

cumulus_library/study_parser.py
dogversioning: hmm - not the way my brain usually works, but i see your point. That's a breaking change, though, so i'll put an issue in the 3.0 bucket for it.

Issue #247, opened 2024-05-31
Allow for study-specific arguments


dogversioning:
It would be good to have a loose dict in the StudyConfig class that can have key/value pairs added via a CLI arg, so that study authors can tweak behavior that may be unique to their specific use case.

dogversioning:
addressed via #248

PR #248, opened 2024-06-05
Added import action

dogversioning:
This adds the following:
- A new subcommand/action for importing previously exported studies
- Some alphasorting of large function blocks
- minor: made all custom exceptions children of CumulusLibraryException
- Added a custom argument param #247 

### Checklist

### DIFF
--- a/cumulus_library/actions/cleaner.py
+++ b/cumulus_library/actions/cleaner.py
@@ -2,7 +2,7 @@ import sys
 
 from rich.progress import Progress, TaskID
 
-from cumulus_library import base_utils, databases, enums, study_parser
+from cumulus_library import base_utils, databases, enums, errors, study_parser
 from cumulus_library.template_sql import base_templates
 
 
@@ -85,14 +85,21 @@ def clean_study(
 ) -> list:
     """Removes tables beginning with the study prefix from the database schema
 
+    :param manifest_parser: a StudyManifestParser for the study
     :param cursor: A DatabaseCursor object
     :param schema_name: The name of the schema containing the study tables
-    :verbose: toggle from progress bar to query output, optional
+    :keyword verbose: toggle from progress bar to query output, optional
+    :keyword stats_clean: will delete stats tables created from random sampling
+    :keyword prefix: override manifest-based prefix discovery with the provided prefix
     :returns: list of dropped tables (for unit testing only)
-    :prefix: override prefix discovery with the provided prefix
+
     """
     if not schema_name:
-        raise ValueError("No database provided")
+        raise errors.CumulusLibraryError("No database provided")
+    if not prefix and not manifest_parser:
+        raise errors.CumulusLibraryError(
+            "Either a manifest parser or a filter prefix must be provided"
+        )
     if not prefix:
         drop_prefix = f"{manifest_parser.get_study_prefix()}__"
         display_prefix = manifest_parser.get_study_prefix()
--- a/cumulus_library/actions/cleaner.py
+++ b/cumulus_library/actions/cleaner.py
@@ -76,7 +76,7 @@ def get_unprotected_stats_view_table(
 
 
 def clean_study(
-    manifest_parser: study_parser.StudyManifestParser,
+    manifest_parser: study_parser.StudyManifestParser | None,
     cursor: databases.DatabaseCursor,
     schema_name: str,
     stats_clean: bool = False,
--- a/cumulus_library/actions/importer.py
+++ b/cumulus_library/actions/importer.py
@@ -14,6 +14,9 @@ def create_table_from_parquet(archive, file, study_name, db, schema):
         parquet_path = pathlib.Path(
             archive.extract(file), path=tempfile.TemporaryFile()
         )
+        # While convenient to access, this exposes us to panda's type system,
+        # which is messy - this could be optionally be replaced by pyarrow if it
+        # becomes problematic.
         table_types = pandas.read_parquet(parquet_path).dtypes
         remote_types = db.col_parquet_types_from_pandas(table_types.values)
         s3_path = db.upload_file(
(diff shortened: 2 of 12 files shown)


cumulus_library/actions/cleaner.py
```diff
@@ -76,7 +76,7 @@ def get_unprotected_stats_view_table(
 
 
 def clean_study(
-    manifest_parser: study_parser.StudyManifestParser,
+    manifest_parser: study_parser.StudyManifestParser | None,
```
dogversioning, line 79: this allows the importer to not have to worry about setting up an empty parser.

cumulus_library/cli_parser.py
dogversioning, line 1: changes in this file not related to import/custom args are just alphasorting.

cumulus_library/errors.py
dogversioning, line 1: These are also alphasorted in addition to now being child classes

cumulus_library/actions/importer.py
```diff
@@ -0,0 +1,81 @@
...
+import pandas
+
+from cumulus_library import base_utils, errors
+from cumulus_library.actions import cleaner
+from cumulus_library.template_sql import base_templates
+
+
+def create_table_from_parquet(archive, file, study_name, db, schema):
+    try:
+        parquet_path = pathlib.Path(archive.extract(file))
```
mikix, line 13: Where does this extract to? `pwd`? Feels like we should point it at a temp folder/file.

cumulus_library/actions/importer.py
```diff
@@ -0,0 +1,81 @@
...
+        files = [file for file in files if file.endswith(".parquet")]
+    except zipfile.BadZipFile as e:
+        raise errors.StudyImportError(
+            f"File {archive_path} is not a valid archive."
+        ) from e
+    if len(files) == 0:
+        raise errors.StudyImportError(
+            f"File {archive_path} does not contain any tables."
+        )
+    study_name = files[0].split("__")[0]
```
mikix, line 56: nit: there isn't a sanity check that there actually is a `__` in that filename. Not super likely, but the surrounding code is pretty paranoid.

tests/test_duckdb.py
```diff
@@ -77,7 +77,6 @@ def test_duckdb_load_ndjson_dir(tmp_path):
         "A.Patient.ndjson": False,
         "1.Patient.ndjson": True,
         "Patient.ndjson": True,
-        "patient.ndjson": False,
```
mikix, line 80: RIP

tests/test_cli.py
```diff
@@ -575,3 +575,29 @@ def test_cli_finds_study_from_manifest_prefix(tmp_path):
...
+
+@mock.patch.dict(os.environ, clear=True)
+@pytest.mark.parametrize(
+    "c_arg,raises",
+    [
+        ("foo:bar", does_not_raise()),
+        ("foo", pytest.raises(SystemExit)),
+    ],
+)
+def test_cli_custom_args(tmp_path, c_arg, raises):
```
mikix, line 588: nit: would be nice to confirm the value gets parsed and sent down all the way to the study config too.

mikix (approved):

cumulus_library/actions/cleaner.py
dogversioning: It does, but only if `prefix` is not supplied.

I missed updating the docstring here - i'll do that to clarify, and maybe and an error if both of these are null.

cumulus_library/cli_parser.py
```diff
@@ -2,37 +2,34 @@
...
 
 
-def add_study_dir_argument(parser: argparse.ArgumentParser) -> None:
-    """Adds --study-dir arg to a subparser"""
+def add_custom_argument(parser: argparse.ArgumentParser) -> None:
     parser.add_argument(
-        "-s",
-        "--study-dir",
+        "-c",
+        "--custom-arg",
```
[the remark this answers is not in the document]
dogversioning: i don't - i went back and forth on this several times and was never thrilled. i'll change this to option.

cumulus_library/databases.py
```diff
@@ -168,6 +169,37 @@ def operational_errors(self) -> tuple[Exception]:
         """
         return ()
 
+    def col_parquet_types_from_pandas(self, field_types: list) -> list:
```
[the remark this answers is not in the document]
dogversioning: > Also... it's rude that Athena can't just grab the types from the parquet file it has sitting there.

GOD right?

I tried this with pyarrow and there was some aspect of it that i had an issue with, though i don't remember the exact details - something about the way I was accessing the types ended up in a very deep set of nested objects that felt like overkill, but perhaps i missed another option. i'll take a look at the fhir-support approach.

cumulus_library/actions/importer.py
dogversioning: good call, will change.

cumulus_library/actions/importer.py
dogversioning: what's a little more paranoia between friends? will add.

tests/test_cli.py
dogversioning: fair point - i can mock something out

cumulus_library/actions/cleaner.py
mikix: Ah yeah doh - I missed its use :smile:

cumulus_library/databases.py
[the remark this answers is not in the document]
mikix: Also, this is fine - like, we aren't wedding ourselves to this approach in a meaningful way - can always refactor later if we decide we don't like pandas.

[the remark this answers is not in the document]
dogversioning: Added a comment about this for posteritiy and leaving it as is.

PR #249, opened 2024-06-07
Chunking sql export queries, coverage

dogversioning:
- Exporter is rewritten to use `pyarrow` to enable large file exports
  - This required regenerating export example data
- Added coverage action to CI
- Added additional tests to clear coverage threshold
- Removed 'all' as a study target since it was increasingly not useful
- Removed remnants of old study creation mechanism

### Checklist


cumulus_library/actions/exporter.py
```diff
@@ -56,13 +70,28 @@ def export_study(
         description=f"Exporting {manifest_parser.get_study_prefix()} data...",
     ):
         query = f"SELECT * FROM {table}"
-        dataframe = db.execute_as_pandas(query)
+        # Note: we assume that, for duckdb, you are unlikely to be dealing with large
+        # exports, so it will ignore the chunksize parameter, as it does not provide
+        # a pandas enabled cursor.
+        dataframe_chunks, db_schema = db.execute_as_pandas(query, chunksize=chunksize)
```
mikix, line 76: nit: we talked about this verbally, but just writing down so it's somewhere concrete: it might be nice to not mix dataframes if we can - like always be talking pyarrow, or move the choice of dataframe to the database backend, I dunno. Not *super* important, since we are forcing a schema, but I could imagine a bad type conversion losing data (like float -> int or whatever)

cumulus_library/actions/exporter.py
```diff
@@ -56,13 +70,28 @@ def export_study(
...
+        first_chunk = next(dataframe_chunks)
```
mikix, line 77: What's this line about? Trying to raise an iteration error? (though that would break out of this method, since it's not caught)

cumulus_library/databases.py
```diff
@@ -652,23 +714,25 @@ def read_ndjson_dir(path: str) -> dict[str, pyarrow.Table]:
 
 def create_db_backend(args: dict[str, str]) -> DatabaseBackend:
     db_config.db_type = args["db_type"]
-    database = args["schema_name"]
+    schema = args["schema_name"]
     load_ndjson_dir = args.get("load_ndjson_dir")
 
     if db_config.db_type == "duckdb":
-        backend = DuckDatabaseBackend(database)  # `database` is path name in this case
+        backend = DuckDatabaseBackend(schema)  # `database` is path name in this case
```
mikix, line 721: this comment should now read "`schema` is path name in this case"

cumulus_library/databases.py
```diff
@@ -525,18 +558,47 @@ def pandas_cursor(self) -> duckdb.DuckDBPyConnection:
...
+                        f"{column[0],column[1]} does not have a conversion type"
+                    )
+        return output
 
     def parser(self) -> DatabaseParser:
         return DuckDbParser()
 
     def operational_errors(self) -> tuple[Exception]:
-        return (duckdb.OperationalError,)
+        return (duckdb.OperationalError,)  # pragma: no cover
```
mikix, line 601: These methods _could_ be tested - and probably should be?

cumulus_library/actions/exporter.py
```diff
@@ -56,13 +70,28 @@ def export_study(
         description=f"Exporting {manifest_parser.get_study_prefix()} data...",
     ):
         query = f"SELECT * FROM {table}"
-        dataframe = db.execute_as_pandas(query)
+        # Note: we assume that, for duckdb, you are unlikely to be dealing with large
+        # exports, so it will ignore the chunksize parameter, as it does not provide
+        # a pandas enabled cursor.
```
mikix, line 75: This feels like a comment that might more naturally fit in the duckdb database implementation of `execute_as_pandas`.

cumulus_library/databases.py
```diff
@@ -151,7 +152,9 @@ def pandas_cursor(self) -> DatabaseCursor:
         """
 
     @abc.abstractmethod
-    def execute_as_pandas(self, sql: str) -> pandas.DataFrame:
+    def execute_as_pandas(
+        self, sql: str, chunksize: int | None = None
+    ) -> (pandas.DataFrame | collections.abc.Iterator[pandas.DataFrame], list[tuple]):
```
mikix, line 157: oof :smile:

cumulus_library/databases.py
```diff
@@ -337,7 +370,7 @@ def upload_file(
         return f"s3://{bucket}/{s3_key}"
 
     def close(self) -> None:
-        return self.connection.close()
+        return self.connection.close()  # pragma: no cover
```
mikix, line 373: We never close out the connection in tests?

main
mikix, line 1: :scissors: ?

test
mikix, line 1: :scissors:?

mikix (approved):

.github/workflows/ci.yaml
```diff
@@ -29,9 +29,23 @@ jobs:
...
-          python -m pytest
+          python -m pytest --cov-report xml --cov=cumulus_library tests
+      - name: Generate coverage report
+        uses: orgoro/coverage@v3.1
+        with:
+            coverageFile: coverage.xml
+            token: ${{ secrets.GITHUB_TOKEN }}
+            thresholdAll: .9
+            thresholdNew: 1
+            thresholdModified: .95
```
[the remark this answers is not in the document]
dogversioning: yeah that is my rough roadmap for this - trying to spread this out rather than just spending a week on tests.

main
dogversioning: where did these even come from

cumulus_library/actions/exporter.py
dogversioning: i think it's in both places - but i could remove it from here.

cumulus_library/actions/exporter.py
dogversioning: you know - this is holdover from when i was trying to do this with dataframes. it can probably be cut.

cumulus_library/actions/exporter.py
dogversioning: #250

cumulus_library/databases.py
dogversioning: I don't think we have anyplace we're explicitly closing a cursor at the moment, actually.

cumulus_library/databases.py
dogversioning: i can look into this in the future

cumulus_library/databases.py
dogversioning: it does feel bad

Issue #225, opened 2024-04-26
Move commonly used params to a config class


dogversioning:
There are certain values that are used over and over in lots of places:
- schema
- cursor
- parser

We should move these things into a single class, and then pass that around, which should make adding additional widely shared bits of config (like database engine type, which should be moved into this after its created) easier to add in the future if there is a need.

dogversioning:
this is started by #233

PR #252, opened 2024-06-18
Allow a study to specify a schema

dogversioning:
- Allows a study to specify an explict schema
  - Auto transforms table names if this is done - so 'study__table' becomes just 'table', so reverting this is a config-only change
- adds tests for --continue and --builder CLI args, fixing one bug with the former.

### Checklist
- [] Update template repo if there are changes to study configuration

### DIFF
--- a/cumulus_library/base_table_builder.py
+++ b/cumulus_library/base_table_builder.py
@@ -86,11 +86,7 @@ class BaseTableBuilder(ABC):
             )
             for query in self.queries:
                 try:
-                    if manifest and manifest.get_dedicated_schema():
-                        query = query.replace(
-                            f"`{manifest.get_study_prefix()}__",
-                            "`",
-                        )
+                    query = base_utils.update_query_if_schema_specified(query, manifest)
                     with base_utils.query_console_output(
                         verbose, query, progress, task
                     ):
--- a/cumulus_library/base_table_builder.py
+++ b/cumulus_library/base_table_builder.py
@@ -6,7 +6,7 @@ import sys
 from abc import ABC, abstractmethod
 from typing import final
 
-from cumulus_library import base_utils
+from cumulus_library import base_utils, study_parser
 from cumulus_library.databases import DatabaseCursor
 
 
@@ -42,6 +42,7 @@ class BaseTableBuilder(ABC):
         verbose: bool,
         *args,
         drop_table: bool = False,
+        manifest: study_parser.StudyManifestParser = None,
         **kwargs,
     ):
         """Executes queries set up by a prepare_queries call
@@ -51,7 +52,7 @@ class BaseTableBuilder(ABC):
         :param verbose: toggle for verbose output mode
         :param drop_table: drops any tables found in prepared_queries results
         """
-        self.prepare_queries(cursor, schema, *args, **kwargs)
+        self.prepare_queries(cursor, schema, *args, manifest=manifest, **kwargs)
         if drop_table:
             table_names = []
             for query in self.queries:
@@ -69,6 +70,7 @@ class BaseTableBuilder(ABC):
                         )
 
                     table_name = table_name[0]
+                    # TODO: this may not be required? reinvestigate
                     # if it contains a schema, remove it (usually it won't, but some
                     # CTAS forms may)
                     if "." in table_name:
@@ -84,6 +86,12 @@ class BaseTableBuilder(ABC):
             )
             for query in self.queries:
                 try:
+                    if manifest:
+                        if manifest.get_dedicated_schema():
+                            query = query.replace(
+                                f"`{manifest.get_study_prefix()}__",
+                                "`",
+                            )
                     with base_utils.query_console_output(
                         verbose, query, progress, task
                     ):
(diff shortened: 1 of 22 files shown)


cumulus_library/base_table_builder.py
```diff
@@ -69,6 +70,7 @@ def execute_queries(
                         )
 
                     table_name = table_name[0]
+                    # TODO: this may not be required? reinvestigate
```
dogversioning, line 73: This PR is big enough that I'm deferring this to later - I'll create an issue for it

cumulus_library/cli.py
```diff
@@ -90,21 +108,23 @@ def clean_study(
...
+                schema = self.get_schema(manifest)
                 cleaner.clean_study(
-                    manifest_parser=manifest_parser,
+                    manifest_parser=manifest,
                     cursor=self.cursor,
-                    schema_name=self.schema_name,
+                    schema_name=schema,
                     verbose=self.verbose,
                     stats_clean=stats_clean,
                     prefix=target,
```
dogversioning, line 119: This happened a bunch in here - I'm not moving things around because i'm trying to not make a breaking change, but this was a headache and i think this might be the last major feature before i make breaking changes.

tests/test_data/study_dedicated_schema/manifest.toml
```diff
@@ -0,0 +1,13 @@
...
+file_names = ["module1.py", "module1.py", "module2.py"]
+
+[sql_config]
+file_names = ["test.sql"]
+
+[export_config]
+export_list = ["study_dedicated_schema__table_raw_sql"]
+
+[options]
+use_dedicated_schema="dedicated"
```
mikix, line 13: nit: I might like to see a beefier test here - like actually checking that the resulting stuff gets placed in a dedicated schema and the sql is replaced etc. This test currently smoketests that we don't blow up when this option is set, but not that it actually changes anything, yeah?

cumulus_library/protected_table_builder.py
```diff
@@ -37,21 +37,30 @@ def prepare_queries(
         study_name: str,
         study_stats: dict,
         *args,
+        manifest: study_parser.StudyManifestParser | None = None,
         **kwargs,
     ):
+        if manifest and manifest.get_dedicated_schema():
+            db_schema = manifest.get_dedicated_schema()
```
mikix, line 44: Why isn't the `get_schema()` check in cli.py not saving you from having to check this flag downstream in order to use the right schema?

I get why you still need some logic for the table names -- though... if we saved somewhere a field like `table_namespace = f"{study_prefix}__"` and set that to `""` when in a dedicated schema study, we could reduce logic checks around the place - which feel a little brittle - easy to forget a place where we need custom logic for this case. Just feels like this feature *could* be, if we were set up for it, a single block of like:
```
if dedicated:
  create_db()
  schema = dedicated
  short_prefix = ""
else:
  schema = cli_schema
  short_prefix = f"{study_prefix}__"
full_prefix = f"{schema}.{short_prefix}"
```
or something.

mikix (approved):

cumulus_library/study_parser.py
```diff
@@ -61,6 +61,16 @@ def get_study_prefix(self) -> str | None:
...
 
+    def get_dedicated_schema(self) -> str | None:
+        """Reads the contents of the dedicated schema in the options dict
+
+        :returns: A dictionay of objects, or None if not found
+        """
+        if options := self._study_config.get("options"):
+            if schema := options.get("use_dedicated_schema"):
+                return schema
+        return None
```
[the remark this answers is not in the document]
dogversioning: image

cumulus_library/base_table_builder.py
```diff
@@ -84,6 +86,12 @@ def execute_queries(
             )
             for query in self.queries:
                 try:
+                    if manifest:
+                        if manifest.get_dedicated_schema():
+                            query = query.replace(
+                                f"`{manifest.get_study_prefix()}__",
+                                "`",
+                            )
```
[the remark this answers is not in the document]
dogversioning: yeah i had that thought breifly - i'll find a home for it.

cumulus_library/protected_table_builder.py
dogversioning: this is some of what drove the 'this is the last breaking change' comment. If the schema moved into the config, this all becomes much simpler.

my instinct is to sort of live with this until biting that off - noting this for #225

docs/creating-studies.md
```diff
@@ -92,12 +100,31 @@ export_list = [
...
+
 # [statistics_config]
 # file_names = 
 # [
 #    "psm_config.toml"
 # ]
 
+# The following section is for advanced/unusual study use cases
+
+# [options]
```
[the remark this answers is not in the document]
dogversioning: so i think 'internal' is a bad move - we might want to have something slightly more general in the future? i could live with 'advance' or 'advanced_options'

tests/test_data/study_dedicated_schema/manifest.toml
dogversioning: yeah, fair point

PR #255, opened 2024-06-25
Added generic logging interface

dogversioning:
This PR moves logging infrastructure from an ad hoc 'as you need it' kind of interface to something in a dedicated module.

### Checklist


cumulus_library/actions/file_generator.py
```diff
@@ -85,7 +85,7 @@ def run_generate_markdown(
         if len(count_tables) > 0:
             f.write(f"## {manifest_parser.get_study_prefix()} count tables\n\n")
             for table in count_tables:
-                manifest_parser._write_md_table(table, study_df, f)
+                _write_md_table(table, study_df, f)
```
dogversioning, line 88: this is an unrelated bug in the markdown generator I found this morning

cumulus_library/protected_table_builder.py
```diff
@@ -1,10 +1,10 @@
 """Builder for creating tables for tracking state/logging changes"""
 
-from cumulus_library import base_table_builder, enums, study_parser
+from cumulus_library import base_table_builder, databases, enums, study_parser
 from cumulus_library.template_sql import base_templates
 
-TRANSACTIONS_COLS = ["study_name", "library_version", "status", "event_time"]
-TRANSACTION_COLS_TYPES = ["varchar", "varchar", "varchar", "timestamp"]
+TRANSACTIONS_COLS = ["study_name", "library_version", "status", "event_time", "message"]
+TRANSACTION_COLS_TYPES = ["varchar", "varchar", "varchar", "timestamp", "varchar"]
```
dogversioning, line 7: i talked about adding a gentle upgrade path for this, but if we're making this part of a 3.0 breaking changeset, i'm inclined to say something like 'just remove previous lib_transaction tables'. Can be talked out of this position.

cumulus_library/template_sql/insert_into.sql.jinja
```diff
@@ -12,6 +12,8 @@ VALUES
     {%- for field in row -%}
     {%- if table_cols[loop.index0] in type_casts.keys() -%}
     {{ type_casts[table_cols[loop.index0]] }} '{{ field }}'
+    {%- elif not field -%}
+    NULL
```
dogversioning, line 16: not actually a bug per se, just the first time we actually tried to explicitly insert null data via this template.

mikix: I've hit errors with Athena where it likes that kind of thing as `CAST(NULL AS VARCHAR)` - are you gonna need that here too?

cumulus_library/log_utils.py
```diff
@@ -0,0 +1,95 @@
...
+        cursor=cursor,
+        schema=schema,
+        manifest=manifest,
+        dataset=[
+            [
+                manifest.get_study_prefix(),
+                __version__,
+                status.value,
+                base_utils.get_utc_datetime(),
+                message if message else None,
```
mikix, line 41: nit: I think `message or None` is maybe slightly more readable, but if/else is good too.

mikix (approved):

cumulus_library/template_sql/insert_into.sql.jinja
dogversioning: almost certaintly, yeah

Issue #246, opened 2024-05-31
Rename StudyManifestParser


dogversioning:
As mentioned in #245, parsing is an implementation detail more than a core facet of this class. So this proposes two changes:
- Make the class name StudyManifest
- Rename the file to `study_manifest.py`

dogversioning:
completed via #256

PR #256, opened 2024-07-01
Actions/builders converted to running via config & manifest

dogversioning:
- Renames the StudyManifestParser to StudyParser #246 
- Moves cursor/schema to config object #225
- Changes most actions & builders to take config + manifest as the first two arguments
  - Eliminates a lot of commonly passed args (verbose etc etc) in favor of the study config
### Checklist


cumulus_library/actions/builder.py
```diff
@@ -199,72 +164,70 @@ def run_counts_builders(
...
         with open(toml_path, encoding="UTF-8") as file:
-            config = toml.load(file)
-            config_type = config["config_type"]
-            target_table = config["target_table"]
+            stats_config = toml.load(file)
+            config_type = stats_config["config_type"]
+            target_table = stats_config["target_table"]
+
+        if (target_table,) in existing_stats and not config.stats_build:
+            print("early break")
```
mikix, line 211: :scissors: here and a couple lines below

cumulus_library/cli.py
```diff
@@ -340,19 +282,31 @@ def run_cli(args: dict):
...
     # all other actions require connecting to the database
     else:
+        # if we're targeting a builder explicitly, we always want to
+        # run in drop mode
+        if builder := args.get("builder"):
+            drop_table = builder
+        else:
+            drop_table = args.get("drop_table")
+        if args.get("db_type") == "duckdb":
+            schema = "main"
```
mikix, line 292: This feels like a little bit of duckdb specific logic escaping from database.py, but it's not upsetting.

cumulus_library/databases.py
```diff
@@ -615,7 +615,11 @@ def operational_errors(self) -> tuple[Exception]:
 
     def create_schema(self, schema_name):
         """Creates a new schema object inside the database"""
-        self.connection.sql(f"CREATE SCHEMA {schema_name}")
+        schemas = self.connection.sql(
+            "SELECT schema_name FROM information_schema.schemata"
+        ).fetchall()
+        if (schema_name,) not in schemas:
+            self.connection.sql(f"CREATE SCHEMA {schema_name}")
```
mikix, line 622: Rude for SQL to not have `CREATE SCHEMA IF NOT EXISTS`

cumulus_library/protected_table_builder.py
```diff
@@ -32,22 +37,28 @@ class ProtectedTableBuilder(base_table_builder.BaseTableBuilder):
...
-        study_stats: dict,
+        config: base_utils.StudyConfig,
+        manifest: study_manifest.StudyManifest | None = None,
         *args,
-        manifest: study_parser.StudyManifestParser | None = None,
+        study_stats: dict | None = None,
         **kwargs,
     ):
+        if study_stats is None:
+            study_stats = {}
```
mikix, line 47: extremely minor nit: this could be a one-liner with `study_stats = study_stats or {}`

mikix (approved):
I did not do a very close read of all the renaming stuff - but looks good!

cumulus_library/cli.py
dogversioning: oh thank you, i was like 'i need to mention something' and couldn't remember what it was.

If we want to get this cat back in the bag, we'll need to undo the overloading we're currently doing, where raw schema name and duckdb file path are going into the same location. So we'd basically add one new CLI arg and then split how that logic is handled in some places.

if we want to do this, i can create a ticket for it, but i don't want to do it on this PR given the density.

mikix: Hmm, sounds like you are talking a bigger change than I was - I'm not 100% on what your vision there is.

I was talking more like the DuckDB database backend simply ignoring the input schema arg and using `main` or a method like `DatabaseBackend.decide_schema_name(args)` or something - just anything to put this tiny business logic under the umbrella of the DuckDB-specific code. Ideally in my mind, nothing outside of `databases.py` every has to know DuckDB exists.

dogversioning: Per offline discussion, leaving this as is for now, but logging a ticket: #258

PR #260, opened 2024-07-08
Schema handling for external tables

dogversioning:
- Updates 3.0 schema handling for external tables in athena
- Adds `rxnorm` as a target
- cleanup: one irritating print statement, module-based import

### Checklist

### DIFF
--- a/cumulus_library/base_utils.py
+++ b/cumulus_library/base_utils.py
@@ -134,10 +134,19 @@ def update_query_if_schema_specified(
     query: str, manifest: study_manifest.StudyManifest
 ):
     if manifest and manifest.get_dedicated_schema():
-        query = query.replace(
-            f"{manifest.get_study_prefix()}__",
-            f"{manifest.get_dedicated_schema()}.",
-        )
+        # External queries in athena require a schema to be specified, so we
+        # just trim the prefix here.
+        # TODO: Move this to be a database defined function.
+        if "CREATE EXTERNAL" in query:
+            query = query.replace(
+                f"{manifest.get_study_prefix()}__",
+                "",
+            )
+        else:
+            query = query.replace(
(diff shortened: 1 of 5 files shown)


cumulus_library/base_utils.py
```diff
@@ -134,10 +134,19 @@ def update_query_if_schema_specified(
     query: str, manifest: study_manifest.StudyManifest
 ):
     if manifest and manifest.get_dedicated_schema():
-        query = query.replace(
-            f"{manifest.get_study_prefix()}__",
-            f"{manifest.get_dedicated_schema()}.",
-        )
+        # External queries in athena require a schema to be specified, so we
+        # just trim the prefix here.
+        # TODO: Move this to be a database defined function.
```
mikix, line 139: require a schema to NOT be specified? Looks like you're not adding the schema in the external flow. Or does this comment mean that the schema is ALREADY specified elsewhere, and shouldn't be repeated here?

mikix (approved):

cumulus_library/module_allowlist.json
```diff
@@ -4,6 +4,8 @@
     "allowlist" : {
         "covid_symptom" : "cumulus_library_covid/covid_symptom",
         "data_metrics" : "cumulus_library_data_metrics",
+        "rxnorm" : "cumulus_library_umls/rxnorm",
```
mikix, line 7: There's an opportunity here to drop the repeated subdir (and switch the pip package here to rxnorm)

dogversioning: oh yes i spaced on this one.

cumulus_library/base_utils.py
dogversioning: the latter - lemme clarify

Issue #259, opened 2024-07-08
gracefully handle lib_transactions update


dogversioning:
rather than just failing hard, either catch the error and present it to the user, or migrate the table to the new schema format

PR #262, opened 2024-07-10
Gracefully migrate transactions

dogversioning:
This PR addresses #259 by altering a given lib_transaction table if an error is raised.

Also added manifest/config to the external API.

### Checklist

### DIFF
--- a/cumulus_library/log_utils.py
+++ b/cumulus_library/log_utils.py
@@ -3,6 +3,7 @@
 from cumulus_library import (
     __version__,
     base_utils,
+    databases,
     enums,
     errors,
     study_manifest,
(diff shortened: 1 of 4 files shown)


tests/test_logging.py
```diff
@@ -88,6 +91,80 @@ def test_transactions(mock_db, schema, study, status, message, expects, raises):
...
+        assert len(cols) == 5
+        assert ("message",) in cols
+
+
+@mock.patch.dict(
+    os.environ,
+    clear=True,
+)
+@mock.patch("pyathena.connect")
+def test_migrate_transactions_athena(mock_pyathena):
```
dogversioning, line 128: this test is extremely silly.

cumulus_library/log_utils.py
```diff
@@ -86,4 +87,37 @@ def _log_table(
         dataset=dataset,
         type_casts=table.type_casts,
     )
-    config.db.cursor().execute(query)
+    cursor = config.db.cursor()
+    try:
+        cursor.execute(query)
+    except Exception as e:
+        # Migrating logging tables
+        if "lib_transactions" in table_name:
```
mikix, line 95: There's not a tighter exception to grab because of two databases, I assume? I don't suppose this one happens to already fall under the operational_errors() set?

But also, stepping back - this is a pretty awkward code path :smile: -- one that you hadn't even planned to write until I hit this and was grumbling about the UX of it -- if you think this is too gross but you were writing this anyway for my sake, I'd be fine with something as basic as "detect this case, print a nice error message about what to do next" if you didn't want to carry this kind of alter logic going forward.

dogversioning: i didn't think of operational_errors - and it might be worth keeping this around if only to actually have something that is using them, since right now they're orphans - the duckdb one is `duckdb.duckdb.BinderException`, which is pretty broad, but we could add it to this list?

anyway, it's here now, and realistically it's probably going to happen again, so we might as well just ride with 'ok this is where DB migrations go' -  this way will preserve state at least, which is different than just dropping, so it is probably 'correct' even if the edge cases are marginal.

mikix: > since right now they're orphans

sql_utils.py uses `operational_errors()` I think.

> anyway, it's here now

Fair - just we might want to give ourselves the grace to chuck it down the line as "deprecated migration stuff" if we ever add more database backends etc.

dogversioning: > sql_utils.py uses operational_errors() I think.

so it does - i don't think that's covered under test yet, which is why that's all #pragma'd up in databases.

Issue #231, opened 2024-05-03
Tech Debt: On upgrade to 3.11, use stdlib toml library


dogversioning:
https://peps.python.org/pep-0680/

PR #263, opened 2024-07-11
Uprev to 3.11

dogversioning:
- Minimum python is now 3.11 #232
- now uses the stdlib `tomllib` module for parsing #231 

### Checklist


dogversioning:
> Doesn't look like much else interesting (for us) happened in 3.11 ☹️

Noting this since it caught me off guard - the stdlib implementation is read only for the time being.

PR #264, opened 2024-07-11
Switch _code_system to _system

dogversioning:
Screenshot 2024-07-11 at 1 02 20 PM

### Checklist


cumulus_library/studies/core/core_templates/condition.sql.jinja
```diff
@@ -61,7 +61,7 @@ WITH temp_condition AS (
 SELECT
     tc.id,
     cdc.code AS category_code,
-    cdc.code_system AS category_code_system,
+    cdc.code_system AS category_system,
```
mikix, line 64: This approach is fine, but I worry that we're setting ourselves up for accidentally using `code_system` again down the line by not changing the denorm code now to give `.system` instead of `.code_system`.

cumulus_library/studies/core/reference_sql/builder_prereq_tables.sql
```diff
@@ -139,7 +139,7 @@ FROM
 -- This includes both the expected URI, as well as several found URIs from
 -- various source systems
 
-CREATE TABLE core__fhir_mapping_code_system_uri AS SELECT * FROM
+CREATE TABLE core__fhir_mapping_system_uri AS SELECT * FROM
```
mikix, line 142: stale reference build?

mikix (approved):

Issue #251, opened 2024-06-18
Seperate databases.py into per-database files


dogversioning:
Databases.py is starting to creep up on 800 lines of code - it would be a good idea to move it into its own folder and have one file per db implementation.

PR #265, opened 2024-07-11
Split out databases into seperate files

dogversioning:
This PR addresses #251 by moving the database wrapper code to a subfolder, and splits it out into one file per db to help manage cognitive complexity.

### Checklist


cumulus_library/databases/__init__.py
```diff
@@ -0,0 +1,16 @@
+from .athena import AthenaDatabaseBackend, AthenaParser
+from .base import DatabaseBackend, DatabaseCursor, DatabaseParser
+from .duckdb import DuckDatabaseBackend, DuckDbParser
+from .utils import create_db_backend, read_ndjson_dir
+
+__all__ = [
```
mikix, line 6: I would recommend not including __all__, especially in a case like this where you don't have other code - it's really only useful if you want to restrict the symbols being exported. But here, all the code is elsewhere and you're just allow-listing-by-importing already.

mikix (approved):

cumulus_library/databases/__init__.py
dogversioning: as discussed off thread, added an ignore for this.

PR #266, opened 2024-07-16
Fix Athena's ability to export "decimal" types

mikix:
Also, fix a small NULL handling issue with array_join in duckdb.

### Checklist

### DIFF
--- a/cumulus_library/databases/athena.py
+++ b/cumulus_library/databases/athena.py
@@ -105,6 +105,8 @@ class AthenaDatabaseBackend(base.DatabaseBackend):
                     output.append((column[0], pyarrow.int64()))
                 case "double":
                     output.append((column[0], pyarrow.float64()))
+                case "decimal":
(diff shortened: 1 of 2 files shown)


dogversioning (approved):

cumulus_library/databases/athena.py
```diff
@@ -105,6 +105,8 @@ def col_pyarrow_types_from_sql(self, columns: list[tuple]) -> list:
                     output.append((column[0], pyarrow.int64()))
                 case "double":
                     output.append((column[0], pyarrow.float64()))
+                case "decimal":
+                    output.append((column[0], pyarrow.decimal128(column[4], column[5])))
```
mikix, line 109: I forgot to mention, but duckdb does not need a similar change - the type comes through as NUMBER there, which is already handled.

dogversioning: i want a library to just handle 'python types to sql types for this database'

PR #268, opened 2024-07-23
Template/manifest updates for vocab

dogversioning:
- Adds two new generic templates (view/union generation)
- Adds a convenience method for getting the appropriate prefix string for a table.

### Checklist

### DIFF
--- a/cumulus_library/base_utils.py
+++ b/cumulus_library/base_utils.py
@@ -2,7 +2,6 @@
 
 import dataclasses
 import datetime
-import json
 import pathlib
 import shutil
 import zipfile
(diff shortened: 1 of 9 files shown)


tests/test_base_templates.py
```diff
@@ -245,6 +247,81 @@ def test_get_column_datatype_query():
     assert query == expected
 
 
+def test_get_create_table_from_union():
+    expected = """CREATE TABLE union_table AS
+SELECT
+foo,
+bar,
+baz
```
dogversioning, line 255: these :should: be indented but SQLFluff was insistent; i decided not to fight it.

cumulus_library/base_utils.py
```diff
@@ -54,6 +54,12 @@ def get_schema(config: StudyConfig, manifest: study_manifest.StudyManifest):
     return config.schema
 
 
+def get_prefix_with_seperator(manifest: study_manifest.StudyManifest):
+    if dedicated := manifest.get_dedicated_schema():
+        return f"{dedicated}."
+    return f"{manifest.get_study_prefix()}__"
```
mikix, line 60: Should this just be a method on StudyManifest?

mikix (approved):

cumulus_library/base_utils.py
dogversioning: i suppose - i started here because of the get_schema method, but without the config dependency that probably makes sense.

PR #270, opened 2024-07-26
Updated cleaner to remove views

dogversioning:
This fixes a bug where views would not be picked up during the clean action.

It also pulls the trigger on forcing modified files to have full coverage.

### Checklist

### DIFF
--- a/cumulus_library/actions/cleaner.py
+++ b/cumulus_library/actions/cleaner.py
@@ -115,7 +115,7 @@ def clean_study(
     table_sql = base_templates.get_show_tables(config.schema, drop_prefix)
     view_table_list = []
     for query_and_type in [[view_sql, "VIEW"], [table_sql, "TABLE"]]:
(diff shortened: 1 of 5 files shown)


cumulus_library/actions/cleaner.py
```diff
@@ -115,8 +115,9 @@ def clean_study(
     cursor = config.db.cursor()
     view_sql = base_templates.get_show_views(config.schema, drop_prefix)
     table_sql = base_templates.get_show_tables(config.schema, drop_prefix)
+    view_table_list = []
     for query_and_type in [[view_sql, "VIEW"], [table_sql, "TABLE"]]:
-        view_table_list = get_unprotected_stats_view_table(
+        view_table_list = view_table_list + get_unprotected_stats_view_table(
```
mikix, line 120: nit: +=

mikix (approved):

PR #273, opened 2024-07-31
Mask buckets in generated sql

dogversioning:
This PR strips bucket/db information from generated sql to mask some not-quite-secret data.

### Checklist


cumulus_library/actions/builder.py
```diff
@@ -94,7 +95,15 @@ def _load_and_execute_builder(
     table_builder_class = table_builder_subclasses[0]
     table_builder = table_builder_class()
     if write_reference_sql:
```
mikix, line 97: nit: this code path is starting to feel like it should call out to a separate function, as it grows logic like this. Not necessary now, just saying.

mikix (approved):

cumulus_library/actions/builder.py
dogversioning: yeah hopefuly this is the last time we touch it like this.

PR #276, opened 2024-08-01
Replace core__medicationrequest code with core__medication code

### LINKED ISSUE #227: Core Medication tables could use a reorganization
**Current State**
- `core__medicationrequest` only contains "inline" codes that are RXNORM. (But adds some niceties like dosage instruction text)
- `core__medication` is built from MedicationRequest tables but includes "inline", "contained", and "external" codes. Does not filter by system.
- I believe we build our `core__count_medicationrequest` table off `core__medicationrequest` and are thus missing a lot of data.

**My Proposal**

- Nuke `core__medicationrequest`
- Rename `core__medication` to be `core__medicationrequest`
- Add in any now-missing features (like `dosageInstruction.text`) to the new tables.

I imagine the `core__medication` tables were named that because they sometimes inspect Medication resources (contained or external). But the rows do not represent Medication resources at all. They are summaries of MedicationRequests that happened to (behind the scenes) grab the correct codes from Medication resources.

Consumers of the tables don't care about all that abstraction work - let's actually provide the abstraction.

In this proposal, we don't need a `core__medication` table, since those aren't patient/encounter linked resources (maybe we could still provide one in the future, if someone wants the ability to look up detailed med info -- but not our concern right now).

**Some other idea**
Or some better idea - but regardless of where we land, I think there's currently some unnecessary duplication between those two tables, and we could stand to develop an actual vision for them.

mikix:
- core__medication no longer exists.
- core__medicationrequest now has all the "abstract where medications come from" logic that core__medication had and no longer retricts itself to just RxNorm.

Changes to the field list for core__medicationrequest:
- Dropped `dosageInstruction_text`, since no one was using it and the  ETL strips the text anyway.
- Add `reported_ref` (to match `reportedBoolean`).
- Add `category_display`.

Fixes #227

### Checklist


cumulus_library/studies/core/core_templates/medicationrequest.sql.jinja
```diff
@@ -1,89 +1,123 @@
...
+            mr.authoredOn_month
+        FROM mr_basics AS mr
+        LEFT JOIN core__medicationrequest_dn_category AS mrc ON mr.id = mrc.id
+    )
+
+    {# Internal: medication data from inline ETL extraction.
+    95% of the time, this dataset is the 'correct' dataset for Cerner.
+    It may not be present in EPIC datasets. #}
+    SELECT
+        mr.*,
```
dogversioning, line 90: Did you run sqlfluff on this? generally star selects are flagged by it.

dogversioning (approved):

cumulus_library/studies/core/core_templates/medicationrequest.sql.jinja
dogversioning: seems like it passed ci... i'm honestly very surprised. is it because of the interstitial table you're creating is inside this query? if so that's very good to know.

PR #278, opened 2024-08-09
Added or/ilike macros to syntax

dogversioning:
This adds a few convenience macros to syntax.jinja.
### Checklist

### DIFF
--- a/cumulus_library/actions/cleaner.py
+++ b/cumulus_library/actions/cleaner.py
@@ -30,7 +30,7 @@ def _execute_drop_queries(
             cursor.execute(drop_view_table)
 
 
-def get_unprotected_stats_view_table(
+def _get_unprotected_stats_view_table(
     config: base_utils.StudyConfig,
     query: str,
     artifact_type: str,
@@ -114,7 +114,7 @@ def clean_study(
     table_sql = base_templates.get_show_tables(config.schema, drop_prefix)
     view_table_list = []
     for query_and_type in [[view_sql, "VIEW"], [table_sql, "TABLE"]]:
-        view_table_list += get_unprotected_stats_view_table(
+        view_table_list += _get_unprotected_stats_view_table(
             config,
             query_and_type[0],
             query_and_type[1],
--- a/cumulus_library/actions/cleaner.py
+++ b/cumulus_library/actions/cleaner.py
@@ -31,7 +31,7 @@ def _execute_drop_queries(
 
 
 def get_unprotected_stats_view_table(
-    cursor: databases.DatabaseCursor,
+    config: base_utils.StudyConfig,
     query: str,
     artifact_type: str,
     drop_prefix: str,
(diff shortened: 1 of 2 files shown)


cumulus_library/actions/cleaner.py
```diff
@@ -31,7 +31,7 @@ def _execute_drop_queries(
 
 
 def get_unprotected_stats_view_table(
-    cursor: databases.DatabaseCursor,
+    config: base_utils.StudyConfig,
```
dogversioning, line 34: this is unrelated - i am midstream in tinkering with this to get specific tables, and need access to the config so i can get at the schema object. But it's functionally equivalent, so I opted to not back it out as part of this PR.

cumulus_library/actions/cleaner.py
```diff
@@ -31,7 +31,7 @@ def _execute_drop_queries(
 
 
 def get_unprotected_stats_view_table(
```
mikix, line 33: This method isn't used anywhere, right? Is it public API?

I didn't think the Library had any public utility methods like this.

cumulus_library/template_sql/shared_macros/syntax.sql.jinja
```diff
@@ -57,7 +60,11 @@ worry about the slightly more performant regex based matching we want to use
 in most cases.
 -#}
 {%- macro like(field, match_string) -%}
-REGEXP_LIKE({{ field }}, {{ wildcard_to_regex(match_string) }})
+REGEXP_LIKE({{ field }}, '^{{ wildcard_to_regex(match_string) }}$')
+{%- endmacro -%}
+
+{%- macro ilike(field, match_string) -%}
+REGEXP_LIKE({{ field }}, '.*(?i:{{ wildcard_to_regex(match_string) }}).*$')
```
mikix, line 67: I don't like that `ilike` does TWO things different from `like` - the case insensitive thing makes sense and is obvious. It would _not_ be obvious to me that it also changes how the regex is anchored. Especially since the `ILIKE` SQL keyword doesn't work like that (at least in duckdb's implementation).

If that piece of this is important, can we instead add a new set of methods? `search` and `isearch` maybe? Whatever names make sense.

mikix (approved):

cumulus_library/actions/cleaner.py
dogversioning: that should be private, yeah

mikix: OK I was just trying to figure out what this method was about - searching for it didn't yield any results. But I had read the search results wrong - I thought it wasn't used at all. Nevermind me.

cumulus_library/template_sql/shared_macros/syntax.sql.jinja
dogversioning: Yeah, good call - I did that as a convenience for the vocab piece but it's probably not good behavior.

Issue #211, opened 2024-04-18
Consider renaming all the `_code_system` aliases to just `_system`


mikix:
When we did the big field rename for 2.0 that aligned us closer to the FHIR spec, we used the following scheme for codeable fields:

```
    interpretation_code,
    interpretation_code_system,
    interpretation_display,
```

But I personally find that a little confusing in a world where we are mostly mapping the FHIR spec. Why not use just `system`, which is what the field is called. `code_system` makes _my_ brain think we're talking about `code.system` instead of `system`.

So my pitch is that for 3.0, we consider a mass rename back like `s/code_system/system/g` as one of its breaking changes.

dogversioning:
For postertity, documenting our off-thread discussion:
- We are going to continue using `code_system` for new tables in 2.0, and then make the change in 3.0
- For studies we control, we should consistently use the pattern `baseResource_code_system` when joining into tables so it will be easier to bulk replace going forward.

PR #281, opened 2024-08-19
code_system replacement round two

dogversioning:
This replaces all remaining instances of code_system in resource tables missed in the first pass of #211, and also from the denormalized coding/codeable_concept tables 

### Checklist


cumulus_library/template_sql/sql_utils.py
```diff
@@ -258,19 +258,17 @@ def denormalize_complex_objects(
 
 def validate_schema(
     database: databases.DatabaseBackend,
-    expected_table_cols: dict,
+    expected_table_cols: dict[dict],
```
mikix, line 261: I'm used to dict[key, value] - what's this typing mean?

mikix (approved):

cumulus_library/template_sql/sql_utils.py
dogversioning: nothing - will update to `dict[key,dict[key,value]]`

PR #282, opened 2024-08-20
Add BaseTableBuilder and CountsBuilder to the public API

mikix:
### Checklist


dogversioning (approved):
oh what a good idea

dogversioning:
PsmBuilder should :probably: also be part of the API, but we can leave that for later

PR #284, opened 2024-08-21
Remove contents of /schemas

### LINKED ISSUE #283: Remove contents of `schema` dir
The artifacts inside schema (with the removal of counts into statistics) are now orphans, and have been for awhile.

We might need to get back to these in the future for reference for some other projects (one reasons why we're tracking this with a ticket), but to avoid confusing users for now, we're just going to remove them ahead of the next release.

dogversioning:
The remaining artifacts in the schema directory have been unused for awhile now, and may be confusing to new users. This removes them - if we need them for reference, we can always go back to any 2.x tag.

Resolves #283 

### Checklist
- [X] Consider if documentation (like in `docs/`, or if you've changed the structure of a table) needs to be updated


mikix (approved):

mikix:
Oh also, there is this section of pyproject.toml that can be removed:
```
# ./cumulus_library/schema needs general revisiting, deferring on style maintenance
# until that occurs
"cumulus_library/schema/valueset.py" = ["E501"]
```

Issue #139, opened 2023-10-23
Add core tables for AllergyIntolerance, Device, DiagnosticReport, and Immunization


mikix:
These four tables are being added to the ETL, and to make them truly useful, we probably want nice little unnested core versions of these tables.

I don't have ideas for **what** those core tables should look like, though. :shrug:

- There is Epic sandbox data in BCH's `cumulusdb-vkca9sswjppm` database
- There is Synthea data in BCH's `cumulusdb-nkxjufzpvwih` database

No Cerner tables yet - those will get filled out as I do bulk exports of that data.

PR #287, opened 2024-08-26
Add AllergyIntolerance support to core study

### LINKED ISSUE #285: Support for Allergy in CORE
FHIR support for CORE is defined here 
https://hl7.org/fhir/us/core/STU4/StructureDefinition-us-core-allergyintolerance.html

AllergyIntolerance.code is either RxNorm or SNOMED. 
Conveniently, VSAC provides this valueset: 
https://vsac.nlm.nih.gov/valueset/2.16.840.1.113762.1.4.1186.8/expansion

Valueset includes 
* 258 SNOMEDCT codes (substance ) 
* 563 RxNorm codes (ingredient) 

Categories we are most interested in are "medication" and "biologic"
https://hl7.org/fhir/R4/valueset-allergy-intolerance-category.html

Note that substance code has >2000 entries, for example, "**Codeine phosphate**"
https://hl7.org/fhir/R4/valueset-substance-code.html

"Findings" may be important to our "Symptoms" work, findings and symptoms may describe related or synonymous phenotypic observations. 
https://hl7.org/fhir/R4/valueset-clinical-findings.html

mikix:
- Adds core__allergyintolerance
- Adds core__count_allergyintolerance_month

Fixes #285
A subset of #139 

### Checklist
- [x] Consider if documentation (like in `docs/`, or if you've changed the structure of a table) needs to be updated


cumulus_library/statistics/statistics_templates/count.sql.jinja
```diff
@@ -102,19 +101,19 @@ CREATE TABLE {{ table_name }} AS (
         {%- endif -%}
 
         {#- resource specific filtering conditions -#}
-        {%- if fhir_resource == 'documentreference' and filter_resource %}
+        {%- if fhir_resource == 'documentreference' %}
```
mikix, line 104: Dropped the `filter_resource` check here because it's inside a giant `{% if filter_resource %}`

cumulus_library/template_sql/codeable_concept_denormalize.sql.jinja
```diff
@@ -120,13 +120,12 @@ CREATE TABLE {{ target_table }} AS (
             ROW_NUMBER()
                 OVER (
                     PARTITION BY id
-                    ORDER BY priority ASC
+                    ORDER BY priority ASC, code ASC
```
mikix, line 123: For better predictability in tests (shouldn't normally happen that you have multiple codes for the same system)

cumulus_library/template_sql/codeable_concept_denormalize.sql.jinja
```diff
@@ -120,13 +120,12 @@ CREATE TABLE {{ target_table }} AS (
...
                 ) AS available_priority
         FROM union_table
         GROUP BY
             id, row, priority, system, code, display, userSelected
             {%- for extra_field in extra_fields %}, {{ extra_field[1] }}{% endfor %}
-        ORDER BY priority ASC
```
mikix, line 129: AFAICT this is not doing anything here - the next bit just pulls a single line from this temp table, based on `available_priority`, so we don't need to sort this.

cumulus_library/studies/core/core_templates/allergyintolerance.sql.jinja
```diff
@@ -0,0 +1,138 @@
...
+
+    {#- type, category, and criticality are not in US Core.
+        But they are useful for clinical interpretation. #}
+    ta.type,
+    tcat.category,
+    ta.criticality,
+
+    dn_code.code AS code_code,
+    dn_code.system AS code_system,
+    dn_code.display AS code_display,
```
mikix, line 108: OK - is this good naming? Do we prefer just `code/system/display` with no prefix? That feels... asymmetrical with how we treat every other field. But also, `code_code` is slightly awkard (but accurate!).

Hey also - did we every accidentally rename any `AS code_system` stuff to `AS system` in the recent rename? Like, in condition.sql.jina, I see us having a triplet like `code/system/code_display` - which maybe feels like it used to be `code/code_system/code_display` - is that how we'd want this triplet in AllergyIntolerance?

cumulus_library/studies/core/core_templates/allergyintolerance.sql.jinja
```diff
@@ -0,0 +1,138 @@
...
+            )
+        }}
+    FROM flattened_reaction AS r
+    LEFT JOIN core__allergyintolerance_dn_reaction_substance AS dn_subs
+        ON r.id = dn_subs.id AND r.row = dn_subs.row
+    LEFT JOIN core__allergyintolerance_dn_reaction_manifestation AS dn_man
+        ON r.id = dn_man.id AND r.row = dn_man.row
+)
+
+SELECT
```
mikix, line 91: Open to feedback on which fields I've included! The [US Core spec](http://hl7.org/fhir/us/core/STU4/StructureDefinition-us-core-allergyintolerance.html) is wildly bare - not even an encounter field or a recordedDate timestamp.

Stuff like `category` is super useful! (I'd think? Maybe not - maybe all you care about is "am I gonna get sick from this thing")

cumulus_library/studies/core/core_templates/allergyintolerance.sql.jinja
```diff
@@ -0,0 +1,138 @@
...
+        But they are useful for clinical interpretation. #}
+    ta.type,
+    tcat.category,
+    ta.criticality,
+
+    dn_code.code AS code_code,
+    dn_code.system AS code_system,
+    dn_code.display AS code_display,
+
+    ta.patient_ref,
```
mikix, line 110: This is the first core resource with a patient field instead of subject - I'm assuming our vibes here are to keep consistent with the spec? But this patient/subject difference is honestly a headache. In the count code, I chose to keep the output field as cnt_subject_ref regardless.

tests/test_counts_templates.py
```diff
@@ -205,3 +206,11 @@ def test_count_query(expected, kwargs):
...
     #     f.write(query)
     assert query == expected
+
+
+def test_count_query_bad_resource():
+    with pytest.raises(
+        errors.CountsBuilderError,
+        match="Tried to create counts table for invalid resource Medication",
+    ):
+        counts_templates.get_count_query("table", "source", [], fhir_resource="Medication")
```
mikix, line 216: This is my regression test tax :smile:

cumulus_library/studies/core/core_templates/allergyintolerance.sql.jinja
comorbidity: **code_code** is awkward, not a huge deal but I previously used just "code" here.

cumulus_library/studies/core/builder_allergyintolerance.py
```diff
@@ -0,0 +1,66 @@
...
+    def prepare_queries(self, *args, config: cumulus_library.StudyConfig, **kwargs):
+        code_sources = [
+            sql_utils.CodeableConceptConfig(
+                source_table="allergyintolerance",
+                column_hierarchy=[("clinicalStatus", dict)],
+                target_table="core__allergyintolerance_dn_clinical_status",
+                filter_priority=True,
+                code_systems=[
+                    # Restrict to just this required binding system
+                    "http://terminology.hl7.org/CodeSystem/allergyintolerance-clinical",
```
dogversioning, line 31: For this and other code filters: Can we check Cerner data to see if it is providing these codes back? Or what the epic sandbox is doing?

cumulus_library/studies/core/reference_sql/builder_observation.sql
```diff
@@ -95,43 +95,167 @@ CREATE TABLE core__observation_dn_code AS (
...
+        display,
+        userSelected
+    FROM union_table
 );
 
+
 -- ###########################################################
 
-CREATE TABLE IF NOT EXISTS "cumulus_mhg_dev_db"."core__observation_component_valuecodeableconcept"
+CREATE TABLE IF NOT EXISTS "main"."core__observation_component_valuecodeableconcept"
```
dogversioning, line 258: this is an aside, but if you're generating these against duckdb, and i'm generating them against athena, there is going to be constant churn here (though my dev db is definetely also the wrong target)

We should probably pick one target and both try to use that all the time - I was going to say making a staging version of the cerner DB, but that means waiting a half hour for a build to run. Perhaps a standalone for this purpose based on synthea data in athena?

cumulus_library/template_sql/codeable_concept_denormalize.sql.jinja
dogversioning: probably a debugging convenience that hung around

docs/core-study-details.md
dogversioning, line 1: can you update the top section of this document to include the non-spec fields related to allergyintolerance you included?

tests/core/test_core.py
```diff
@@ -125,12 +127,15 @@ def test_core_tiny_database(tmp_path):
     """Verify that we can generate core tables with some minimal data filled in"""
     testbed = testbed_utils.LocalTestbed(tmp_path)
     # Just add bare resources, with minimal data
+    testbed.add_allergy_intolerance("AllA")
```
dogversioning, line 130: I can see myself getting confused by this abbreviation at some point (I already have a hard enough time holding the abstractions against the abbreviations here as is, but this one parses as "All A" too easily). Can it be more verbose?

tests/test_counts_templates.py
dogversioning: 💸

cumulus_library/studies/core/core_templates/allergyintolerance.sql.jinja
dogversioning: I like it until we're told otherwise.

dogversioning (approved):

cumulus_library/studies/core/reference_sql/builder_observation.sql
dogversioning: or, i mean, just always do it in duckdb, though that's not quite accurate as a reference for third parties as to what's going to run live

mikix: I added a [section to the contributing docs](https://github.com/smart-on-fhir/cumulus-library/blob/main/CONTRIBUTING.md#rebuilding-the-reference-sql) a while back with a one-liner that pointed at the duckdb data. So that's our currently-documented process. I didn't mean to slide that by you, I just wanted to document a process that worked.

I agree it's not 100% accurate, but it's easy for randos if they ever filed a PR and quick for everyone. I guess I also have questions about the utility of being fully accurate, versus a smoke test catch of "oh shoot, my SQL is crazier than I thought" (which is how I've been using the reference SQL)

cumulus_library/studies/core/builder_allergyintolerance.py
mikix: `clinicalStatus` and `verificationStatus` are a few of the rare Required bindings in the base spec. That's why I'm being so restrictive/simplistic here. These CodeableConcept fields are basically glorified (read: complicated) Code fields. Every other resource (except Condition) uses simple Code fields for status values, instead of a required-binding-CodeableConcept. :shrug: Either one of those weird "two people working on different parts of spec" things, or they anticipated some local expansions of the status values.

So support is good in EHRs. I checked against sample exports from Epic & Cerner - they both provide it.

cumulus_library/studies/core/reference_sql/builder_observation.sql
dogversioning: ah yep, that's fine, now i just need to remember it

cumulus_library/studies/core/builder_allergyintolerance.py
dogversioning: ok, good deal

mikix:
OK, I just added a few commits, separated out for easier reviewing:
- Have GitHub ignore *.bsv files when detecting language
- Update the docs (and the PR template) around non-US-Core ("optional") fields in the docs - both removed fields referenced by US Core and added missing fields we use that aren't in US Core
- Addressed other review comments
- Switched the default AllergyIntolerance count to be by-patient and added test around that
- Updated the completion tracking to require AllergyIntolerance too

tests/core/test_core.py
mikix: Updated :+1:

cumulus_library/statistics/counts.py
```diff
@@ -6,6 +6,9 @@
 from cumulus_library import base_table_builder, errors, study_manifest
 from cumulus_library.statistics.statistics_templates import counts_templates
 
+# Defined here for easy overriding by tests
+DEFAULT_MIN_SUBJECT = 10
+
```
dogversioning, line 11: ooh, this is a very good idea

dogversioning:
> OK, I just added a few commits, separated out for easier reviewing:

Took a scan, looks good

Issue #243, opened 2024-05-28
Allow a user to get the version of a particular study from the CLI


dogversioning:
It would be helpful for debugging study run issues to have a user do something like `cumulus-library --version covid-symptom` to get the version of a particular study they had installed.

PR #289, opened 2024-08-28
Study version display

dogversioning:
This PR adds the `--study-version` arg, for reading the version of a given study from it's __init__.py. #243 

Also updated the docs to make this new __init__.py requirement to the study authoring docs.

### Checklist


cumulus_library/cli.py
```diff
@@ -392,6 +390,36 @@ def main(cli_args=None):
...
+        args["study_dir"] = posix_paths
+
+    if args.get("action") == "study-version":
+        studies = get_study_dict(args["study_dir"])
+        for target in args.get("target"):
+            if target in ("core", "discovery", "vocab"):
+                print(
+                    f"{target} distributed with Cumulus library.\n"
+                    f"Cumulus library:  {__version__}"
+                )
```
mikix, line 406: I'm a little leery of this special logic here. The user can point at a different / in-development version of these built-in studies with `-s`, where this may not give the right version. And `vocab` is slated for movement - so depending on how that goes down, old versions of `library` may not give the correct version after the move, especially if `-s` is involved again.

I understand you're doing that so that you don't have to keep track of versions inside core/discovery/vocab. But also... maybe we should? Like, `discovery` has its own semantic version lifecycle separate from `core`, which in theory lives a different versioning life from the outer library study runner (though in practice, those two are often closely entwined).

Anyway, just a nit comment, this is fine for now.

cumulus_library/cli_parser.py
```diff
@@ -297,4 +297,10 @@ def get_parser() -> argparse.ArgumentParser:
     add_target_argument(markdown)
     add_verbose_argument(markdown)
 
+    # Get study version
+
+    study_version = actions.add_parser("study-version", help="Gets the version of a specific study")
```
mikix, line 302: This is fine. But I would vote for a single command that shows all known versions it can find. Like... if `--version` (and/or a `version` or `debug` subcommand) gave output like:
```
$ cumulus-library --version
Cumulus Library: 3.0.0
Python: 3.11.1
Studies:
  covid_symptom: 2.1.0
  ...etc...
  umls: 1.2.0
```

Since we register known studies in the Library, we can just iterate over that (plus any extra targets). You could throw in paths to the study modules if you wanted there too.

In my mind, this has some advantages:
- Easier to tell a user how to do it: "run --version" vs "run study-version -t covid_symptom -t core"
- Less error prone for the user
  - harder to forget a `-t core` or similar argument that we'd want to see, everything is built-in
  - or a situation where they run a `study-version -t covid_symptom` line they saved from us a week ago instead of the new `study-version -t umls` command we just gave them.
- Easier to add future useful bits, like the version of some major dependency like Jinja or whatever.

But this approach works too - I'm just brainstorming.

mikix (approved):

tests/test_cli.py
```diff
@@ -807,6 +807,26 @@ def test_version(capfd):
         assert out == __version__
 
 
+@pytest.mark.parametrize(
+    "study,msg",
+    [("core", "distributed_with"), ("study_valid", "version"), ("not_a_study", "not an installed")],
```
mikix, line 812: Maybe use `version 1.0.0` to confirm that we actually do pull the version number from the study

cumulus_library/cli_parser.py
dogversioning: i buy this (as long as I alpha sort the keys for lookup)

cumulus_library/cli.py
mikix: Well I wasn't even talking about repos - just a version in their `__init__.py` files

mikix: But the separate-version approach also matters less if we do a "all versions printed" approach from below.

dogversioning: this was sort of a side thought - i'm doing the all versions thing below for now, i guess i'm going to pin core to a v3 since it's mostly tied to the core infra and discovery to 1.0, but i could see a decoupling case generally? not taking action on it atm.

cumulus_library/studies/template/counts.sql
dogversioning, line 1: While doing this versioning __init__ pass - I'm pulling this since we're no longer doing cli template creation.

cumulus_library/cli_parser.py
dogversioning: ok, this is now the behavior.

PR #291, opened 2024-08-28
Use python import machinery to find known studies

### LINKED ISSUE #290: Library can't find studies installed with `pip install -e .`
**Describe the bug**
When I use `--target covid_symptom` without a `--study-dir`, Library can't find my covid_symptom study, despite it being python-importable.

It seems to do with my use of `pip install -e .` on that study. If I do `pip install .` or `pip install cumulus-library-covid`, it does get picked up.

The `-e .` path does use a weird python install trick of putting in a `.pth` file that points at the real location. So maybe Library just needs to rely on Python import plumbing a little more in this case.

But curiously, it does not seem to affect Matt on his Mac (I'm on Linux).

**To Reproduce**
1. pip install -e . (for a study)
2. cumulus-library version (notice it's not there)

**Expected behavior**
All python-importable studies should be able to be found.

mikix:
Previously, we assumed they'd be directories inside site-packages.

But in some cases (like pip install -e . in linux), that installs a .pth file that points at the real location.

So let's just ask Python where these known modules are.

Fixes #290 

### Checklist

### DIFF
--- a/cumulus_library/module_allowlist.json
+++ b/cumulus_library/module_allowlist.json
@@ -1,11 +1,13 @@
 {
     "desc" : 
         "this dict maps pip installable studies to expected paths inside site-packages",
-    "allowlist" : {
-        "covid_symptom" : "cumulus_library_covid",
-        "data_metrics" : "cumulus_library_data_metrics",
-        "rxnorm" : "cumulus_library_rxnorm",
-        "umls" : "cumulus_library_umls"
-
-    }
+    "allowlist" : [
+        "cumulus_library_covid",
+        "cumulus_library_data_metrics",
+        "cumulus_library_hypertension",
+        "cumulus_library_opioid",
+        "cumulus_library_rxnorm",
+        "cumulus_library_suicidality_los",
+        "cumulus_library_umls"
+    ]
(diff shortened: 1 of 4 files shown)


cumulus_library/module_allowlist.json
```diff
@@ -1,11 +1,13 @@
...
     "desc" : 
         "this dict maps pip installable studies to expected paths inside site-packages",
-    "allowlist" : {
-        "covid_symptom" : "cumulus_library_covid",
-        "data_metrics" : "cumulus_library_data_metrics",
-        "rxnorm" : "cumulus_library_rxnorm",
-        "umls" : "cumulus_library_umls"
-
-    }
+    "allowlist" : [
```
mikix, line 4: I added some new public studies (even though not all are on pypi yet) and removed the name - letting our manifest discovery code find the name (figured that would be more robust to name changes or whatever we do in future). But if you'd prefer to be explicit here, I can put them back no problem.

dogversioning: nah this makes sense.

dogversioning (approved):

PR #295, opened 2024-09-06
Move MedReq and Allergy references to end of table

mikix:
### Checklist


cumulus_library/studies/core/core_templates/medicationrequest.sql.jinja
```diff
@@ -64,60 +64,69 @@ CREATE TABLE core__medicationrequest AS (
...
+        mr.reported_ref,
+
+        uc.medication_code,
+        uc.medication_system,
+        uc.medication_display,
+
+        mr.authoredOn,
+        mr.authoredOn_month,
+
+        concat('MedicationRequest/', mr.id) AS medicationrequest_ref,
```
mikix, line 126: It's a little hard to see because I rearranged how this table gets generated (to more easily inject the code triplet in the middle of the list). But this line here is new - I added it because it seems standard for core tables to have this kind of field (and it makes sense).

cumulus_library/studies/core/reference_sql/builder_encounter.sql
```diff
@@ -582,7 +582,8 @@ temp_encounter_completion AS (
         SELECT
             ece.encounter_id,
             (
-                BOOL_OR(ec.table_name = 'condition')
+                BOOL_OR(ec.table_name = 'allergyintolerance')
+                AND BOOL_OR(ec.table_name = 'condition')
```
mikix, line 586: This is from a previous PR where I forgot to regenerate SQL after making a change.

Since the current flow is to generate SQL from test data, we could add a check in the CI to regenerate SQL and see if there are any changes - failing the PR if so... Just a thought for later

dogversioning: #182 - it's been hanging around for a bit, but it remains a good idea.

dogversioning (approved):

Issue #301, opened 2024-10-03
2016 filtering is still present


mikix:
**Describe the bug**
We thought we got rid of that, but if you search for 2016 in *.jinja files, you'll still see us filtering the core tables by year.

Let's drop those.

PR #304, opened 2024-10-03
core: drop 2016 cutoff from core tables

mikix:
Fixes https://github.com/smart-on-fhir/cumulus-library/issues/301

### Checklist


dogversioning (approved):

tests/regression/reference/core__count_medicationrequest_month.csv
mikix, line 1: This file proves that the other regression updates are alright - we didn't have the 2016 cutoff on the MedicationRequest resource for some reason (maybe it lost it in the refactors or it never had it).

And the only changes here are different quoting.

PR #309, opened 2024-10-24
Add an advanced dynamic_study_prefix field to manifests.

mikix:
This will allow a study to provide a Python file that accepts study options and generates an appropriate study prefix.

It's largely targeted at the data_metrics study, but might have application elsewhere.

### Checklist

### DIFF
--- a/MAINTAINER.md
+++ b/MAINTAINER.md
@@ -1,6 +1,6 @@
 # Maintainer notes
 
-This document is intended for users contributing/manitaining this repository.
+This document is intended for users contributing/maintaining this repository.
 It is not comprehensive, but aims to capture items relevant to architecture
 that are not covered in another document.
 
@@ -10,7 +10,7 @@ Since these terms are used somewhat haphazardly in different database implementa
 we'll quickly define them for the purposes of this document:
 
 - database - a single instance of a database product, local or in the cloud. it can
-contain serveral schemas.
+contain several schemas.
 - schema - a namespaced collection of tables inside a database
 
 Cumulus, as a holistic system, is designed to allow querying against the entire history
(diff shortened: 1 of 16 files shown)


cumulus_library/builders/counts.py
```diff
@@ -14,37 +13,28 @@ class CountsBuilder(BaseTableBuilder):
     """Extends BaseTableBuilder for counts-related use cases"""
 
     def __init__(self, study_prefix: str | None = None):
-        if study_prefix is None:
-            # This slightly wonky approach will give us the path of the
-            # directory of a class extending the CountsBuilder
-            study_path = Path(sys.modules[self.__module__].__file__).parent
```
mikix, line 20: I moved this whole chunk of logic to `prepare_queries` and just used the existing manifest that it sends down, rather than trying to re-discover the manifest and its options.

cumulus_library/study_manifest.py
```diff
@@ -117,13 +133,20 @@ def get_export_table_list(self) -> list[str] | None:
...
+                export_table_list.append(table)
+            elif "__" in table:  # has a prefix, just the wrong one
                 raise errors.StudyManifestParsingError(
                     f"{table} in export list does not start with prefix "
                     f"{self.get_study_prefix()}__ - check your manifest file."
                 )
+            else:
+                # Add the prefix for them (helpful in dynamic prefix cases where the prefix
+                # is not known ahead of time)
+                export_table_list.append(f"{self.get_study_prefix()}__{table}")
```
mikix, line 148: This is a notable change - this allows a study author to do:
```
[export_config]
export_tables = [
  "meta_version",
]
```

without a prefix.

dogversioning: Interesting - I don't think we should encourage that, though. We don't have to change the logic, but let's just not recommend it.

dogversioning (approved):

MAINTAINER.md
```diff
@@ -36,4 +36,58 @@ via cross-database joining. Additional tables should not be created by users in
...
 
 A user could elect to use these vocabulary builders and skip the entire rest of the
-Cumulus ecosystem, if they wanted to. 
\ No newline at end of file
+Cumulus ecosystem, if they wanted to. 
+
+## Advanced study features
+
+These features are for very narrow and advanced use cases,
+designed for internal project studies (like `core`, `discovery`, or `data_metrics`).
```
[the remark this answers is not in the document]
mikix: Yeah fair, I wasn't sure what phrase to use.

PR #314, opened 2024-10-30
Added multiple export types

dogversioning:
This PR breaks out the existing `export_list` section of the manifest into several different typed sections, and appends suffixes based on which list section things are in.

### Checklist


cumulus_library/study_manifest.py
```diff
@@ -111,19 +111,29 @@ def get_statistics_file_list(self) -> list[str] | None:
         stats_config = self._study_config.get("statistics_config", {})
         return stats_config.get("file_names", [])
 
-    def get_export_table_list(self) -> list[str] | None:
+    def get_export_table_list(self) -> list[tuple] | None:
```
mikix, line 114: nit: I don't personally love using tuples as data types. They're so opaque when using them and don't play well with type hinting. A namedtuple or dataclass might be a good lightweight option?

cumulus_library/study_manifest.py
```diff
@@ -111,19 +111,29 @@ def get_statistics_file_list(self) -> list[str] | None:
...
-                raise errors.StudyManifestParsingError(
-                    f"{table} in export list does not start with prefix "
-                    f"{self.get_study_prefix()}__ - check your manifest file."
-                )
+        export_table_list = []
+        for section in [
+            ("export_list", "cube"),
+            ("count_list", "cube"),
+            ("flat_list", "flat"),
+            ("meta_list", "meta"),
```
mikix, line 126: nit: I like the move away from the generic `export_list`, but `count` and `meta` are semantic, while `flat` is syntactic. I know that we maybe don't know the semantic range of flat tables (and/or think the range will be so large, we don't want a field for each), so `flat` is safe. But maybe `count` should be `cube`? And `meta` can continue being a snowflake. Not a strong opinion over here, just bumped on it.

I guess one question is - can we imagine a second kind of cube we'd add in the future? Would we want a new field for it or would we want to put it into the same field?

cumulus_library/study_manifest.py
```diff
@@ -111,19 +111,29 @@ def get_statistics_file_list(self) -> list[str] | None:
...
+        ]:
+            section_list = export_config.get(section[0], []) or []
+            for table in section_list:
+                if not table.startswith(f"{self.get_study_prefix()}__"):
```
mikix, line 130: I think you need to merge from main, because I added a check here to allow tables that don't start with the prefix (for dynamic prefix cases)

tests/test_study_parser.py
```diff
@@ -44,6 +47,7 @@ def test_load_manifest(manifest_path, expected, raises):
         else:
             path = None
         manifest = study_manifest.StudyManifest(path)
+        print(str(manifest))
```
mikix, line 50: :scissors:

tests/test_data/study_valid/manifest.toml
```diff
@@ -8,6 +8,14 @@ file_names = [
...
 export_list = [
-    "study_valid__table",
+    "study_valid__table"
+]
+count_list = [
+    "study_valid__table2"
+]
+flat_list = [
+    "study_valid__table2"
+]
```
mikix, line 18: This is an interesting set up - do we want to error out if the same table is in two lists? When would you intentionally want that?

mikix (approved):

cumulus_library/study_manifest.py
dogversioning: sure, i can dataclass this.

cumulus_library/study_manifest.py
dogversioning: good old naming things problems!

I kind of like semantic for these, in case someone who is not super steeped in this stuff tries to do it - it's a lot clearer to say 'put a count table here from the count generator' than 'make a powerset cube output and put it here'. If I had a semantic name for flat, i would've used it - obviously i don't want to call it 'metric', but maybe there's a better name?

my vibe on cubes is 'maybe at one point we would want a cube that wouldn't be aggregated for some reason'? I'm not sure if I can think of another use case off the dome.

cumulus_library/study_manifest.py
dogversioning: oh jeez, yep - i'll rebase and and adjust

tests/test_data/study_valid/manifest.toml
dogversioning: i don't think you would, no.

I can make this more realistic and add an error - note that this will generate a lot of test churn, sorry in advance.

cumulus_library/study_manifest.py
mikix: Yeah OK - leave it as it is :+1:

tests/test_data/study_valid/manifest.toml
mikix: Hah, jokes on you, I just won't read the diff

mikix: But it does seem nice to catch the duplicate-table-name thing. Seems like a classic copy and paste in a hurry error that folks might actually do.

dogversioning: OK, we'll now throw an error in the case.

PR #315, opened 2024-11-11
Quote local parquet CTAS queries

dogversioning:
This PR quote-escapes local parquet create table statements, in case those statements contain reserved sql words.

### Checklist

### DIFF
--- a/cumulus_library/template_sql/ctas_from_parquet.sql.jinja
+++ b/cumulus_library/template_sql/ctas_from_parquet.sql.jinja
@@ -2,10 +2,12 @@
 {%- if db_type == 'athena' -%}
 CREATE EXTERNAL TABLE IF NOT EXISTS `{{ schema_name }}`.`{{ table_name }}` (
 {%-  elif db_type == 'duckdb' -%}
-CREATE TABLE IF NOT EXISTS {{ table_name }} AS SELECT  
+CREATE TABLE IF NOT EXISTS "{{ table_name }}" AS SELECT  
 {%- endif %}
 {%- for col in table_cols %}
-    {{ col }}{% if db_type == 'athena' %} {{ remote_table_cols_types[loop.index0] }}{%- endif -%}
+    {%- if db_type == 'athena' %} {{ col }} {{ remote_table_cols_types[loop.index0] }}
+    {%- elif db_type == 'duckdb' %} "{{ col }}"
+    {%- endif -%}
     {{- syntax.comma_delineate(loop) }}
 {%- endfor %}   
 {%- if db_type == 'athena' %}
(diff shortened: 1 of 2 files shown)


dogversioning:
I expect unit tests to fail here until the corresponding [library PR](https://github.com/smart-on-fhir/cumulus-library/pull/315) is in, and a release is cut.

cumulus_library/template_sql/ctas_from_parquet.sql.jinja
```diff
@@ -2,10 +2,12 @@
 {%- if db_type == 'athena' -%}
 CREATE EXTERNAL TABLE IF NOT EXISTS `{{ schema_name }}`.`{{ table_name }}` (
 {%-  elif db_type == 'duckdb' -%}
-CREATE TABLE IF NOT EXISTS {{ table_name }} AS SELECT  
+CREATE TABLE IF NOT EXISTS "{{ table_name }}" AS SELECT  
 {%- endif %}
 {%- for col in table_cols %}
-    {{ col }}{% if db_type == 'athena' %} {{ remote_table_cols_types[loop.index0] }}{%- endif -%}
+    {%- if db_type == 'athena' %} {{ col }} {{ remote_table_cols_types[loop.index0] }}
+    {%- elif db_type == 'duckdb' %} "{{ col }}"
```
mikix, line 9: Why only quote for duckdb and not all the time?

tests/test_base_templates.py
```diff
@@ -559,9 +557,7 @@ def test_ctas_empty_query_creation(expected, schema, table, cols, types):
             ["String", "Int"],
         ),
         (
-            """CREATE TABLE IF NOT EXISTS local_table AS SELECT
-    a,
-    b
+            """CREATE TABLE IF NOT EXISTS "local_table" AS SELECT "a", "b"
```
mikix, line 560: So what is the risk you are guarding against in this PR? Like, that a table name wouldn't have a prefix and also be a reserved word? Cause I assume no table with a study prefix could be a problem. Should we instead/additionally enforce that there is a prefix (look for a dunder)?

mikix (approved):

cumulus_library/template_sql/ctas_from_parquet.sql.jinja
dogversioning: Athena didn't like that, irritatingly - the serde style queries follow a slightly different set of syntax rules, it seems?

tests/test_base_templates.py
dogversioning: Specifically, the LOINC groups dataset has a file just named 'Group', and columns named 'group'. I think you're right that, if a dunder was present, this would largely not be an issue. But since this is one of these static dataset situations that you want to use across multiple studies/datasources, I'm inclined to just live with the quoting here?

mikix: Ah yeah and the "dundering" there is done via a different namespace. Sure - quoting is fine, was just curious if we should go even further.

PR #317, opened 2024-11-18
Converted manifest to single list of files

dogversioning:
- Manifest list of files is now a single list, and finds the appropriate running mechanism via file extension
  - The old file lists are deprecated, but will still work for older studies
- TOML configuration-driven table construction is now called a workflow

### Checklist


cumulus_library/actions/builder.py
```diff
@@ -256,34 +211,52 @@ def run_matching_table_builder(
...
     for file in all_generators:
-        if builder and file.find(builder) == -1:
-            continue
-        _load_and_execute_builder(
-            config=config,
-            manifest=manifest,
-            filename=file,
-            db_parser=db_parser,
-        )
+        if builder and file.find(builder) != -1:
```
mikix, line 216: should this be `if not builder or file.find():`? -- if the builder option is falsy, nothing gets included - is that what we want?

cumulus_library/builders/protected_table_builder.py
```diff
@@ -64,12 +67,22 @@ def prepare_queries(
...
-                    db_schema,
-                    statistics,
-                    STATISTICS_COLS,
-                    STATISTICS_COLS_TYPES,
-                )
-            )
+        files = manifest.get_file_list()
+        files = [file for file in files if file.endswith(".toml")]
+        if len(files) == 0:
+            return
```
mikix, line 73: nit: this check is redundant - the loop will immediately end below if there are no files.

cumulus_library/builders/protected_table_builder.py
```diff
@@ -64,12 +67,22 @@ def prepare_queries(
...
-            self.queries.append(
-                base_templates.get_ctas_empty_query(
-                    db_schema,
-                    statistics,
-                    STATISTICS_COLS,
-                    STATISTICS_COLS_TYPES,
-                )
-            )
+        files = manifest.get_file_list()
+        files = [file for file in files if file.endswith(".toml")]
```
mikix, line 71: This method works a bit differently than I'd expect - it looks through the file list itself, finds the first statistics type and then returns after creating tables for the first one it finds.

Why isn't this done here and not as part of a workflow (and thus part of the normal walk-through-the-file-list sequence), inside PsmBuilder or whatever?

cumulus_library/study_manifest.py
```diff
@@ -179,11 +206,13 @@ def get_export_table_list(self) -> list[ManifestExport] | None:
 
     def get_all_generators(self) -> list[str]:
         """Convenience method for getting files that generate sql queries"""
-        return (
-            self.get_table_builder_file_list()
-            + self.get_counts_builder_file_list()
-            + self.get_statistics_file_list()
-        )
+        files = self.get_file_list()
+        return [file for file in files if file.endswith(".py")]
```
mikix, line 210: I feel like this is a pattern I'm seeing a bit in this PR - do you want to add an optional suffix argument to `get_file_list`?

cumulus_library/builders/protected_table_builder.py
dogversioning: So what this is doing is creating a table that statistic type runs can write to, if it doesn't exist - the idea being that if there is more than one set of statistics, they'll all have one place to go - and also that this table will persist between individual study runs. It's basically a registry of all the sampled statistics.

I could add a clarifying comment?

cumulus_library/builders/protected_table_builder.py
dogversioning: what if you want to be :really: sure, though

cumulus_library/study_manifest.py
dogversioning: i could? i kind of like the semantics of a function that is named after what it does, but those could just be wrappers for something in one place that takes an arg

cumulus_library/actions/builder.py
dogversioning: yeah that might read better

PR #323, opened 2024-12-06
Packaging mode for studies

dogversioning:
-Adds a new cli param to the build argument, `--prepare`, which renders queries instead of execute them
-Adds a fixture to replace individual instances of clearing os environment variables

### Checklist


cumulus_library/actions/builder.py
```diff
@@ -89,7 +97,10 @@ def _load_and_execute_builder(
     # remove it so it doesn't interfere with the next python module to
     # execute, since the subclass would otherwise hang around.
     table_builder_class = table_builder_subclasses[0]
-    table_builder = table_builder_class()
+    if issubclass(table_builder_class, CountsBuilder):
+        table_builder = table_builder_class(study_prefix=manifest.get_study_prefix())
+    else:
+        table_builder = table_builder_class()
```
mikix, line 103: nit: should we just eventually retool builders to just accept a manifest object in the init method?

cumulus_library/actions/builder.py
```diff
@@ -136,13 +158,31 @@ def run_protected_table_builder(
...
     :param manifest: a StudyManifest object
+    :param filename: Filename of the workflow config
+    :param prepare: If true, will render query instead of executing
+    :param data_path: If prepare is true, the path to write rendered data to
+    :param query_count: if prepare is true, the number of queries already rendered
+    :returns: a count of processed queries
     """
+    toml_path = pathlib.Path(f"{manifest._study_path}/{filename}")
+    if prepare:
+        with open(toml_path) as file:
```
mikix, line 180: Interesting... I am use to the pylint check to always pass an `encoding=` arg to `open` (usually `encoding='utf8'`). Ruff isn't doing that here? I might be interested in seeing which package that's in and adding those checks, since that seems like a real issue (not high risk, but not zero) that can be hard to catch on dev machines and hard to debug later.

cumulus_library/actions/builder.py
```diff
@@ -136,13 +158,31 @@ def run_protected_table_builder(
...
+            workflow_config = file.read()
+        _render_output(
+            manifest.get_study_prefix(), [workflow_config], data_path, filename, query_count, True
```
mikix, line 183: nit: I hate to see a bare True/False - can you make that parameter kwarg only? (using `*` in the param list).

cumulus_library/actions/builder.py
```diff
@@ -228,35 +285,87 @@ def build_study(
...
     :keyword continue_from: Name of a file to resume table creation from
-    :returns: loaded queries (for unit testing only)
+    :keyword prepare: If true, will render query instead of executing
+    :keyword data_path: If prepare is true, the path to write rendered data to
     """
+    if prepare and manifest.get_study_prefix() in ["core", "discovery", "data-metrics"]:
+        sys.exit(
+            f"{manifest.get_study_prefix()} does not support prepare mode. It must be run "
+            "directly against a target database."
+        )
```
mikix, line 304: nit: I saw a user string repeated, so my "move this to a utility method" alarm went off.

cumulus_library/actions/builder.py
```diff
@@ -228,35 +285,87 @@ def build_study(
...
     if file_list is None:
         file_list = manifest.get_file_list(continue_from)
+    if prepare:
+        data_dir = data_path / manifest.get_study_prefix()
+        files = data_dir.glob("*")
+        [file.unlink() for file in files if file.is_file()]
```
mikix, line 310: nit: a "logic-only" list comprehension like that feels like an odd construction. You could still do this in just one more line with:
```
        for file in data_dir.glob("*"):
            if file.is_file():
                file.unlink()
```

cumulus_library/actions/builder.py
```diff
@@ -203,13 +243,23 @@ def build_matching_files(
     *,
     builder: str | None,
     db_parser: databases.DatabaseParser = None,
+    prepare: bool = False,
+    data_path: pathlib.Path,
```
mikix, line 247: Same comment here about default args - This feature feels like a thing that a future dev (including us) might totally forget exists when adding code that calls these methods. If everyone calling this is passing `prepare=xxx` right now (though I did not check if that were true here, to be fair), I'm guessing we'll want all callers in future to pass it in too.

cumulus_library/actions/builder.py
```diff
@@ -228,35 +285,87 @@ def build_study(
...
+                data_path=data_path,
+                query_count=query_count,
+            )
         else:
             raise errors.StudyManifestParsingError
+    if prepare:
+        with zipfile.ZipFile(
+            f"{data_path}/{manifest.get_study_prefix()}.zip", "w", zipfile.ZIP_DEFLATED
+        ) as z:
+            for file in data_dir.glob("*"):
```
mikix, line 347: Is there a reason you use `glob("*")` a couple times, over `iterdir()`? Is there a functional difference, like does `glob(*)` skip hidden files with `.` prefixes (and also: is that desirable)? Or just a matter of dev preference?

cumulus_library/actions/builder.py
```diff
@@ -228,35 +285,87 @@ def build_study(
...
 
 def _run_raw_queries(
-    config: base_utils.StudyConfig, manifest: study_manifest.StudyManifest, filename: str
-):
+    config: base_utils.StudyConfig,
+    manifest: study_manifest.StudyManifest,
+    filename: str,
+    prepare: bool,
+    data_path: pathlib.Path,
+    query_count: int,
```
mikix, line 357: nit: again, bool and int args (plus a long arg list) make me want to require some kwargs with `*`. Not really necessary, I'm just a big fan of the feature, apparently :smile:

cumulus_library/actions/builder.py
```diff
@@ -266,25 +375,49 @@ def _run_raw_queries(
...
+            )
+    return len(queries)
+
+
+def _render_output(study_name, outputs, data_path, filename, count, is_toml=False):
+    if is_toml:
+        suffix = "toml"
+    else:
+        suffix = "sql"
+    for output in enumerate(outputs):
```
mikix, line 409: I personally find it more readable to do something like `for index, output in enumerate(outputs):` so you can skip the odd `output[0]` and `output[1]` stuff

cumulus_library/actions/builder.py
```diff
@@ -266,25 +375,49 @@ def _run_raw_queries(
...
+            name = "config"
+        else:
+            name = re.search(r"(.*)__\w*", output[1])[0].lower().replace(" ", "_")
+        file_path = data_path / (
+            f"{study_name}/{count+output[0]:04d}.{filename.rsplit('.', 1)[0]}"
+            f".{output[0]:02d}.{name}.{suffix}"
         )
-    return queries
+        file_path.parent.mkdir(exist_ok=True, parents=True)
+        with open(file_path, "w") as f:
```
mikix, line 419: nit: encoding='utf8'

cumulus_library/actions/builder.py
```diff
@@ -266,25 +375,49 @@ def _run_raw_queries(
...
+        suffix = "sql"
+    for output in enumerate(outputs):
+        if is_toml:
+            name = "config"
+        else:
+            name = re.search(r"(.*)__\w*", output[1])[0].lower().replace(" ", "_")
+        file_path = data_path / (
+            f"{study_name}/{count+output[0]:04d}.{filename.rsplit('.', 1)[0]}"
+            f".{output[0]:02d}.{name}.{suffix}"
         )
```
mikix, line 417: These lines are pretty dense - I think the regex search line could use a small comment about like "# search for table name" or something. And I think the file path building might benefit from intermediate variables, like:
```
total_index = count + index
new_filename = f"{total_index:04d}.{filename.stem}.{index:02d}.{name}.{suffix}" 
file_path = data_path / study_name / new_filename
```

mikix (approved):

tests/test_data/study_dynamic_prefix/reference_sql/counts.sql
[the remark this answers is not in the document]
dogversioning: probably a mistake, lemme verify but i'll likely trim

cumulus_library/actions/builder.py
dogversioning: oh gosh, probably?

cumulus_library/actions/builder.py
dogversioning: yeah ruff didn't flag that, i'll see if i can find the right rule for it.

cumulus_library/actions/builder.py
dogversioning: i skipped it since it was in module and internal, but for you? sure.

cumulus_library/actions/builder.py
dogversioning: no particular reason here, other than 'i was already thinking with glob', fine to switch.

cumulus_library/actions/builder.py
dogversioning: yeah that's fine, same comment about being lazy with internal functions.

cumulus_library/cli_parser.py
```diff
@@ -38,7 +38,7 @@ def add_custom_option(parser: argparse.ArgumentParser) -> None:
 def add_data_path_argument(parser: argparse.ArgumentParser) -> None:
     """Adds path arg to a subparser"""
     parser.add_argument(
-        "data_path",
+        "--data_path",
```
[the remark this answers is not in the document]
dogversioning: lemme make an alias for the old behavior (which i think is/has always been wrong) - goes to show you how often i've been getting this in another way.

cumulus_library/cli_parser.py
```diff
@@ -208,6 +208,11 @@ def get_parser() -> argparse.ArgumentParser:
         action="store_true",
         help="Forces file downloads/uploads to occur, even if they already exist",
     )
+    build.add_argument(
+        "--prepare",
+        action="store_true",
+        help=("Prepares queries for remote execution."),
+    )
```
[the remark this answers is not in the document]
dogversioning: - parens, no idea - may have been longer at one point and they were ruff inserted

- i can supress this, this is probably mostly just for us shipping data via queues

- there is :so: much overlap with build under the hood, i'm reluctant to make it standalone.

cumulus_library/cli_parser.py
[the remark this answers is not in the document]
dogversioning: ok you know what - this is more trouble than it's worth, i'm leaving it as is.

cumulus_library/actions/builder.py
dogversioning: #324

cumulus_library/actions/builder.py
dogversioning: https://docs.astral.sh/ruff/rules/unspecified-encoding/ - since its marked unstable, i'm inclined to not do it for now.

cumulus_library/actions/builder.py
dogversioning: ok, one difference - if you glob a directory that does not exist, you do not throw an error, wheras iterdir does. i may keep this for the 'delete the files in the dir if they already exist' check.

cumulus_library/cli.py
```diff
@@ -93,17 +93,24 @@ def clean_and_build_study(
         *,
         options: dict[str, str],
         continue_from: str | None = None,
+        prepare: bool = False,
+        data_path: pathlib.Path | None = None,
```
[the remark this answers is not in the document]
dogversioning: I'm leaving this in at the main entrypoint since a lot of tests assume it's there, but i've pulled everything downstream

PR #328, opened 2024-12-16
Breakout connection from athena DB init

dogversioning:
The automatic db connection is causing some issues for the packaging use case (where we want to generate sql based on db type, but aren't actually using a connection), so this creates an explicit connection function that is called seperately from the database object init. Connection logic is then moved from the individual db init functions to the connect function.

### Checklist

### DIFF
--- a/cumulus_library/databases/athena.py
+++ b/cumulus_library/databases/athena.py
@@ -105,8 +105,11 @@ class AthenaDatabaseBackend(base.DatabaseBackend):
                     output.append((column[0], pyarrow.int64()))
                 case "double":
                     output.append((column[0], pyarrow.float64()))
+                # This is future proofing - we don't see this type currently.
                 case "decimal":
-                    output.append((column[0], pyarrow.decimal128(column[4], column[5])))
+                    output.append(  # pragma: no cover
+                        (column[0], pyarrow.decimal128(column[4], column[5]))
+                    )
                 case "boolean":
                     output.append((column[0], pyarrow.bool_()))
                 case "date":
@@ -171,8 +174,8 @@ class AthenaDatabaseBackend(base.DatabaseBackend):
             glue_client.create_database(DatabaseInput={"Name": schema_name})
 
     def close(self) -> None:
-        if self.connection is not None:
-            self.connection.close()  # pragma: no cover
+        if self.connection is not None:  # pragma: no cover
+            self.connection.close()
 
 
 class AthenaParser(base.DatabaseParser):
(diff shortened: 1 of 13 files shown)


cumulus_library/databases/athena.py
```diff
@@ -32,19 +32,22 @@ def __init__(self, region: str, work_group: str, profile: str, schema_name: str)
         self.work_group = work_group
         self.profile = profile
         self.schema_name = schema_name
+        self.connection = None
+
+    def connect(self):
         # the profile may not be required, provided the above three AWS env vars
         # are set. If both are present, the env vars take precedence
```
mikix, line 39: This comment is lightly out of date now, since it refers to "above three AWS env vars" and it got moved around

cumulus_library/databases/athena.py
```diff
@@ -102,8 +105,11 @@ def col_pyarrow_types_from_sql(self, columns: list[tuple]) -> list:
                     output.append((column[0], pyarrow.int64()))
                 case "double":
                     output.append((column[0], pyarrow.float64()))
+                # This is future proofing - we don't see this type currently.
```
mikix, line 108: I thought we did see it in one edge case - I don't think a normal production flow, but it might have been when dealing with flat tables? Or I'm making it up.

cumulus_library/databases/base.py
```diff
@@ -120,6 +120,10 @@ def __init__(self, schema_name: str):
         # technology
         self.db_type = None
 
+    @abc.abstractmethod
+    def connect(self):
+        """Initiates connection configuration of the database"""
```
mikix, line 125: If we wanted to be Extremely Python :tm:, we would probably write this abstract class as a context manager, so you could do:
```
with AthenaDatabase(...) as db:
    # use db
```

To automatically get connect/close support (and make sure that callers don't forget one of the calls, even when an exception happens).

But not necessary right now and can often be a little annoying to reorganize code to support that style (but does have some benefits - mainly the exception handling).

cumulus_library/databases/utils.py
```diff
@@ -110,5 +108,10 @@ def create_db_backend(args: dict[str, str]) -> (base.DatabaseBackend, str):
             sys.exit("Loading an ndjson dir is not supported with --db-type=athena.")
     else:
         raise errors.CumulusLibraryError(f"'{db_config.db_type}' is not a supported database.")
-
+    if "prepare" not in args.keys():
+        backend.connect()
+    elif not args["prepare"]:
+        backend.connect()
```
mikix, line 114: nit: `if not args.get("prepare"):`

PR #329, opened 2024-12-30
Add core__procedure table

mikix:
Puts another brick in the wall of #139

### Checklist


cumulus_library/studies/core/core_templates/completion_utils.jinja
```diff
@@ -45,6 +45,7 @@ in this table.
                 AND BOOL_OR(ec.table_name = 'documentreference')
                 AND BOOL_OR(ec.table_name = 'medicationrequest')
                 AND BOOL_OR(ec.table_name = 'observation')
+                AND BOOL_OR(ec.table_name = 'procedure')
```
mikix, line 48: So I think this deserves a tiny callout.

For context, the way completion works is that we ignore any encounters for which all the listed associated tables have not been loaded by the ETL yet.

So adding a new resource like this to the list of "required resources" means:
- _Transition pain_: if a site hasn't ETL'd procedures yet, the next time they rebuild `core`, all encounters will disappear.
- _Ongoing pain_: `core` requires more and more resources (like procedures) to be exported and ETL'd before it can function as expected

However... This is also... the point of completion? Would we want to draw a line between "resources we actually care about" and "resources we do technically support"? If we didn't add Procedure here, and your study used `core__procedure`, you would now be subject to the reasons we added the completion table - engineers ETL'ing data can mess you up.

This "ignore encounters" trick was so that studies didn't have to know about the whole completion table feature. They just would `inner join` on encounters at some point and incomplete data would be ignored. But maybe we have a list of 2nd-tier resources for which studies are expected to manually check the completion tables? (the logic isn't necessarily fun, but we could create some docs with examples)

Or we just accept the pains listed above as we grow.

comorbidity:
Should include Procedure **Category** , which "should" be implemented at most sites because it is easy, BCH Cerner is not a good indicator for procedure coding because it was DIY home grown.

comorbidity:
> Should include Procedure **Category** , which "should" be implemented at most sites because it is easy, BCH Cerner is not a good indicator for procedure coding because it was DIY home grown.

STU4 https://hl7.org/fhir/us/core/STU4/StructureDefinition-us-core-procedure.html
https://hl7.org/fhir/R4/valueset-procedure-category.html

mikix:
> Should include Procedure Category , which "should" be implemented at most sites because it is easy, BCH Cerner is not a good indicator for procedure coding because it was DIY home grown.

Done!

cumulus_library/studies/core/core_templates/completion_utils.jinja
dogversioning: I'm of two minds about this:

- Long term, I think this is ok/most correct.
- Short term, we are getting a lot of partial drips of data, and this makes it harder for folks to spin up/participate

I kinda want to have a discussion at the product level about what's best here and then circle back on the implementation side of things? We can take this as is for now and then backsolve later if we want.

cumulus_library/studies/core/core_templates/procedure.sql.jinja
```diff
@@ -0,0 +1,93 @@
...
+
+-- This table includes all fields of interest to the US Core Procedure profile.
+-- EXCEPT FOR:
+-- * the 'performedAge' and 'performedRange' fields, simply because they are annoying to
+--   represent and not frequently used. They aren't even marked as Must Support by the profile
+--   (heck, neither is performedPeriod, but we include that since EHRs often like to use periods)
+--
+-- AND ADDING:
+-- * the `category` field, because it's helpful for classification
+-- * the `encounter` field, because come on, why is it left out of the US Core profile
```
dogversioning, line 11: you've got to be kidding me

dogversioning (approved):

cumulus_library/studies/core/core_templates/completion_utils.jinja
mikix: After discussion on slack, we're still OK with this for now. Will land as-is.

PR #332, opened 2025-01-06
docs: explain which resources are completion tracked

mikix:
### Checklist


cumulus_library/__init__.py
```diff
@@ -6,4 +6,4 @@
 from cumulus_library.study_manifest import StudyManifest
 
 __all__ = ["BaseTableBuilder", "CountsBuilder", "StudyConfig", "StudyManifest"]
-__version__ = "4.1.3"
+__version__ = "4.2.0"
```
mikix, line 9: Only because I forgot to bump the version in the last "add core__procedure" PR. With this change, you _could_ release when you want, but I have no urgency for it.

dogversioning (approved):

PR #334, opened 2025-01-07
core: add some missing Encounter.class codes

mikix:
These codes were not handled, but now are:
- OBSENC (v3)
- I (v2 -> IMP)
- P (v2 -> PRENC)

Additionally, we verify the code system too now.

Fixes: #333

### Checklist

### DIFF
--- a/cumulus_library/studies/core/fhir_mapping_tables.sql
+++ b/cumulus_library/studies/core/fhir_mapping_tables.sql
@@ -118,23 +118,31 @@ SELECT * FROM
 
 -- ############################################################
 -- FHIR mapping of as found Encounter codes to the expected encounter code from
--- http://hl7.org/fhir/STU3/v3/ActEncounterCode/vs.html
+-- http://hl7.org/fhir/R4/v3/ActEncounterCode/vs.html
 
 CREATE TABLE core__fhir_mapping_expected_act_encounter_code_v3 AS
 SELECT * FROM
     (
         VALUES
-        ('AMB', 'AMB'),
-        ('AMB', 'R'),
-        ('AMB', 'O'),
-        ('EMER', 'EMER'),
-        ('EMER', 'E'),
-        ('FLD', 'FLD'),
-        ('HH', 'HH'),
-        ('IMP', 'IMP'),
-        ('ACUTE', 'ACUTE'),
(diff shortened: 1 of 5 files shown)


cumulus_library/studies/core/fhir_mapping_tables.sql
```diff
@@ -118,23 +118,31 @@ SELECT * FROM
...
+        ('SS', 'SS', 'http://terminology.hl7.org/CodeSystem/v3-ActCode'),
+        ('VR', 'VR', 'http://terminology.hl7.org/CodeSystem/v3-ActCode'),
+        -- http://terminology.hl7.org/CodeSystem/v2-0004
+        -- This is an older class system, still in active use apparently (at BCH Cerner).
+        -- Unmapped codes: B (obstetrics), C (commercial), N (not applicable), U (unknown)
+        ('EMER', 'E', 'http://terminology.hl7.org/CodeSystem/v2-0004'),
+        ('IMP', 'I', 'http://terminology.hl7.org/CodeSystem/v2-0004'),
+        ('AMB', 'O', 'http://terminology.hl7.org/CodeSystem/v2-0004'),
+        ('PRENC', 'P', 'http://terminology.hl7.org/CodeSystem/v2-0004'),
+        ('AMB', 'R', 'http://terminology.hl7.org/CodeSystem/v2-0004')  -- recurring patient
```
dogversioning, line 147: have we checked this against a recent-ish epic? just to make sure these codes are actually present in the form we'd expect as opposed to the proprietary format?

dogversioning (approved):

cumulus_library/studies/core/fhir_mapping_tables.sql
mikix: In the export I have from the Epic sandbox, it's always `urn:oid:1.2.840.114350.1.72.1.7.7.10.696784.13260` and codes like `13` or `17`

mikix: And this field is just a raw coding - so only room for one code & system pair :frowning:

dogversioning: Can we use multiple entries in this table? like core/fhir_mapping_tables.sql?

Issue #324, opened 2024-12-09
Builders should take a manifest as an init argument


dogversioning:
**Problem Statement**
We sometimes need a piece of data from the manifest/config in a builder - we could just default to passing them as init arguments and ignoring them if not needed

**Desired behavior**
Consistent repeatable behavior for getting data into builders regardless of where it happens

PR #342, opened 2025-01-21
Better duckdb schema, manifest/template entrypoint tweaks

dogversioning:
This PR primarily updates how schemas are handled in duckdb for insert/parquet CTAS cases.

A few minor things that are done along the way:
- Builders now take a manifest as an optional argument #324
  - This allows us to skip some jankiness bootstrapping counts builders with a study prefix. Marked this as deprecated.
- Since the `get_base_templates` function is used frequently to get tables in studies with the core macros loaded in, made the function `get_templates` and converted the former to an alias

### Checklist

### DIFF
--- a/cumulus_library/builders/counts.py
+++ b/cumulus_library/builders/counts.py
@@ -27,6 +27,10 @@ class CountsBuilder(BaseTableBuilder):
                 " and will be removed in a future version"
             )
             self.study_prefix = study_prefix
+        else:
+            raise errors.CountsBuilderError(
+                "CountsBuilder should be initiated with a valid manifest.toml"
+            )
 
     def get_table_name(self, table_name: str, duration=None) -> str:
         """Convenience method for constructing table name
@@ -34,11 +38,6 @@ class CountsBuilder(BaseTableBuilder):
         :param table_name: table name to add after the study prefix
         :param duration: a time period reflecting the table binning strategy
         """
-        if not self.study_prefix:
-            raise errors.CountsBuilderError(
-                "CountsBuilder must be either initiated with a study prefix, "
-                "or be in a directory with a valid manifest.toml"
-            )
         if duration:
             return f"{self.study_prefix}__{table_name}_{duration}"
         else:
@@ -357,5 +356,4 @@ class CountsBuilder(BaseTableBuilder):
         This should be overridden in any count generator. See studies/core/count_core.py
         for an example
         """
-        if manifest and not self.study_prefix:
-            self.study_prefix = manifest.get_study_prefix()
+        pass
--- a/cumulus_library/builders/counts.py
+++ b/cumulus_library/builders/counts.py
@@ -2,6 +2,8 @@
 
 from pathlib import Path
 
+from rich.console import Console
+
 from cumulus_library import BaseTableBuilder, errors, study_manifest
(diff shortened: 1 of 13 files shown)


cumulus_library/builders/counts.py
```diff
@@ -12,9 +14,19 @@
...
         super().__init__()
-        self.study_prefix = study_prefix
+        if manifest:
+            self.study_prefix = manifest.get_study_prefix()
+        else:
+            console = Console()
+            console.print(
+                "[yellow]Warning: providing study_prefix to a CountsBuilder is deprecated"
+                " and will be removed in a future version"
+            )
```
mikix, line 28: This error is lightly awkwardly worded, for the case where you don't provide either argument. Should this error message be behind an `if study_prefix:` guard?

tests/test_data/study_python_counts_valid/module1.py
```diff
@@ -4,7 +4,7 @@
 class ModuleOneRunner(cumulus_library.CountsBuilder):
     display_text = "module1"
 
-    def __init__(self, study_prefix):
+    def __init__(self, *args, **kwargs):
         super().__init__(study_prefix="study_python_counts_valid")
```
mikix, line 8: Alternatively, you could delete this __init__ block now if the prefix can be guessed from the manifest, yeah?

tests/test_data/study_python_counts_valid/module2.py
```diff
@@ -4,7 +4,7 @@
 class ModuleTwoRunner(cumulus_library.CountsBuilder):
     display_text = "module2"
 
-    def __init__(self, study_prefix):
+    def __init__(self, *args, **kwargs):
         super().__init__(study_prefix="study_python_counts_valid")
```
mikix, line 8: same "can we delete this block?" comment here

cumulus_library/template_sql/base_templates.py
```diff
@@ -17,7 +17,12 @@ class TableView(enum.Enum):
     VIEW = "VIEW"
 
 
-def get_base_template(filename_stem: str, path: pathlib.Path | None = None, **kwargs: dict) -> str:
+def get_base_template(*args, **kwargs):
+    """Legacy wrapper for get_template"""
+    return get_template(*args, **kwargs)
+
+
+def get_template(filename_stem: str, path: pathlib.Path | None = None, **kwargs: dict) -> str:
```
mikix, line 25: So are we promoting this to public API? If so:
- do we want to expose it in the main `__init__.py` file then?
- that implies promoting all the "base" templates and macros to be public too - do we have thoughts about stability for that chunk of code across Library releases?
- or maybe it doesn't imply that, and we can maybe remove the default (no `path` provided) behavior of looking in the base template folder. (and make `get_base_template` not just a legacy wrapper, but the version used by Library-internal code, which keeps that default)

mikix (approved):

cumulus_library/template_sql/base_templates.py
dogversioning: For posterity:

We're keeping this in the current state, which is a semi-private API that tied to major versions for breaking changes, but we are going to move the function interface to the module level and add a note about intention and caveats. We may make this API fully public in the future.

PR #344, opened 2025-01-30
fix: don't load NDJSON data into memory all at once

mikix:
This commit changes how --load-ndjson-dir works, telling DuckDB to load the data from files on disk itself, rather than us loading that data into memory all at once, then handing it to DuckDB.

Which allows querying larger-than-memory data sets.

But... the SQL itself can still be a memory bottleneck, if it requires loading too much data into memory.

### Checklist

### DIFF
--- a/cumulus_library/databases/athena.py
+++ b/cumulus_library/databases/athena.py
@@ -70,7 +70,7 @@ class AthenaDatabaseBackend(base.DatabaseBackend):
     def parser(self) -> base.DatabaseParser:
         return AthenaParser()
 
-    def operational_errors(self) -> tuple[Exception]:
+    def operational_errors(self) -> tuple[type[Exception], ...]:
         return (pyathena.OperationalError,)
 
     def col_parquet_types_from_pandas(self, field_types: list) -> list:
--- a/cumulus_library/databases/utils.py
+++ b/cumulus_library/databases/utils.py
@@ -1,27 +1,62 @@
 import pathlib
 import sys
+from collections.abc import Iterable
 
 import cumulus_fhir_support
 import pyarrow
+import pyarrow.dataset
 
-from cumulus_library import db_config, errors
+from cumulus_library import base_utils, db_config, errors
 from cumulus_library.databases import athena, base, duckdb
 
 
-def _read_rows_for_resource(path: pathlib.Path, resource: str) -> list[dict]:
-    rows = []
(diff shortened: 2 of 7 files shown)


cumulus_library/databases/athena.py
```diff
@@ -70,7 +70,7 @@ def execute_as_pandas(
     def parser(self) -> base.DatabaseParser:
         return AthenaParser()
 
-    def operational_errors(self) -> tuple[Exception]:
+    def operational_errors(self) -> tuple[type[Exception], ...]:
```
mikix, line 73: Unrelated, but I noticed that this method had incorrect typing (I think my fault).

cumulus_library/databases/utils.py
```diff
@@ -49,32 +82,49 @@ def read_ndjson_dir(path: str) -> dict[str, pyarrow.Table]:
...
-        if rows:
-            # Auto-detecting the schema works for these simple tables
-            all_tables[metadata_table] = pyarrow.Table.from_pylist(rows)
+        if dataset := _load_custom_etl_table(f"{path}/{metadata_table}"):
+            all_tables[metadata_table] = dataset
 
     return all_tables
 
 
+def _handle_load_ndjson_dir(args: dict[str, str], backend: base.DatabaseBackend) -> None:
```
mikix, line 104: This method is mostly a refactor, with the only new functionality a progress bar for loading scanning the NDJSON for its schema. (Which in my 26G test folder takes four minutes.)

I was feeling like the load-ndjson-dir logic was spread around a few if/elses in a way that was hard to reason about. So moving it all to one method helped, especially now that there is extra code for the progress bar.

cumulus_library/databases/utils.py
```diff
@@ -1,27 +1,60 @@
...
+    # Instead, we let PyArrow infer the types, with one hiccup: its JSON parser will interpret
+    # datetime strings as a TIMESTAMP type, which isn't insane, but not what we want - we want to
+    # keep those as strings. So we let PyArrow infer from Python objects, where it works like we
+    # want.
+    #
+    # Thus, we sip on the data a bit to infer a schema from the first row as a Python object.
+    first_row = next(iter(_rows_from_files(files)), None)
+    if not first_row:
+        return None
+    schema = pyarrow.Table.from_pylist([first_row]).schema
```
dogversioning, line 50: noting (mostly for myself) that this may be a reasonable way to get out schemas when generating tables from parquet

dogversioning (approved):

PR #345, opened 2025-02-05
fix: avoid memory bottlenecks when flattening tables

mikix:
With DuckDB, row_number() over an unnested field seems to blow up the memory usage, rather than spilling over onto the disk.

So instead, this commit switches the flatten() macro into two branches:
- DuckDB branch: switch to generate_subscripts() which Athena does not have. This is memory-performant.
- Fallback/Athena branch: switch to WITH ORDINALITY which DuckDB does not have. This is simpler code, and DuckDB will hopefully one day support WITH ORDINALITY.

This also fixes a bug where the order of rows was not guaranteed by the SQL engine. This seemed to only affect DuckDB. Now both branches guarantee proper ordering.

### Checklist

### DIFF
--- a/cumulus_library/template_sql/base_templates.py
+++ b/cumulus_library/template_sql/base_templates.py
@@ -42,7 +42,8 @@ def get_template(filename_stem: str, path: pathlib.Path | None = None, **kwargs:
             macro_paths.append(path)
         loader = jinja2.FileSystemLoader(macro_paths)
         env = jinja2.Environment(loader=loader).from_string(template)  # noqa: S701
-        return env.render(**kwargs, db_type=db_config.db_type)
+        env.globals["db_type"] = db_config.db_type
+        return env.render(**kwargs)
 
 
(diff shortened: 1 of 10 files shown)


cumulus_library/template_sql/base_templates.py
```diff
@@ -42,7 +42,8 @@ def get_template(filename_stem: str, path: pathlib.Path | None = None, **kwargs:
             macro_paths.append(path)
         loader = jinja2.FileSystemLoader(macro_paths)
         env = jinja2.Environment(loader=loader).from_string(template)  # noqa: S701
-        return env.render(**kwargs, db_type=db_config.db_type)
+        env.globals["db_type"] = db_config.db_type
+        return env.render(**kwargs)
```
mikix, line 46: This change means we don't have to pass db_type down into every macro that wants to use it.

dogversioning (approved):

Issue #339, opened 2025-01-17
counts tables should always be named like `[study]__count_[resource]_...`


dogversioning:
**Problem Statement**
The dashboard parses data from the table name in a specific way to make certain descisions

**Desired behavior**
The counts builder should prepend count table names appropriately, rather than expecting a user to do so

dogversioning:
@mikix @vlad-ignatov Flagging this issue as the one for us to come back to after we discuss this functionality in more detail.

Issue #331, opened 2025-01-02
Auto-create export directories if they don't exist


dogversioning:
**Problem Statement**
When using as part of an external process, having to manually set these up is kind of silly.

**Desired behavior**
If an export directory is not found, create it.

dogversioning:
resolved via #347

Issue #254, opened 2024-06-24
When counting powersets with a secondary ID field, values less than subject limit are counted


dogversioning:
This primarily affects conditions. If we are counting encounters, and limiting by subjects >=10, we are getting encounter values that are either <10, or 0, in the resultant output. 

it's unlcear why this is happening (more than one patient per encounter? issue with the count query?) but we should address this either by changing the query logic or restricting it to suppress secondary ID counts <10.

dogversioning:
resolved via #347

PR #347, opened 2025-02-07
Counts template updates & misc bugfixes

dogversioning:
This PR is a catch all of technical debt & bugfixes that happened while we were pushing towards a demo.

- Updates to counts generation to prevent small bin sizes when counting a secondary resource #254 
- Auto creation of export directories #331 
- Updated the default aggregator URL to point to the current production location
- Adjusted some test data for uploads to match the new naming conventions (`*.meta.*`,`*.cube.*`)
- Alphasorted enums
- Fixed cli arg help prompt to match the expected actual usage
- Updated some tests away from using the deprecated `study_prefix` to mocking out manifest files
- Made valueset tests use tmpdir to prevent race conditions when parallelising test runs
- Made some progress towards incrementally adjusted tech debt:
  - Moving more files from object import to module import
  - Moving some expected console output from vanilla print to rich console (for future CI checking of errant print commits)

Additionally, there was work done, and then commented out pending an architecture discussion about the implications
- Trying to coerce study names for the dashboard's FHIR resource discovery algorithm #339 

### Checklist


cumulus_library/actions/uploader.py
```diff
@@ -5,22 +5,23 @@
...
+    file_upload_progress: progress.TaskID,
     file_path: Path,
     version: str,
     args: dict,
 ):
     """Fetches presigned url and uploads file to aggregator"""
     study = file_path.parts[-2]
     file_name = file_path.parts[-1]
-    progress.update(file_upload_progress, description=f"Uploading {study}/{file_name}")
+    c = console.Console()
```
mikix, line 23: Is there a meaningful difference between creating your own console object vs the singleton console object from `rich.get_console()`? (here and elsewhere)

cumulus_library/actions/uploader.py
```diff
@@ -79,13 +80,15 @@ def upload_files(args: dict):
     if not args["user"] or not args["id"]:
         sys.exit("user/id not provided, please pass --user and --id")
     try:
-        meta_version = next(filter(lambda x: str(x).endswith("__meta_version.parquet"), file_paths))
+        meta_version = next(
+            filter(lambda x: str(x).endswith("__meta_version.meta.parquet"), file_paths)
+        )
```
mikix, line 85: Do we want to look for both filenames? To continue to support the old style too?

mikix (approved):

cumulus_library/actions/uploader.py
dogversioning: So - this is mostly a byproduct of rich's documentation - if you're looking at the console API docs, they don't mention this other method existing, so I was unaware of it until now.

The actual functional difference is that the `get_console` method produces a global instance, whereas the rich.console.Console() approach gets a local one. If we cared about style bleed, this might be an issue; I don't believe that we do.

I'm happy to standardize on either style. if it's get_console, i'd take that update as a followon issue (and probably bite the bullet on doing it all in one go).

cumulus_library/actions/uploader.py
dogversioning: I don't think so - this was part of the v4 breaking change, any dangling issues related to this should be considered bugs.

cumulus_library/actions/uploader.py
dogversioning: based on our offline chat - #350

PR #349, opened 2025-02-11
Make discovery slightly more performant with UNION ALL

mikix:
Also, fix a typo that prevented discovery from finding code definitions for AllergyIntolerance.verificationStatus.

### Checklist


cumulus_library/studies/discovery/discovery_templates/code_system_pairs.sql.jinja
```diff
@@ -48,5 +48,5 @@ FROM (
     AS t (table_name, column_name, code, display, system)
 {%- endif %}
 
-{{ syntax.union_delineate(loop) }}
+{{ syntax.union_all_delineate(loop) }}
```
mikix, line 51: You'll need to look at the rest of this file to confirm this, but the tables being union'd already have unique table_name and column_name columns, and are de-duplicated themselves with SELECT DISTINCT. So we don't need the de-duplication of UNION here, I don't believe.

dogversioning (approved):

Issue #354, opened 2025-02-20
When a query fails due to missing columns, suggest running `cumulus-etl init`


dogversioning:
**Problem Statement**
People are missing running this command and then trying to debug behavior

**Desired behavior**
Suggest that initializing the schemas may help

PR #356, opened 2025-02-26
Suggest cases where a non-initialized DB may cause errors

dogversioning:
In cases where an error :might: be related to a database not being initialized by the etl, we'll point out the documentation. Addresses #354 

### Checklist

### DIFF
--- a/cumulus_library/actions/builder.py
+++ b/cumulus_library/actions/builder.py
@@ -459,7 +459,14 @@ def _query_error(
     rich.print(query_and_filename[0], file=sys.stderr)
     rich.print("--------", file=sys.stderr)
     log_utils.log_transaction(config=config, manifest=manifest, status=enums.LogStatuses.ERROR)
-    sys.exit(exit_message)
+    rich.print(exit_message)
+
+    if any(init_error in exit_message for init_error in config.db.init_errors()):
+        rich.print(
+            "Have you initialized your database?\n"
+            "https://docs.smarthealthit.org/cumulus/etl/setup/initialization.html"
+        )
+    sys.exit()
 
 
 def _execute_build_queries(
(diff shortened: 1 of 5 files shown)


cumulus_library/actions/builder.py
```diff
@@ -459,7 +459,14 @@ def _query_error(
     rich.print(query_and_filename[0], file=sys.stderr)
     rich.print("--------", file=sys.stderr)
     log_utils.log_transaction(config=config, manifest=manifest, status=enums.LogStatuses.ERROR)
-    sys.exit(exit_message)
+    rich.print(exit_message)
+
+    if any(init_error in exit_message for init_error in config.db.init_errors()):
+        rich.print(
+            "Have you initialized your database?\n"
+            "https://docs.smarthealthit.org/cumulus/etl/setup/initialization.html"
```
mikix, line 467: In our docs about our docs: https://github.com/smart-on-fhir/cumulus/blob/main/CONTRIBUTING.md
We suggest (well I suggest since I think I wrote that) that cross-repo doc links should only link to the top level, to reduce chances of having broken links when one repo reorganizes things.

I buy that in general, but I know it’s helpful here… is it bad to point at the top level and say “look at the ETL docs for initializing” or something without directly pointing at it? Or do you want to ignore the guidance and deal with the risk of link rot?

dogversioning: I think in this case I do want to take the risk in part because we've already seen instances of people not navigating the docs well. I do think that the changes in the structure make it a little easier to find this information, so you could talk me out of this, but I kind of just want to hit people over the head with the specific information we're asking them to read.

mikix (approved):
Meant to approve, sorry

PR #357, opened 2025-03-03
Switch to DB native export format

dogversioning:
This changes exports to use whatever the db-appropriate method of exporting while preserving types is, as opposed to trying to cast columns based on an observed type converted to a parquet value (which can be a little lossy).

### Checklist


cumulus_library/databases/athena.py
```diff
@@ -168,6 +141,46 @@ def upload_file(
...
+        )
+        if "Contents" in res:
+            for file in res["Contents"]:
+                s3_client.delete_object(Bucket=bucket, Key=file["Key"])
+
+        self.connection.cursor().execute(f"""UNLOAD
+            (SELECT * FROM {table_name})
+            TO '{s3_path}'
+            WITH (format='PARQUET', compression='SNAPPY')
+            """)  # noqa: S608
```
dogversioning, line 168: I could? create a centralized jinja template for these DB specific queires. I don't know how much this export mechanism needs the injection protection, since it's not in queries that are being distributed.

cumulus_library/actions/exporter.py
```diff
@@ -24,32 +24,23 @@ def reset_counts_exports(
...
-
 def export_study(
     config: base_utils.StudyConfig,
     manifest: study_manifest.StudyManifest,
     *,
     data_path: pathlib.Path,
     archive: bool,
     chunksize: int = 1000000,
-) -> list:
+):
```
mikix, line 34: `-> None` would be more accurate yeah?

cumulus_library/actions/exporter.py
```diff
@@ -64,34 +55,27 @@ def export_study(
...
+                index=False,
+            )
+        else:
+            skipped_tables.append(table.name)
+
+    if len(skipped_tables) > 0:
+        c = console.Console()
+        c.print("The following tables were empty and were not exported:")
+        for table in skipped_tables:
+            c.print(table)
```
mikix, line 79: nit: I thought we were doing rich.print

Also, maybe do `print(f"  {table}")` or something to give a touch of indent. Or even add a hyphen prefix. Something to make the list of tables more closely associated with the comment above.

cumulus_library/databases/duckdb.py
```diff
@@ -207,6 +182,28 @@ def operational_errors(self) -> tuple[type[Exception], ...]:
             duckdb.BinderException,
         )
 
+    def export_table_as_parquet(
+        self, table_name: str, table_type: str, location: pathlib.Path, *args, **kwargs
+    ) -> str | None:
```
mikix, line 187: The hint here and in athena.py are for `str` but should be `pathlib.Path`

cumulus_library/databases/athena.py
```diff
@@ -168,6 +141,46 @@ def upload_file(
             )
         return f"s3://{bucket}/{s3_key}"
 
+    def export_table_as_parquet(
+        self, table_name: str, table_type: str, location: pathlib.Path, *args, **kwargs
+    ) -> str | None:
+        s3_client = boto3.client("s3")
+        output_path = location / f"{table_name}.parquet"
```
mikix, line 148: This line gets overridden later.

cumulus_library/databases/base.py
```diff
@@ -208,6 +205,16 @@ def upload_file(
         have an API for file upload (i.e. cloud databases)"""
         return None
 
+    @abc.abstractmethod
+    def export_table_as_parquet(
+        self, table_name: str, table_type: str, location: pathlib.Path, *args, **kwargs
```
mikix, line 210: nit: `table_type` seems like an odd addition here. In both implementations, it's just used to create the output file name, which felt like a bit of duplicated business logic.

What if you used a "input/output, did it write anything" pattern like (totally ignorable suggestion, just brainstorming):
```
def export_table_as_parquet(self, table_name: str, output_path: pathlib.Path) -> bool:
```

cumulus_library/databases/duckdb.py
```diff
@@ -207,6 +182,28 @@ def operational_errors(self) -> tuple[type[Exception], ...]:
...
+        parquet_path = location / f"{table_name}.{table_type}.parquet"
+        parquet_path.parent.mkdir(exist_ok=True, parents=True)
```
mikix, line 189: The caller already did this yeah? - Maybe add a comment in the docstring for the abstract method about what the assumption about the dir is.

cumulus_library/databases/duckdb.py
```diff
@@ -207,6 +182,28 @@ def operational_errors(self) -> tuple[type[Exception], ...]:
...
+            (SELECT * FROM {table_name})
+            TO '{parquet_path}'
+            (FORMAT parquet)
+            """  # noqa: S608
+        self.connection.execute(query)
+
+        df = pandas.read_parquet(parquet_path)
+        df = df.sort_values(
+            by=list(df.columns), ascending=False, ignore_index=True, na_position="first"
+        )
```
mikix, line 203: nit: I think you could do this in the SQL above: `ORDER BY ALL DESC NULLS FIRST`

cumulus_library/databases/athena.py
mikix: The table name comes from the manifest yeah? So it is user input, which does mean it's subject to chicanery, by the user or study author or malicious 3rd party app modifying files in $HOME... (btw: do we have any sanitizing of table names when reading the manifest?)

But I'm not overly stressed - I'll leave it up to you on risk assessment here.

cumulus_library/databases/athena.py
```diff
@@ -168,6 +141,46 @@ def upload_file(
...
+        workgroup = self.connection._client.get_work_group(WorkGroup=self.work_group)
+        wg_conf = workgroup["WorkGroup"]["Configuration"]["ResultConfiguration"]
+        s3_path = wg_conf["OutputLocation"]
+        bucket = "/".join(s3_path.split("/")[2:3])
+        output_path = location / f"{table_name}.{table_type}.parquet"
+        s3_path = f"s3://{bucket}/export/{table_name}.{table_type}.parquet"
+
+        # Cleanup location in case there was an error of some kind
```
mikix, line 156: It's annoying that UNLOAD doesn't have a flag to do this for you.

cumulus_library/databases/athena.py
```diff
@@ -168,6 +141,46 @@ def upload_file(
...
+            df = awswrangler.s3.read_parquet(s3_path)
+        except awswrangler.exceptions.NoFilesFound:
+            return None
+        df = df.sort_values(by=list(df.columns), ascending=False, na_position="first")
+        df.to_parquet(output_path)
+        res = s3_client.list_objects_v2(
+            Bucket=bucket, Prefix=f"export/{table_name}.{table_type}.parquet"
+        )
+        for file in res["Contents"]:
+            s3_client.delete_object(Bucket=bucket, Key=file["Key"])
```
mikix, line 181: This bit is identical to above - maybe worth a tiny helper method.

tests/test_data/duckdb_data/expected_export/core/core__count_allergyintolerance_month.cube.csv
```diff
@@ -1,3 +1,3 @@
-"cnt","category","recordedDate_month","code_display","reaction_manifestation_display"
+cnt,category,recordedDate_month,code_display,reaction_manifestation_display
 17,,,,
-16,,"2018-08-01",,
+16,,2018-08-01,,
```
mikix, line 3: I know that Vlad's dashboard sometimes makes assumptions about the quoting scheme in use. Have you tested whether this scheme works correctly? You might have to add some flags to the csv exporter to keep the same quoting behavior if not.

mikix (approved):

tests/test_data/duckdb_data/expected_export/core/core__count_allergyintolerance_month.cube.csv
dogversioning: Confirmed this format works.

cumulus_library/databases/duckdb.py
mikix: If you don't want to have optimized-per-db versions of this ordering, maybe bring it up out of the db-specific code, so it's not duplicated?

cumulus_library/databases/athena.py
dogversioning: but... on the other hand, a user can run a DROP TABLE query if they edit the manifest and it would be valid, so... maybe we're just not safe from a malicious user.

mikix: Or a malicious script on the user's machine

cumulus_library/databases/base.py
dogversioning: ok, as discussed offline, this now takes a `filename` arg and returns a bool

PR #359, opened 2025-03-05
Added example IAM permissions for export

dogversioning:
This adds a line to the documentation describing the required permissions to export data from Athena.

### Checklist

### DIFF
--- a/docs/sharing-data.md
+++ b/docs/sharing-data.md
@@ -26,6 +26,24 @@ in a `CUMULUS_LIBRARY_DATA_PATH` variable. The library will create subfolders in
 this directory based on the names studies you are exporting. These will be emptied
 before each export run.
 
+If you are exporting data from Amazon Athena, it is assumed that you will have permissions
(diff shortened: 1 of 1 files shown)


mikix (approved):
Does this need to be added to the aws template we offer in the ETL docs?

dogversioning:
> Does this need to be added to the aws template we offer in the ETL docs?

Hmm... probably, yeah.

dogversioning:
> > Does this need to be added to the aws template we offer in the ETL docs?
> 
> Hmm... probably, yeah.

Though - we're not really dealing with bucket permissions in a prod-y sense with that, right? it's sort of a 'here's a starting point and customize permissions to fit your use case" kind of situation?

Anyway - we can chat about that outside of this context.

mikix:
Yeah looking at the template again, we don't really get into access control: https://docs.smarthealthit.org/cumulus/etl/setup/cumulus-aws-template.yaml

PR #363, opened 2025-04-07
Added network arg to upload command

dogversioning:
This allows a user to specify a network as part of an aggregator upload target.

### Checklist

### DIFF
--- a/cumulus_library/actions/uploader.py
+++ b/cumulus_library/actions/uploader.py
@@ -23,6 +23,9 @@ def upload_data(
     c = rich.get_console()
     progress_bar.update(file_upload_progress, description=f"Uploading {study}/{file_name}")
     data_package = file_name.split(".")[0]
+    url = args["url"]
+    if args["network"]:
+        url += args["network"]
     prefetch_res = requests.post(
         args["url"],
         json={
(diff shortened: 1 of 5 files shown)


cumulus_library/actions/uploader.py
```diff
@@ -23,6 +23,9 @@ def upload_data(
     c = rich.get_console()
     progress_bar.update(file_upload_progress, description=f"Uploading {study}/{file_name}")
     data_package = file_name.split(".")[0]
+    url = args["url"]
+    if args["network"]:
+        url += args["network"]
```
mikix, line 28: Just a simple concat? Why not have the caller modify the URL arg instead? Or do we use the base URL sometimes by itself and sometimes concat'ed?

mikix (approved):

cumulus_library/actions/uploader.py
dogversioning: this is mostly for us, since we will be using the same base url, but different network names, to upload data to various aggregators. But anyone who is participating in more than one network would have a similar use case. it felt like less typing to me; you can talk me into 'just replace the url'.

mikix: I don't feel strongly.

Separately, I see that the current concatenation also relies on the user putting a slash at the end of the url - maybe use a joiner method that allows for some sloppiness?

dogversioning: yeah, i thought about that, two people thinking of that probably means do it

PR #368, opened 2025-04-17
Ignore customer specific Epic OID

dogversioning:
This forces `discovery` to exclude all the potential PHI data in the customer specific OID EPIC uses for individual deployments.

### Checklist

### DIFF
--- a/cumulus_library/__init__.py
+++ b/cumulus_library/__init__.py
@@ -17,4 +17,4 @@ from cumulus_library.template_sql.base_templates import get_template
 # is all code level. See template_sql for more information.
 
 __all__ = ["BaseTableBuilder", "CountsBuilder", "StudyConfig", "StudyManifest", "get_template"]
-__version__ = "4.5.0"
+__version__ = "4.5.1"
--- a/cumulus_library/studies/discovery/discovery_templates/code_system_pairs.sql.jinja
+++ b/cumulus_library/studies/discovery/discovery_templates/code_system_pairs.sql.jinja
@@ -36,6 +36,7 @@ UNNEST(col_{{ index }}.{{ table.column_hierarchy[index][0].split('.')[0] }}) as 
 {%- elif table.column_hierarchy[index][1].__name__ == 'dict' and index +1 == table.column_hierarchy|length -%},
 UNNEST(col_{{ index }}.{{ table.column_hierarchy[index][0].split('.')[0] }}.coding) as table_{{ index +1 }} (col_{{ index +1 }})
 {%- endif %}
+WHERE REGEXP_LIKE({{ get_source_col(table.column_hierarchy) }}.code, '^((?!1.2.840.1.114350).)*$')
 {%- endfor %}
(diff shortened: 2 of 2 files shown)


cumulus_library/studies/discovery/discovery_templates/code_system_pairs.sql.jinja
```diff
@@ -36,6 +36,7 @@ UNNEST(col_{{ index }}.{{ table.column_hierarchy[index][0].split('.')[0] }}) as
 {%- elif table.column_hierarchy[index][1].__name__ == 'dict' and index +1 == table.column_hierarchy|length -%},
 UNNEST(col_{{ index }}.{{ table.column_hierarchy[index][0].split('.')[0] }}.coding) as table_{{ index +1 }} (col_{{ index +1 }})
 {%- endif %}
+WHERE REGEXP_LIKE({{ get_source_col(table.column_hierarchy) }}.code, '^((?!1.2.840.1.114350).)*$')
```
mikix, line 39: What about this? (code=>system, add urn:oid:, escape dots, add a last dot into the look-behind, move the line start into the look-behind). Have not tested

```
WHERE REGEXP_LIKE({{ get_source_col(table.column_hierarchy) }}.system, '(?!^urn:oid:1\.2\.840\.1\.114350\.).*$')
```

cumulus_library/__init__.py
```diff
@@ -17,4 +17,4 @@
 # is all code level. See template_sql for more information.
 
 __all__ = ["BaseTableBuilder", "CountsBuilder", "StudyConfig", "StudyManifest", "get_template"]
-__version__ = "4.5.0"
+__version__ = "4.5.1"
```
mikix, line 20: We're only removing this from the `discovery` study - I know that modifying `core` would be a much more invasive problem, but just noting that these codes would exist there too.

cumulus_library/studies/discovery/discovery_templates/code_system_pairs.sql.jinja
mikix: Maybe worth adding a unit test

dogversioning: i would, but i was trying to get this out before 11am and worry about the testing after

cumulus_library/__init__.py
dogversioning: core handles this by allowlisting codes in places we care about them - i'm less worried about that

mikix: I know we allow list in a couple places, but there's lots of places and resources we just denormalize codeable concepts and toss in there. I think Observation codes for one.

mikix: (I'm not saying we necessarily have to fix this issue there - just noting it. The official recommendation can just be "rerun your ETL")

dogversioning:
closing as wont do due to deciding on ETL ownership

PR #370, opened 2025-05-13
Restrict upload to require meta_date

dogversioning:
- Enforces the presence of a `meta_date` table in an export before allowing an upload
- Removes the 'upload everything' mode in favor of requiring one or more targets

### Checklist


cumulus_library/cli.py
```diff
@@ -291,7 +291,7 @@ def run_cli(args: dict):
     if args["action"] == "upload":
         try:
             uploader.upload_files(args)
-        except requests.RequestException as e:
+        except requests.RequestException as e:  # pragma: no cover
```
mikix, line 294: Why not test this path? An error during upload feels like a reasonable test to have.

mikix (approved):

cumulus_library/cli.py
dogversioning: you found my 'why did this break' frustration! i'll figure it out

PR #380, opened 2025-07-01
Add code to condition, annotated cubes

dogversioning:
This PR does the following:
- Adds a new field to the count condition table
- Adds support for a new data type, `annotated_cube`, which allows for gluing metadata to powerset rows

### Checklist


cumulus_library/builders/counts.py
```diff
@@ -126,16 +81,15 @@ def get_count_query(
             raise errors.CountsBuilderError(
                 f"count_query missing required arguments. output table: {table_name}"
             )
-        ### TODO: removing this for now pending architectural discussions
```
dogversioning, line 129: We talked about this already - I just never got around to the deletion

docs/api.md
mikix, line 1: :scissors: ?

docs/creating-studies.md
```diff
@@ -85,6 +85,17 @@ count_list = [
     "my_study__count_influenza_test_month",
 ]
 
+# In some cases, you may want to annotate count data with metadata from another
+# source. Most commonly, this is used to add data related to a code (like its
+# display value or code system) from another source. These tables should go
+# in this section.
```
mikix, line 91: So how do you specify the source of metadata? This is pretty "magic" so far yeah? Like this just changes the file suffix, and the aggregator knows to do a special thing for certain well-known tables?

Second question: nothing in core makes and uses a CountAnnotation() yet, right?

mikix (approved):

docs/creating-studies.md
dogversioning: The CountAnnotation specifies the metadata source. I was originally going to point to the API docs to describe this better, and that's still my intent, but it doesn't block using it for the short term deliverable.

Nothing in Core uses it; I'm testing this in catalog currently, which is the next PR down the list. All we're exercising in core today is whether the query is generated.

mikix: :+1:

PR #381, opened 2025-07-02
API docs

dogversioning:

- A high-ish level API doc of the stuff in the public interface
- Some tweaks to docstrings as a result of writing that API doc
- Updates the blank CLI output to point more explicitly to how to get more information

### Checklist

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1,0 +1,312 @@
+---
+title: Clinical study API reference
+parent: Library
+nav_order: 10
+# audience: Clinical researchers/engineers trying to create a study
+# type: reference
+---
+# Package API reference
+{: .no_toc }
+
+The following describes at a high level some of the classes available to help
+you build studies programmatically. They are all importable directly from
+the base `cumulus_library` package.
+
+This is aimed at clinical study usage, rather than data prep usage,
+so some of the fussier bits will be glossed over
+
+- TOC
+{:toc}
+
+## BaseTableBuilder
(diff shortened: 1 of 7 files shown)


docs/api.md
```diff
@@ -0,0 +1,312 @@
+---
```
mikix, line 1: This level of detail makes me worry about drift between the docs and the code, but if we want API docs, that's a little unavoidable, unless we generate it all from docstrings - which, may be worth putting a pin in.

mikix (approved):

docs/api.md
dogversioning: So - I originally went down the route of 'Just generate from docstrings', and I walked back from it, because I think we have a slightly different target here - This isn't aimed at like somone working with raw FHIR or doing complex data wrangling, this is 'just lemme extract some data from core', so I was able to kind of skip over some lower level stuff and focus on a few targeted functions. There's arguably a different class of user who could benefit from a generate from docstring type use case - I don't see one of those in the short term, but I'm open to doing it later (or, for example, doing it for all the utility methods for accessing templates).

Totally willing to change tack, though. I'm gonna add a note to the PR checklist for maintaining this.

mikix: This is fine for now regardless. But I'm not sure I see the same issue you do. Worth chatting about over lunch.

We're used to writing docstrings that are for really low level usage, but it doesn't have to be so techy. And you could certainly have docstrings on an outer API layer separate from guts-of-code docstrings.

PR #389, opened 2025-07-22
Added alternate targets to CountsAnnotation

dogversioning:
This update to CountsAnnotation allows you to change the value by which to count to a column from the metadata table. This is a little obtuse, so here's an example:

Counted table:
|  A |  B | count|
|---|---|---|
|a|x|1|
|a|y|2|
|b|z|3|
|b|zz|4|

metadata table:
|field|label|
|----|-----|
|a|foo|
|b|bar|

Using `A=field` as the joined column, no other columns from the counted table with `label` as the alternate target, you'd get the following:
|field|label|count|
|----|-----|------|
|a|foo|3|
|b|bar|7|

### Checklist

### DIFF
--- a/cumulus_library/builders/statistics_templates/counts_templates.py
+++ b/cumulus_library/builders/statistics_templates/counts_templates.py
@@ -46,6 +46,8 @@ class CountAnnotation:
     :keyword join_field: the column in the table to join on
     :keyword list: a list of tuples like ('column_name', 'alias' or None) to
         define columns to add to your dataset.
+    :keyword alt_target: if present, a column to use from the annotation table
+        instead of the one defined in `field` as the primary countable target
 
     """
 
@@ -53,6 +55,7 @@ class CountAnnotation:
     join_table: str
     join_field: str
     columns: list[tuple[str, str | None]]
+    alt_target: str | None = None
 
 
 def get_count_query(
@@ -81,7 +84,6 @@ def get_count_query(
         else:
             table_col_classed.append(CountColumn(name=item, db_type="varchar", alias=None))
     table_cols = table_col_classed
-
     query = base_templates.get_template(
         "count",
         path,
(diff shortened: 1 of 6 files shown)


tests/test_counts_templates.py
```diff
@@ -282,6 +282,90 @@
...
+    FROM powerset AS p
+    JOIN secondary_powerset AS s on s.id = p.id
+        JOIN other_table j ON p.code = j.other_field
+    WHERE 
+        p.cnt_subject_ref >= None
+        AND s.cnt_encounter_ref >= None
+    GROUP BY
+        j."A",
+        j."B"
+    ORDER BY A ASC
```
mikix, line 353: Relying on this kind of test pattern (template string checking) makes me lightly nervous, down the road. I'm thinking of some non-Matt engineer making changes to the template and it causes this template not to match.

How well will they understand the *point* of this test - which piece of this template is supposed to be changed due to which inputs? Vs just thinking... "well I did make changes to this template, so of course this test fails now, let's just update it", even if their change affects this new code path and they don't understand the implications.

An alternative test in my head would be something that sets up some fake data and checks the resulting created table for the correct columns and values. (Both test styles could exist at same time - my concern is only relying on a template string check).

Not a blocker, just thinking out loud.

cumulus_library/builders/statistics_templates/counts_templates.py
```diff
@@ -46,13 +46,16 @@ class CountAnnotation:
     :keyword join_field: the column in the table to join on
     :keyword list: a list of tuples like ('column_name', 'alias' or None) to
         define columns to add to your dataset.
+    :keyword alt_target: if present, a column to use from the annotation table
+        instead of the one defined in `field` as the primary countable target
```
mikix, line 50: Is this description accurate? If this is defined, *all* columns from the annotation become the primary countable target. The result is merely _sorted_ by `alt_target`.

PR #394, opened 2025-08-28
Add example_nlp builtin study

mikix:
This tiny study just supports our documentation, which has a sample walkthrough of our NLP flow. This study takes some NLP results and post-processes them into some labels, which the ETL can then upload to Label Studio.

I also fixed a typo in the DocRef.category system filter.

### Checklist


cumulus_library/actions/builder.py
```diff
@@ -437,7 +437,7 @@ def _check_if_preparable(prefix):
     # This list should include any study which requires interrogating the database to
     # find if data is available to query (outside of a toml-driven workflow),
     # which isn't doable as a distributed query
-    if prefix in ["core", "discovery", "data-metrics"]:
+    if prefix in ["core", "discovery", "data_metrics", "example_nlp"]:
```
mikix, line 440: This bit right here is interesting to me. How important is prepare mode? I'm imagining most studies that do post-processing on NLP results are gonna want to do a similar thing - only run that part of the study if the NLP results are sitting there. So this list will grow, likely.

dogversioning: We talked about this offline - https://github.com/smart-on-fhir/cumulus-library/issues/395 covers the details about why we need to be concerned about prepare mode here.

dogversioning (approved):

PR #403, opened 2025-09-17
core: add Location, Organization, Practitioner, and PractitionerRole

mikix:
Just very basic support - add tables for them, but not count results.

### Checklist


docs/core-study-details.md
```diff
@@ -96,12 +96,23 @@ Per resource, the optional fields are as follows:
 - Encounter
```
mikix, line 96: This file doesn't have any count table updates, because I didn't add count tables for the new resources - discuss? Are these resources worth having new count tables for?

PR #408, opened 2025-10-01
core: add links from old resources to new aux resources

mikix:
Stats about how much we'll explode the (BCH Epic) core tables below. The only real pop is DxReports.

docrefs
new: 53,357,324
old: 53,185,185

dxreports
new: 970,176,912
old: 477,748,599

encounters
new: 17,959,420
old: 16,368,586

observations
new: 617,566,245
old: 617,043,803

### Checklist


cumulus_library/studies/core/core_templates/diagnosticreport.sql.jinja
```diff
@@ -4,8 +4,6 @@
 -- This table includes all fields of interest to the US Core DiagnosticReport profiles.
 -- EXCEPT FOR:
 -- * the 'presentedForm' field, which is an attachment array that is stripped out by the ETL.
--- * the `reporter` field, simply due to it not likely being interesting to consumers
---   and being an array field, which would require a lot of row duplication.
```
mikix, line 8: This line should have said `performer` not `reporter` - and the math on interest-to-consumers has changed now that we have actual practitioners in the database. It still requires a lot of row duplication, but...

PR #416, opened 2025-10-28
Swapping out PSM for matplotlib-less version

dogversioning:
This PR moves a subseciton of the psmpy code into the repo, allowing us to drop the matplotlib dependency, and replaces graphs with a dump of data outputs.

### Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,7 +56,7 @@ build-backend = "flit_core.buildapi"
 
 [tool.coverage.run]
 omit = [
-    "builders/psmpy_lite.py"
+    "cumulus_library/builders/psmpy_lite.py"
 ]
 
 [tool.flit.sdist]


cumulus_library/builders/psmpy_lite.py
```diff
@@ -0,0 +1,593 @@
...
+        df_preds_b4 = self.data[[self.treatment, *self.xvars]]
+        df_preds_after_float = df_preds_after.astype(float)
+        df_preds_b4_float = df_preds_b4.astype(float)
+
+        data = []
+        for cl in self.xvars:
+            data.append([cl, "before", self.cohenD(df_preds_b4_float, self.treatment, cl)])
+            data.append([cl, "after", self.cohenD(df_preds_after_float, self.treatment, cl)])
+        self.effect_size = pandas.DataFrame(data, columns=["Variable", "matching", "Effect Size"])
+        return self.effect_size
```
dogversioning: these are the only new functions in this file

cumulus_library/builders/psmpy_lite.py
dogversioning: ok, so specifically the kinds of changes for this file from the version at https://github.com/adriennekline/psmpy:
- Ruff reformatting/ removal of bare exceptions
- A (failing) attempt to patch some of the random behavior of LogisticRegression
- trimming one or two obvious errors

But otherwise, this is as found, and I'm explicitly not trying to guarantee full coverage of this file/update this code
to match our internal house style, to make it easier to graft in updates if we need to.

cumulus_library/builders/psmpy_lite.py
```diff
@@ -0,0 +1,593 @@
+# The following code is adapted from https://github.com/adriennekline/psmpy,
+# less the graphing elements and with some light production-y adaptations.
+# Most of the commenting is from the existing library, the get_* functions
+# are returning datasets that would be processed for charts in psmpy.
```
mikix: I might add a chunk here noting that this is licensed under MIT and give the copyright holder's name (Adrienne Kline) - I know you link to the repo, but it's nice to have it local too - who knows when that repo will shutter for some reason.

pyproject.toml
```diff
@@ -53,6 +54,11 @@ cumulus-library = "cumulus_library.cli:main_cli"
 requires = ["flit_core >=3.4,<4"]
 build-backend = "flit_core.buildapi"
 
+[tool.coverage.run]
+omit = [
+    "builders/psmpy_lite.py"
+]
+
```
mikix: Nice - I was wondering

cumulus_library/builders/psmpy_lite.py
dogversioning: Yeah good call on the license.

I'm not sure about storing the repo in the project - we could keep a fork in our org?

PR #417, opened 2025-10-29
Added db parallel writes, updated clean action to be parallel

dogversioning:
This PR implements the first portion of parallel execution of queries by adding a `parallel_write` method to the database classes. Since a clean option is always all or nothing, it also flips clean to use this as the primary write method.

I'm doing this separate from updating manifest parsing/BaseTableBuilder updates to make this a little more reasonable, and to get the parallel stuff finalized before dealing with all the repercussions.

The means for doing concurrent executions are driven by the recommendations in each of the relevant library's docs. Might be easier to chat about those if there are any questions.

### Checklist

### DIFF
--- a/cumulus_library/databases/duckdb.py
+++ b/cumulus_library/databases/duckdb.py
@@ -12,6 +12,7 @@ import datetime
 import pathlib
 import re
 import threading
+import time
 
 import duckdb
 import pandas
@@ -224,6 +225,10 @@ class DuckDatabaseBackend(base.DatabaseBackend):
             t.start()
         for t in threads:
             t.join()
+        # This is a temporary workaround for cases where duckdb is still doing :something:
+        # after the threads have resolved, where if you try to recreate a table it may throw
+        # an error, which is only affecting the --statistics flag ¯\_(ツ)_/¯
+        time.sleep(0.25)
 
     def create_schema(self, schema_name):
         """Creates a new schema object inside the database"""


cumulus_library/databases/duckdb.py
```diff
@@ -197,6 +201,35 @@ def export_table_as_parquet(
...
+                )
+            )
+        for t in threads:
+            t.start()
+        for t in threads:
+            t.join()
+        # This is a temporary workaround for cases where duckdb is still doing :something:
+        # after the threads have resolved, where if you try to recreate a table it may throw
+        # an error, which is only affecting the --statistics flag ¯\_(ツ)_/¯
+        time.sleep(0.25)
```
mikix: We discussed this already - no further comment, except bummer

mikix (approved):

cumulus_library/databases/duckdb.py
dogversioning: it's occasionally happening in the other PR without the async cursor too - i think there's something else going on but i might want to work on that in isolation. I'll make an issue for it.

cumulus_library/databases/athena.py
```diff
@@ -182,6 +188,25 @@ def export_table_as_parquet(
...
+    def parallel_write(
+        self,
+        queries: list[str],
+        verbose: bool,
+        progress_bar: progress.Progress,
+        task: progress.Task,
+    ) -> None:
+        def query_completed(f: futures.Future):
+            with base_utils.query_console_output(verbose, query, progress_bar, task):
+                pass
```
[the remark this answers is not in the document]
dogversioning: yeah these are on the main thread, as far as I understand.

i could make like... the current content manager a wrapper function? and then move the updating code to a new function, which we could call here? but then again maybe there will be a point where we want to do something as these things resolved, so maybe it's ok?

PR #421, opened 2025-11-17
Add parallel builds for studies

dogversioning:
This PR allows studies to be configured to run queries in parallel, rather than just one after another, and deals with the fallout of allowing both run modes to occur, primarily by allowing both implicit and explicit fallback to serial mode.

Maybe start with the doc updates first and then scan through the rest of it?

### Checklist

### DIFF
--- a/cumulus_library/actions/builder.py
+++ b/cumulus_library/actions/builder.py
@@ -244,7 +244,7 @@ def _execute_build_queries(
             with base_utils.query_console_output(config.verbose, query, progress, task):
                 cursor.execute(query)
 
-        except Exception as e:  # pylint: disable=broad-exception-caught pragma: no cover
+        except Exception as e:  # pragma: no cover
             _query_error(
                 config,
                 manifest,
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -425,6 +425,21 @@ def test_clean(tmp_path, args, expected, raises):
             [
                 "build",
                 "-t",
+                "study_valid_parallel",
+                "-s",
+                "tests/test_data/study_valid_parallel/",
+                "--continue",
+                "test4",
+            ],
+            ["export", "-t", "study_valid_parallel", "-s", "tests/test_data/study_valid_parallel/"],
+            4,
+            pytest.raises(errors.StudyManifestParsingError),
+            [],
+        ),
+        (
+            [
+                "build",
+                "-t",
                 "study_valid",
                 "-s",
                 "tests/test_data/study_valid/",


cumulus_library/actions/builder.py
dogversioning: noting for posterity: i found the organization of this file confusing while I was working on it, so I'm going to reorder the functions into blocks after review. But there's enough churn in here that I though it would just hide the changes to do that out of the gate.

tests/test_cli.py
```diff
@@ -498,7 +502,7 @@ def test_cli_executes_queries(
             .execute("SELECT table_schema,table_name FROM information_schema.tables")
             .fetchall()
         )
-        assert len(found_tables) == expected_tables
+        assert len(found_tables) == expected_tables + FHIR_RESOURCE_TABLE_COUNT
```
dogversioning: another materialization byproduct - before the core tables would be dropped when we got a new cursor for validating results.

cumulus_library/actions/builder.py
```diff
@@ -362,41 +442,50 @@ def _run_raw_queries(
...
+            manifest.get_study_prefix(),
+            cleaned_queries,
+            data_path,
+            filename,
+            query_count,
+            run_stage,
         )
-    else:
+        return cleaned_queries
+    if not parallel:
```
mikix: There's something lightly "too magic" to me that a method called _run_raw_queries doesn't actually do that if you are in parallel mode, it simply prepares them and returns them - in the parallel mode, you care about the return value, but in the serial mode you don't - feels like two different code flows where the function boundaries aren't quite right.

You don't need to refactor this all in response to my vibe here, just calling it out - I know you've got more refactoring of this file in your brain. Maybe you can disentangle these a bit more later, or maybe this is perfect actually.  Naively, it feels like both flows could share most of the "prepare sql queries" code - just the final "run in parallel or serial" step needs to fork. But then you have to stop executing-as-you-go, which maybe has some conveniences.

tests/conftest.py
```diff
@@ -141,6 +141,10 @@ def update_nested_obj(id_path, obj, i):
             for null_bool_col in ["multipleBirthBoolean"]:
                 if null_bool_col in output_df.columns:
                     output_df[null_bool_col] = output_df[null_bool_col].replace({0.0: False})
+            # workaround for arrow int/float casting issues
+            for int_col in ["multipleBirthInteger"]:
+                if int_col in output_df.columns:
+                    output_df[int_col] = output_df[int_col].round().astype("Int64")
```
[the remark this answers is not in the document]
mikix: It looks like the `select *` is not preserving some part of the schema? So it's coming in as a float... hmph. That could potentially cause issues in code that looks at int fields (like in data-metrics maybe?, I dunno). Looks like you already had to do a similar thing above for a boolean field? Is this like, a lurking problem we should solve?

PR #423, opened 2025-11-26
Better parallelization of the core study

dogversioning:
This PR adds the following:

- allows for passing a parameter to specify the number of threads for parallel jobs
- breaks up the core tables so that the majority of them can be run in parallel. Schema detection is still serial, that may be adressed in a followon update.

### Checklist

### DIFF
--- a/cumulus_library/cli_parser.py
+++ b/cumulus_library/cli_parser.py
@@ -84,7 +84,11 @@ def add_db_config(parser: argparse.ArgumentParser, input_mode: bool = False) -> 
     group.add_argument(
         "-c",
         "--max-concurrent",
-        help="Specifies the upper limit of queries allowed to run in parallel",
+        help=(
+            "Specifies the upper limit of queries allowed to run in parallel. "
+            "The default is database dependent, but will be 20 in most cases. "
+            "If you run into memory issues with larger studies, you can dial this back."
+        ),
         type=int,
     )
     if input_mode:


cumulus_library/studies/core/builder_allergyintolerance.py
dogversioning: for all of these, we're just pulling out the bits that do schema detection and moving them into a second file.

cumulus_library/cli_parser.py
```diff
@@ -81,7 +81,12 @@ def add_db_config(parser: argparse.ArgumentParser, input_mode: bool = False) ->
         #
         help="Database/schema name (for Athena) or schema (for DuckDB)",
     )
-
+    group.add_argument(
+        "-c",
+        "--max-concurrent",
+        help="Specifies the upper limit of queries allowed to run in parallel",
```
mikix: maybe mention the default? (and maybe use `default=20` instead of doing that downstream in code) - (though maybe you are leaving the possibility open that a future db backend might want a different default?)

But also for my own understanding - as a user, when/why do I reach for this argument? Like, what prompted you to add this? Like, if I want things to run faster, I guess, but at cost of memory?

dogversioning: yeah, the idea was that a different DB might have a different limit, which is why i pushed it down to the db class

as for when you want to use this - i had a couple thoughts:
- We can use it, to play around with speed - 20 is a guess of a number
- if you're in a tight environment (like git ci maybe? or a thin vm)? you might want to turn it down

this mostly came out of a conversation we had when i originally set it to 5. but maybe it remains a kind of dev-centric flag? and we use it to test dbs to find a reasonable default?

mikix: I don't have a strong opinion here, but you could also mark `help=argparse.SUPPRESS` to hide it from `--help` if you wanted to keep it as a secret dev thing.

But it's fine to leave open too.

Issue #425, opened 2025-12-01
crashing due to OOM with large ndjson datasets


dogversioning:
**Describe the bug**
the 'copy all registered ndjson to tables' approach seems to be causing OOM issues for large datasets

**To Reproduce**
1. download and extract the `2022-Q4` bch data
2. run any build command, i.e. cumulus-library build -t discovery --db-type duckdb --load-ndjson-dir ~/data/ --database ~/duck.db
3. Killed message recieved after about 16 minutes

**Expected behavior**
table creation should happen succesfully

this block testing on #348

Issue #348, opened 2025-02-10
Fix `discovery` study's memory use


mikix:
When running the `discovery` study on a m5-xlarge (16GB memory) on 65G of data (the `2022-Q4` bch data), Library gets OOM killed during the big `discovery__code_sources` union.

Thoughts:
- One easy improvement is switching from UNION to UNION ALL, since the tables don't have duplicates between them. But it's not enough by itself.
- We should keep the SELECT DISTINCT for each table, since that's actually doing work (though I assume it does take a performance hit)
  - I did test without DISTINCT though, and we still got OOM.
- Next thing to try: converting each intermediate SELECT into a `WITH table AS ()` temporary table - or actualize it as a real table if that's not enough, and then do a final UNION ALL between all the intermediate tables.

PR #426, opened 2025-12-10
DuckDB schema caching, better discovery mem usage

dogversioning:
- We remove the previous attempt to materialize tables in duckdb since it was making performance worse in a lot of cases
- We now cache the pyarrow datasets for loading ndjson in duckdb and retrieve them during parallel writes
- Discovery now writes out intermediate tables and unions in a seperate step, eliminating known large dataset crashes #425 #348
- futures now have their results checked to prevent silently ignoring failures during study builds

### Checklist

### DIFF
--- a/cumulus_library/studies/discovery/reference_sql/cleanup_tmp.sql
+++ b/cumulus_library/studies/discovery/reference_sql/cleanup_tmp.sql
@@ -3,3 +3,183 @@
 -- CLI. Its format is tied to the specific database it was run against,
 -- and it may not be correct for all databases. Use the CLI's build 
 -- option to derive the best SQL for your dataset.
+
+-- ###########################################################
+
+DROP table IF EXISTS discovery__tmp_allergyintolerance_clinicalstatus;
+
+-- ###########################################################
+
+DROP table IF EXISTS discovery__tmp_allergyintolerance_code;
+
+-- ###########################################################
+
+DROP table IF EXISTS discovery__tmp_allergyintolerance_reaction_exposureroute;
+
+-- ###########################################################
+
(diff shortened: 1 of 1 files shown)


cumulus_library/databases/duckdb.py
```diff
@@ -96,18 +111,89 @@ def connect(self):
...
+                if cached_files is not None:
+                    cached_files = json.loads(cached_files[0])
+                else:
+                    cached_files = []
+
+                fragment_infos = []
+                for fragment in fragments:
+                    fragment_modtime = pathlib.Path(fragment).lstat().st_mtime
+                    fragment_infos.append([fragment, fragment_modtime])
+                if not cached_files == fragment_infos:
```
mikix: nit: !=

tests/test_data/duckdb_data/README.md
```diff
@@ -2,3 +2,7 @@ This folder holds a small selection of results from Synthea.
 
 The selection is designed to be big enough to get past most of the 10-patient
 bucketing cutoffs, while small enough to not overwhelm the repo with fake patient data.
+
+If you change the data in this folder, don't forget to rerun update_pyarrow_cache.py -
+We use the cache to speed up DB initialization during unit tests. Tests won't fail if
+you forget, but the cache will get rebuilt during each test using the database.
```
mikix: Maybe you should make at least one test fail if this happens, as a forcing mechanism. I can totally imagine someone not reading this very thrice-nested README.

mikix (approved):

tests/test_data/duckdb_data/README.md
dogversioning: my expletive was related to byproducts of this. there's now a conftest hook which just runs this each time.

PR #429, opened 2025-12-16
Added file upload workflow, better parallel error messaging

dogversioning:
- A new workflow type is added for uploading files from a toml config
- Parallel errors where just kind of dumped out as a mess, so now they are instead caught and formatted to be more parsable while debugging
- a few counts getters have a skip_validation flag added

### Checklist

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -40,7 +40,7 @@ jobs:
           echo -e "[test]\naws_access_key_id = test\naws_secret_access_key = test" > ~/.aws/credentials
       - name: Test with pytest
         run: |
-          python -m pytest --cov-report xml --cov=cumulus_library tests
+          python -m pytest -n auto --cov-report xml --cov=cumulus_library tests
       - name: Log missing coverage
         run: |
           coverage report -m --skip-covered
--- a/tests/builders/test_upload_builder.py
+++ b/tests/builders/test_upload_builder.py
@@ -54,3 +54,20 @@ def test_unsupported_filetype(mock_manifest, mock_db_config, tmp_path):
             workflow_config={"tables": {"bad_type": {"file": "upload.pdf"}}},
         )
         builder.prepare_queries(config=mock_db_config, manifest=manifest)
+
+
+@mock.patch("cumulus_library.study_manifest.StudyManifest")
+@mock.patch("platformdirs.user_cache_dir")
+def test_wrong_cols_filetype(mock_cache, mock_db_config, tmp_path):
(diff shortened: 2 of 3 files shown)


cumulus_library/builders/file_upload_builder.py
```diff
@@ -0,0 +1,101 @@
+"""Class for generating counts tables from templates"""
```
mikix: nit: needs updating

cumulus_library/builders/file_upload_builder.py
```diff
@@ -0,0 +1,101 @@
...
+        **kwargs,
+    ):
+        with base_utils.get_progress_bar() as progress:
+            task = progress.add_task(
+                "Uploading static files...",
+                total=len(self.workflow_config["tables"]),
+            )
+            for table_name in self.workflow_config["tables"]:
+                table = self.workflow_config["tables"][table_name]
+                if table["file"].endswith(".xlsx"):
```
mikix: Do you know this is needed or was this thrown in because pandas makes it easy? Because maybe it's worth holding off on features like this unless we know we need them (just thinking about like, a world where we switch away from pandas for some reason and also want to keep feature parity while doing so).

cumulus_library/builders/file_upload_builder.py
```diff
@@ -0,0 +1,101 @@
...
+                    df = pandas.read_excel(self.toml_config_dir / table["file"])
+
+                elif (
+                    table["file"].endswith(".csv")
+                    or table["file"].endswith(".tsv")
+                    or table["file"].endswith(".bsv")
+                ):
+                    parquet_path = self.toml_config_dir / f"{table['file'][:-4]}.parquet"
+                    if "delimiter" not in table:
+                        table["delimiter"] = ","
```
mikix: The delimiter is already in the filename, why bother with a config value

cumulus_library/builders/file_upload_builder.py
```diff
@@ -0,0 +1,101 @@
...
+    ):
+        with base_utils.get_progress_bar() as progress:
+            task = progress.add_task(
+                "Uploading static files...",
+                total=len(self.workflow_config["tables"]),
+            )
+            for table_name in self.workflow_config["tables"]:
+                table = self.workflow_config["tables"][table_name]
+                if table["file"].endswith(".xlsx"):
+                    parquet_path = self.toml_config_dir / table["file"].replace(".xlsx", ".parquet")
```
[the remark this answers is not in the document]
dogversioning: yeah, i buy that

PR #433, opened 2025-12-23
Moved resource specific count logic out of jinja

dogversioning:
- All hard coded logic about resources is removed from the counts builder jinja
- That logic is now moved to the various convenience functions, with some additional data structures to handle the additional filtering rules there
- The decision to have `get_count_query` have a bunch of undefined kwargs, and then using presence/absence to determine how to handle it the template, has been reversed; it is now explicit about those args (but allows other kwargs for backwards compatibility reasons.
- SQL is regenned, which includes some sql updates from parallelization that should've gotten in here already.

### Checklist

### DIFF
--- a/cumulus_library/builders/statistics_templates/count.sql.jinja
+++ b/cumulus_library/builders/statistics_templates/count.sql.jinja
@@ -81,18 +81,19 @@ CREATE TABLE {{ table_name }} AS (
             {%- endif %}
         {%- endif %}
         WHERE 
-        {%- for filter_col in filter_cols %}
-        {{ syntax.and_delineate(loop) }}
-        {%- if filter_col.include_nulls  %}
+        {%- for filter_col in filter_cols %} {{ syntax.and_delineate(loop) }}
+        (
+            {%- if filter_col.include_nulls  %}
             p.{{ filter_col.name }} IS null OR 
-        {%- endif  %} 
+            {%- endif  %} 
             p.{{ filter_col.name }} IN (
             {%- for val in filter_col.values %}
                 '{{val}}'
                 {{- syntax.comma_delineate(loop) }}
-            {%- endfor -%}
+            {%- endfor %}
             )
-        {%- endfor -%}
+        )
+        {%- endfor %}
     ),
     {% endif %}
     null_replacement AS (
--- a/cumulus_library/builders/statistics_templates/counts_templates.py
+++ b/cumulus_library/builders/statistics_templates/counts_templates.py
@@ -21,7 +21,7 @@ class FilterColumn:
     # alias is only included for jinja compatibility reasons with CountColumn
     @property
     def alias(self) -> str | None:
-        return None
+        return None  # pragma: no cover
 
 
 @dataclass(kw_only=True)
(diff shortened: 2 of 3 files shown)


tests/test_data/core/core__count_observation_lab_month.txt
```diff
@@ -10,5 +10,5 @@
 (10, None, None, 'Brown color (qualifier value)', None)
 (10, None, None, 'Urine smell ammoniacal (finding)', 'ambulatory')
 (10, None, None, 'Urine smell ammoniacal (finding)', None)
-(20, None, None, None, 'ambulatory')
-(20, None, None, None, None)
+(10, None, None, None, 'ambulatory')
+(10, None, None, None, None)
```
dogversioning: looking into this change

tests/builders/test_counts_templates.py
```diff
@@ -337,18 +258,16 @@
...
-    JOIN secondary_powerset AS s on s.id = p.id
         JOIN other_table j ON p.code = j.other_field
     WHERE 
         p.cnt_subject_ref >= None
-        AND s.cnt_encounter_ref >= None
     GROUP BY
-        j."A",
-        j."B"
+            j."A",
+            j."y"
```
dogversioning: also looking into this change

cumulus_library/builders/statistics_templates/counts_templates.py
```diff
@@ -1,40 +1,27 @@
...
 
 
+@dataclass(kw_only=True)
+class FilterColumn:
+    name: str
+    values: list[str]
+    include_nulls: bool
+    # this should never be set, and is only included for jinja compatibility reasons
+    # with CountColumn
+    alias: str = None
```
mikix: Another option is adding something like this, to enforce the readonly bit:
```
# alias is deprecated and only included for jinja compatibility reasons with CountColumn
@property
def alias() -> str | None:
    return None
```

cumulus_library/studies/core/reference_sql/builder_practitioner_prereq.sql
```diff
@@ -0,0 +1,17 @@
+-- noqa: disable=all
```
mikix: You know, it's probably worth a CI job that runs reference_sql and complains if there's any new/changed/deleted files

cumulus_library/builders/statistics_templates/count.sql.jinja
```diff
@@ -1,263 +1,184 @@
...
-{%- set tertiary = tertiary_resource(fhir_resource) -%}
-
 {%- set missing_null = 'cumulus__none' -%}
 {#- LT02 is a indentation rule that overfires in the where
 clause construction; it simultaneously asks for indentation of 4 and 8 spaces. -#}
 -- noqa: disable=LT02
 CREATE TABLE {{ table_name }} AS (
     WITH 
-    {%- if filter_resource %}
+    {%- if filter_status == True %}
```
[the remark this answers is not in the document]
mikix: Jinja syntax is so Python-like, I assumed it was similar. Little differences like that are annoying

Issue #442, opened 2026-01-12
Make valueset generation more coding system agnostic


dogversioning:
**Problem Statement**
When it was originally put together, the valueset workflow primarily tried to shoe in RXNORM related data. This was relevant for medication-related questions, but is less useful in other contexts

**Desired behavior**
- Valuesets should allow a user to define system(s) to look up in the umls APIs
- the default keyword behavior for looking up rules should become non-default, and selectable by a flag

Issue #440, opened 2026-01-07
VSAC workflow should support adding systems


dogversioning:
**Problem Statement**
The current implementation assumes you are always pulling SNOMED codes

**Desired behavior**
Codes should be visible in vsac output

**Additional context**
Being able to provide a set of codes for filtering on would also be useful.

Issue #435, opened 2025-12-27
VSAC fails during study build "umls_valuesets_rels does not exist"


comorbidity:
```
pip freeze | grep cumulus                                                                                             
cumulus-fhir-support==1.8.0
cumulus-library==6.0.0b1
```
`cumulus-library build -s . -t glioma`

**POTENTIAL CAUSE**: _cumulus_library/builders/valueset/umls.py_.**generate_umls_tables**(...)

**EXPECTED**: `umls.py` fetches and uploads the valuesets. 

ACTUAL: 

```
 Connecting to database...
Creating/updating system tables... ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 0:00:01
Removing glioma study artifacts... ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 0:00:02
Uploading static files... ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 0:00:02
Building file_upload.toml tables... ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 0:00:02
dx_brain_tumor data present at /Users/andy/chip/cumulus-library-glioma/cumulus_library_glioma/valueset_data, skipping download.
dx_cancer data present at /Users/andy/chip/cumulus-library-glioma/cumulus_library_glioma/valueset_data, skipping download.
dx_focal_deficit data present at /Users/andy/chip/cumulus-library-glioma/cumulus_library_glioma/valueset_data, skipping download.
dx_neuropathy data present at /Users/andy/chip/cumulus-library-glioma/cumulus_library_glioma/valueset_data, skipping download.
rx_cancer_directed data present at /Users/andy/chip/cumulus-library-glioma/cumulus_library_glioma/valueset_data, skipping download.
rx_chemo_advanced data present at /Users/andy/chip/cumulus-library-glioma/cumulus_library_glioma/valueset_data, skipping download.
rx_chemo_platinum data present at /Users/andy/chip/cumulus-library-glioma/cumulus_library_glioma/valueset_data, skipping download.
rx_chemo_platinum_carboplatin data present at /Users/andy/chip/cumulus-library-glioma/cumulus_library_glioma/valueset_data, skipping download.
[...]

PR #443, opened 2026-01-12
Allow valueset workflows to only specify some sources

dogversioning:
This PR addresses the following two issues:

- #435, which occured due to a previously known issue wherein a valueset workflow expected all of a keyword, valueset, and umls lookup to be present in order to run
  - The fix for this is hotfix-adjacent, just creating empty tables in cases where certain configs aren't provided. As part of #442 , we probably need to overhaul this in a more comprehensive manner. The spirit of this PR is just to get out of the critical path.
- This does the bare minimum to start on #440 by including the coding system in tables generated from VSAC queries

Additionally, this moves the VSAC download directory to the user_cache folder, mostly to make tests not step over each other writing data to the same location.

### Checklist

### DIFF
--- a/cumulus_library/builders/valueset/umls.py
+++ b/cumulus_library/builders/valueset/umls.py
@@ -104,7 +104,7 @@ def generate_umls_tables(
         "code",
         "str",
     ]
-    if valueset_config.umls_stewards != {}:
+    if valueset_config.umls_stewards:
         query = base_templates.get_create_table_from_tables(
             table_name=f"{study_prefix}{table_prefix}umls_valuesets",
             tables=[
--- a/cumulus_library/builders/valueset/static_builder.py
+++ b/cumulus_library/builders/valueset/static_builder.py
@@ -25,9 +25,9 @@ class TableConfig:
 
 
 class StaticBuilder(BaseTableBuilder):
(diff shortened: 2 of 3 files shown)


cumulus_library/cli.py
```diff
@@ -304,7 +304,7 @@ def run_cli(args: dict):
             db=db,
             schema=schema,
             drop_table=drop_table,
-            force_upload=args.get("replace_existing", False),
+            force_upload=args.get("force_upload", False),
```
dogversioning: this has been wrong... for a bit, but since valuesets weren't getting used, it wasn't coming up.

cumulus_library/builders/valueset/umls.py
```diff
@@ -91,13 +91,37 @@ def generate_umls_tables(
...
+        "cui1",
+        "rel",
+        "rela",
+        "cui2",
+        "sab",
+        "tty",
+        "code",
+        "str",
+    ]
+    if valueset_config.umls_stewards != {}:
```
mikix: nit: maybe just `if valueset_config.umls_stewards:` ? To be futureproof against it being None or something.

mikix (approved):
Yay for adding systems to code values!

PR #447, opened 2026-01-14
More workflow reorg

dogversioning:
This PR does the following:
- Lightly backs off the cache_dir as a place to save vsac downloads, since you'll often want to include them in a study's code, and then use the file upload workflow to push them to athena
- Changes vsac downloads to include headers, and have table creation in valueset gen expect headers.
- More explicitness on typecasts around RXCUIs, which are now always strings

### Checklist

### DIFF
--- a/cumulus_library/actions/builder.py
+++ b/cumulus_library/actions/builder.py
@@ -504,7 +504,7 @@ def _run_workflow(
     query_count: int,
     run_stage: int,
     parallel: bool = False,
-) -> (list[str], bool):
+) -> tuple[list[str], bool]:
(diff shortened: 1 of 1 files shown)


cumulus_library/actions/builder.py
```diff
@@ -598,7 +599,7 @@ def _run_workflow(
                 table_name=f"{target_table}_{safe_timestamp}",
                 view_name=target_table,
             )
-    return builder.queries
+    return builder.queries, builder and builder.parallel_allowed
```
mikix: Maybe use `bool(builder and builder.parallel_allowed)` so you don't return a Builder object

PR #475, opened 2026-03-02
Removed the build-stage concept from the manifest

dogversioning:
We decided that build stages were sort of redundant, so this PR removes it. Composability is still possible with submanifests.

### Checklist

### DIFF
--- a/tests/actions/test_builder.py
+++ b/tests/actions/test_builder.py
@@ -249,3 +249,12 @@ def test_invalid_file_in_manifest(mock_db_config, tmp_path):
     manifest = study_manifest.StudyManifest(tmp_path)
     with pytest.raises(errors.StudyManifestParsingError):
         builder.build_study(mock_db_config, manifest)
+
+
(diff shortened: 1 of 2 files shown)


cumulus_library/cli_parser.py
```diff
@@ -114,12 +114,10 @@ def add_study_dir_argument(parser: argparse.ArgumentParser) -> None:
...
 
-def add_build_type_argument(parser: argparse.ArgumentParser) -> None:
-    """Adds --build-type arg to a subparser"""
+def add_stage_argument(parser: argparse.ArgumentParser) -> None:
+    """Adds --stage arg to a subparser"""
     parser.add_argument(
-        "-b",
-        "--build-type",
-        help="Selects which build type from the manifest to use",
+        "--stage", help="Selects which stage from the manifest to run", default="default"
```
mikix: Should this be comma separated and/or allowed multiple times?

Issue #474, opened 2026-03-02
Refactor manifest regular table vs dedicated schema table name selection to be centrally located


dogversioning:
This is done ad hoc in the builder and the cleaner, and should have instead a single centralized point where this logic is run.

PR #478, opened 2026-03-06
Workflow for counts tables

dogversioning:
This adds a new workflow type, `counts`, and updates the counts builder to optionally read from the workflow
as a source.

### Checklist

### DIFF
--- a/cumulus_library/study_manifest.py
+++ b/cumulus_library/study_manifest.py
@@ -204,8 +204,8 @@ class StudyManifest:
 
     def get_formatted_study_prefix(self) -> str | None:
         """Returns the appropriately formatted value for a study prefix"""
-        if self._study_config.get("advanced_options", {}).get("dedicated_schema"):
-            return f"{self._study_prefix}."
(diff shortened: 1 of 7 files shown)


cumulus_library/study_manifest.py
```diff
@@ -202,6 +201,12 @@ def get_study_prefix(self) -> str | None:
         """
         return self._study_prefix
 
+    def get_formatted_study_prefix(self) -> str | None:
+        """Returns the appropriately formatted value for a study prefix"""
+        if self._study_config.get("advanced_options", {}).get("dedicated_schema"):
+            return f"{self._study_prefix}."
+        return f"{self._study_prefix}__"
+
```
dogversioning: this starts to address #474, though i need to walk through and update usage throughout, which will be a standalone PR

mikix (approved):
Seems good - but before you release this, I would recommend converting core to it (and/or other studies), to use it for realz in case there is anything you forgot or could be smoother.

PR #479, opened 2026-03-10
Converted core counts to workflow

dogversioning:
- Cuts over the core counts to use the new workflow format
  - Adds table descriptions to the workflow for future API use
- As a byproduct of this, removes the count wrapper functions, which were actually fairly core specific in their application
- Some additional docs cleanup around the cutover

### Checklist
- [] Update template repo if there are changes to study configuration in `manifest.toml`

### DIFF
--- a/cumulus_library/studies/core/count_core.workflow
+++ b/cumulus_library/studies/core/count_core.workflow
@@ -4,7 +4,7 @@ config_type = "counts"
 source_table = "core__allergyintolerance"
 description = """A general count of patient allergic reactions by month.
 
-This table provides a summary snapshot of all allergic reactions in  the entire patient population
+This table provides a summary snapshot of all allergic reactions for the entire patient population
 that have been loaded into a database for use by the Cumulus ecosystem. It bins by intolerance category,
 intolerance code, the reaction manifestation, and the month the reaction was observed in. 
 It is primarily intended as a validation tool to ensure that data has been successfully extracted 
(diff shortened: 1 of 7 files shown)


cumulus_library/studies/core/count_core.workflow
```diff
@@ -0,0 +1,266 @@
...
+source_table = "core__encounter"
+table_cols = [
+    "class_display",
+    "type_display",
+    "serviceType_display",
+    "priority_display",
+]
+
+[tables.count_encounter_all_types_month]
+description = """A general count of encounter states by month.
```
mikix: Question of taste, I don't feel strongly: would these descriptions read better if the first line had a new line in front of it, so it was flush with the second paragraph? And we just tell consumers of this to strip() whitespace on descriptions first.

Speaking of which... do you ever describe what "format" these descriptions are in? You've got newlines instead of spaces in the middle of paragraphs, so it's probably Markdown or HTML. Do you intend to allow any formatting? Like bold etc? There should probably be validation of the string, so that study authors can't put a tracking image or scripts in there. I believe you don't use this field yet, so this can all be delayed. Just thoughts.

cumulus_library/studies/core/count_core.workflow
```diff
@@ -0,0 +1,266 @@
...
+"""
+source_table = "core__encounter"
+table_cols = [
+    "class_display",
+    "priority_display",
+    "period_start_month"
+]
+
+[tables.count_encounter_service_month]
+description = """A general count of lab observations by month.
```
mikix: Not lab obs, but encounter services. Here and in the next paragraph.

cumulus_library/studies/core/count_core.workflow
dogversioning: I don't have a strong preference on the newline thing either.

In terms of format... I probably need to have a discussion about the API part, but it my head, since the dashboard takes markdown, we should allow markdown here. Once that decision is made, I'll circle back here. I made an issue to track this: #480

Issue #490, opened 2026-03-17
README 404


comorbidity:
HTTP 404 from page 
https://docs.smarthealthit.org/cumulus/library/study-list.html

to 

https://docs.smarthealthit.org/cumulus/library/github.com/smart-on-fhir/cumulus-library-loinc

Issue #486, opened 2026-03-12
Cast to types before writing to parquet in file upload workflow


dogversioning:
things can get a little loosey goosey inferring data types from csv/parquets, and it may create downstream complications. in a case where a user provides the expected datatypes, we should try to get that into the parqeut files before we write them to disk, so that we are more assured of things working correctly when we get to the parquet CTAS serde step.

PR #491, opened 2026-03-17
Example study, v6 bugfixes

dogversioning:
This PR has the following changes:
- Docs: 
  - Added a 'write a study from scratch' document
    - There's now a new `example` study referenced by those docs
  - Fixed duplicate naming issue in counts doc masking the file upload doc
- User request: made "build:serial" the type used if no type is provided
- Bugfixes:
  - Addressed #486
  - Addressed #490 
  - Fixed workflows not properly adding tables to the build log
  - Changed annotated counts join from inner to left, preserving more cube properties
  - Made labels render in UI when actions are run in serial mode

### Checklist

### DIFF
--- a/cumulus_library/studies/example/manifest.toml
+++ b/cumulus_library/studies/example/manifest.toml
@@ -36,7 +36,7 @@ files = [
 
 [[stages.analysis]]
 label = "Generate summary counts"
-types = "build:parallel"
+type = "build:parallel"
 files = [ 
     "workflows/counts.workflow",
 ]


tests/builders/test_counts_builder.py
```diff
@@ -226,17 +227,13 @@ def test_counts_workflow(mock_db_core_config):
...
+        None,
+        None,
+        None,
+        None,
     )
     assert res[-1] == (2, "422650009", None, "http://snomed.info/sct", "Social isolation (finding)")
-    # let's also validate annotation properties - i.e. it should not have empty fields from cubing
-    for row in res:
-        assert row[-1] is not None
-        assert row[-2] is not None
```
mikix: What is this all about?

PR #509, opened 2026-03-31
Merge main into the NLP feature branch

mikix:
Just to handle some of the conflicts before they get too big.

### Checklist


dogversioning (approved):
For posterity - reviewed locally, github was supremely unhelpful about this commit for some reason.

PR #511, opened 2026-04-01
Changed the default stage behavior to match "all"

dogversioning:
By request, running the entire study manifest is now the default behavior if a default stage is not specified.

### Checklist

### DIFF
--- a/cumulus_library/study_manifest.py
+++ b/cumulus_library/study_manifest.py
@@ -243,7 +243,7 @@ class StudyManifest:
         stage_contents = self._study_config.get("stages", {}).get(stage, [])
         # if we're looking for a default stage and we don't find it, assume we should run everything
         if stage == "default" and stage_contents == []:
-            stage_contents = self._study_config.get("stages", {})["all"]
+            stage_contents = self._study_config.get("stages", {}).get("all", [])
         return stage_contents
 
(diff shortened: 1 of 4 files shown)


cumulus_library/study_manifest.py
```diff
@@ -243,7 +240,11 @@ def get_stages(self) -> list:
 
     def get_stage(self, stage) -> list:
         """Returns the contents of the specified stage"""
-        return self._study_config.get("stages", {}).get(stage, [])
+        stage_contents = self._study_config.get("stages", {}).get(stage, [])
+        # if we're looking for a default stage and we don't find it, assume we should run everything
+        if stage == "default" and stage_contents == []:
+            stage_contents = self._study_config.get("stages", {})["all"]
```
mikix: This is odd - you have a default value in case `stages` in not in the config, but then you unconditionally reference ["all"] - maybe you should unconditionally reference `stages` too? Do we know that it's available?

mikix (approved):

cumulus_library/study_manifest.py
dogversioning: chalk this up to insomnia brain - setting this back to conditional access as well. 99% of the time we should have it, but it might not be set in some unit test cases.

Issue #508, opened 2026-03-30
DATE in file_upload not working as expected


comorbidity:
**Describe the bug**
Providing a simple CSV with DATE,DATE,boolean failed 

**To Reproduce** (cumulus-library==**6.0.3**) 
```
[tables.include_study_period]
delimiter = ','
col_types = ["DATE","DATE","boolean"] 
file = "include_study_period.csv"
```

`include_study_period.csv`
```
period_start,period_end,include_history
2016-01-01,2026-01-01,true
```

**Expected behavior**
DATE should be supported as per 
https://docs.smarthealthit.org/cumulus/library/workflows/file-upload.html

**Additional context**
```
File "/Users/andy/chip/cumulus-library-ibd-cds/ve/lib/python3.11/site-packages/cumulus_library/builders/file_upload_builder.py", line 187, in prepare_queries
    df = df.astype(type_dict)
         ^^^^^^^^^^^^^^^^^^^^
  File "/Users/andy/chip/cumulus-library-ibd-cds/ve/lib/python3.11/site-packages/pandas/core/generic.py", line 6642, in astype
    res_col = col.astype(dtype=cdt, copy=copy, errors=errors)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/andy/chip/cumulus-library-ibd-cds/ve/lib/python3.11/site-packages/pandas/core/generic.py", line 6665, in astype
    new_data = self._mgr.astype(dtype=dtype, copy=copy, errors=errors)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/andy/chip/cumulus-library-ibd-cds/ve/lib/python3.11/site-packages/pandas/core/internals/managers.py", line 449, in astype
    return self.apply(
           ^^^^^^^^^^^
  File "/Users/andy/chip/cumulus-library-ibd-cds/ve/lib/python3.11/site-packages/pandas/core/internals/managers.py", line 363, in apply
    applied = getattr(b, f)(**kwargs)
              ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/andy/chip/cumulus-library-ibd-cds/ve/lib/python3.11/site-packages/pandas/core/internals/blocks.py", line 784, in astype
    new_values = astype_array_safe(values, dtype, copy=copy, errors=errors)
[...]

PR #512, opened 2026-04-01
Better type casting in file_upload action

dogversioning:
This addresses #508 by making an explicit cast from hive types to the corresponding numpy types, which prevents type cast errors in pandas.

### Checklist

### DIFF
--- a/cumulus_library/base_utils.py
+++ b/cumulus_library/base_utils.py
@@ -12,7 +12,7 @@ import rich
 import sqlglot
 from rich import progress
 
-from cumulus_library import databases, study_manifest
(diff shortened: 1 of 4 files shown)


mikix (approved):
`ETOOMANYTYPESYSTEMS`

dogversioning:
> `ETOOMANYTYPESYSTEMS`

speaking of, switched to a third type system due to pandas not respecting dates

comorbidity:
FYI issues remain in DATE fields for Manifest TOMLs as documented here: 
https://github.com/smart-on-fhir/cumulus-library/issues/517

PR #513, opened 2026-04-01
Add support for select_by_table by querying the database

mikix:
This also adds a new CLI arg: --etl-phi-dir to tell Library where the PHI folder is, which holds the codebook with the salt we need to match original note IDs to Athena table values.

FYI, there are now two new CLI args in the NLP feature work: `--etl-phi-dir` and `--note-dir`

### Checklist

### DIFF
--- a/cumulus_library/note_utils.py
+++ b/cumulus_library/note_utils.py
@@ -1,16 +1,20 @@
 import base64
+import binascii
 import dataclasses
 import email
+import hmac
+import json
 import os
+import pathlib
 import re
 import urllib.parse
-from collections.abc import Callable, Generator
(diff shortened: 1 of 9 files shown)


cumulus_library/note_utils.py
```diff
@@ -60,6 +81,20 @@ def _scan(self) -> None:
         self._files = self._flatten_note_dirs()
         self._total_size = sum(f.size for f in self._files)
 
+    def _get_fsspec_path(self, path: str) -> tuple[fsspec.AbstractFileSystem, str]:
```
dogversioning: nit: type path as pathlib.Path or string

dogversioning (approved):

PR #514, opened 2026-04-07
Use new note utils code in cfs

mikix:
### Checklist


cumulus_library/databases/utils.py
```diff
@@ -13,7 +13,7 @@
 from cumulus_library import base_utils, db_config, errors
 from cumulus_library.databases import athena, base, duckdb
 
-SQL_NAME_CHARS = set(string.ascii_letters + string.digits + "_")
+SQL_NAME_CHARS = set(string.ascii_letters + string.digits + "-_")
```
dogversioning: ok, can you tell me a little more about this? hyphens are technically not allowed in SQL names, and you have to escape table names to get them into a query. I :think: we're doing this everyplace but i'm not sure, and just disallowing them solves this problem.

dogversioning (approved):

cumulus_library/databases/utils.py
mikix: Ah, because I noticed this when touching the ETL code that was similar - it allows hyphens because the cohort tables that we generate for the GIC stuff. I know we haven't really gone ahead with that much, but we decided at the time that it was better to just use the GIC uuid directly in the table name rather than convert hyphens to underscores and have to manage both versions of the ID.

And those cohort tables are exactly the kind of thing you'd pass as a basis table.

We could revisit some of those decisions, but that was the thinking here anyway.

mikix: We talked elsewhere and since we don't need the GIC stuff right now, we're going to go with the tighter requirement and disallow hyphen. Can always add it back later if we need.

PR #515, opened 2026-04-08
Detect changes in table ref counts

dogversioning:
This PR adds a new system table, [study]__lib_ref_summary, which counts the instances of a ref in every table in a study, and reports out the net change from the prior run.

A few byproducts of this change:
- Some updates to the single select template, which was poorly exercised before
- Since we're using parallel for retrieving data now, changed the db `parallel_write` method to `parallel_execute` (the old version is now an alias for the new one)
- Did a little bit of refactor on local determination of presence/absence of prefix.

### Checklist


cumulus_library/actions/builder.py
```diff
@@ -631,6 +634,131 @@ def _run_workflow(
     return builder.queries, bool(builder and builder.parallel_allowed)
 
 
+def log_ref_summary(config: base_utils.StudyConfig, manifest: study_manifest.StudyManifest):
```
dogversioning, line 637: this function is long, mostly due to all the template lookups - I :could: split this up? but it would be just for readability's sake. Open to pitches.

cumulus_library/actions/builder.py
```diff
@@ -631,6 +634,131 @@ def _run_workflow(
...
+                    f'"{config.schema}"."{summary_table}")'
+                ]
+            ],
+        )
+        prior_results_raw = config.db.cursor().execute(prior_results_query).fetchall()
+        prior_results = {}
+        for res in prior_results_raw:
+            if res[0] not in prior_results.keys():
+                prior_results[res[0]] = {}
+            prior_results[res[0]][res[1]] = res[2]
```
mikix: Adding some variables and using setdefault might make this easier to scan:

```
for res in prior_results_raw:
    table_name = res[0]
    ref_type = res[1]
    ref_count = res[2]
    table_dict = prior_results.setdefault(table_name, {})
    table_dict[ref_type] = ref_count
```

cumulus_library/actions/builder.py
```diff
@@ -631,6 +634,131 @@ def _run_workflow(
...
+        for res in parallel_results:
+            formatted_table = (
+                str(sqlglot.parse_one(res.query, dialect=config.db.db_type).find(sqlglot.exp.Table))
+                .split(".")[1]
+                .replace('"', "")
+            )
+            prior = prior_results.get(formatted_table, {}).get(res.columns[0], None)
+            if prior:
+                prior = format(((res.rows[0][0] - prior) / prior) * 100, ".3f")
+                if prior != 0:
```
mikix: `prior` was just formatted to a string, so it'll never equal 0, yeah? Maybe save the pre-formatted calculation to a variable.

mikix (approved):

Issue #517, opened 2026-04-08
Minor: CURRENT_DATE and NULL not supported as DATE in workflow TOML(s)


comorbidity:
**Problem Statement**
* `CURRENT_DATE` is a valid DATE in Athena but not in Apache Arrow / HIVE types / etc. 
* NULL is not an allowed DATE field but STRING type is 

**Desired behavior**
* `CURRENT_DATE` is treated like a true date field like `CURRENT_DATE` = date('2026-04-08') 
* NULL is allowable DATE field 

**Has workaround**
* IBD studies and presumably other studies as well can implement default  CURRENT_DATE behavior when DATE fields are NULL. 
* changing the TOML to STRING and manually casting in SQL space to date(....)

**How to reproduce** 
**cumulus-library version** = `6.0.4`

Example TOML file
```
[tables.include_study_period]
col_types = ["DATE","DATE","boolean"] #period_start,period_end,include_history
file = "include_study_period.csv"
```

Example CSV file with `CURRENT_DATE`
```
period_start,period_end,include_history
2016-01-01,CURRENT_DATE,true
```

Example CSV file with NULL
```
period_start,period_end,include_history
2016-01-01,,true
```

**cumulus-library build -t ibd** 
```
 Connecting to database...
Creating/updating system tables... ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 0:00:03
Removing ibd study artifacts... ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━   0% 0:00:00
Uploading static files... ━━━━━━━━━━━━━━━━━━━━╺━━━━━━━━━━━━━━━━━━━  50% 0:00:01
Traceback (most recent call last):
  File "/Users/andy/chip/cumulus-library-ibd-cds/ve/bin/cumulus-library", line 8, in <module>
    sys.exit(main_cli())
             ^^^^^^^^^^
  File "/Users/andy/chip/cumulus-library-ibd-cds/ve/lib/python3.11/site-packages/cumulus_library/cli.py", line 501, in main_cli
    main()  # pragma: no cover
    ^^^^^^
  File "/Users/andy/chip/cumulus-library-ibd-cds/ve/lib/python3.11/site-packages/cumulus_library/cli.py", line 497, in main
    return run_cli(args)
           ^^^^^^^^^^^^^
[...]

Issue #492, opened 2026-03-17
minor: array characters "[]" stripped from description messages


comorbidity:
example 

```
[[actions]]
description = "cohort for case definition aspects [dx, rx, lab, proc]"
type = "build:parallel"
files = [
	"athena/ibd__cohort_casedef_dx.sql",
	"athena/ibd__cohort_casedef_rx.sql",
	"athena/ibd__cohort_casedef_lab.sql",
	"athena/ibd__cohort_casedef_proc.sql"]
```
during build time shows as 
```
Building cohort for case definition aspects  tables... ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 0:00:22
```

dogversioning:
ok, dug into this a bit - brackets are used by the console UI library for indicating textual style changes. I'll add a validator on the manifest parsing to suggest replacing them with another character.

PR #519, opened 2026-04-14
Support nullable dates in file upload, forbid brackets in labels

dogversioning:
This PR adresses the following issues:
- Addresses #517 with some multilayer date casting
- Addresses #492 by forbidding rich control characters in console output

### Checklist

### DIFF
--- a/cumulus_library/builders/file_upload_builder.py
+++ b/cumulus_library/builders/file_upload_builder.py
@@ -5,27 +5,13 @@ import re
 import sys
 
 import msgspec
+import numpy
 import pandas
 import platformdirs
 import pyarrow
 
 from cumulus_library import BaseTableBuilder, base_utils, errors, study_manifest
 from cumulus_library.template_sql import base_templates
-
-"""
-A file upload config looks like this
-
-type="file_upload"
-[tables]
-[table.name] # the table in db will be called 'name', snaked cased for sql compatibility
-files = ["relative/path/from/config/file.csv"]
-# The rest of these are optional
-delimiter = "|" # comma by default, only applies to csv/tsv/bsv
-col_types =["STRING","DATE","DOUBLE"] # default is all strings. Only relevant for Athena.
-# See https://cwiki.apache.org/confluence/display/Hive/LanguageManual+Types for valid types
-[table.other_name]
-file = "other/path/file.csv"
-"""
 
 
 class FileUploadTask(msgspec.Struct, forbid_unknown_fields=True):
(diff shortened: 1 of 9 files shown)


dogversioning:
Noting that I tried adding 3.14 to the test matrix and it failed, so i made an issue rather than piling on: #520

cumulus_library/builders/file_upload_builder.py
```diff
@@ -79,6 +65,25 @@ def __init__(self, *args, toml_config_path: pathlib.Path, **kwargs):
                     f"{self._toml_config_path} is invalid."
                 )
 
+    def get_pandas_types(self, df, pandas_types):
+        dtype_dict = {}
+        date_cols = []
+        if pandas_types is None or len(pandas_types) != len(df.columns):
```
Luke-Sikina: `len(pandas_types) != len(df.columns)` Interesting predicate - is there a chance this is too defensive? When do you expect these arrays to be mismatched, and why should that be treated the same as a null?

cumulus_library/builders/file_upload_builder.py
```diff
@@ -79,6 +65,25 @@ def __init__(self, *args, toml_config_path: pathlib.Path, **kwargs):
                     f"{self._toml_config_path} is invalid."
                 )
 
+    def get_pandas_types(self, df, pandas_types):
```
Luke-Sikina: This does a bit more than just getting types, right? You also do the date -> str casting here. I wouldn't have guessed that from the method name.

Luke-Sikina (approved):

cumulus_library/builders/file_upload_builder.py
dogversioning: good point - this started off smaller and the scope expanded, i'll reword this.

cumulus_library/builders/file_upload_builder.py
dogversioning: so down on line 200, there's a point at which we do a check that says 'ok, you've got the raw data, and perhaps a list of user defined columns, and we'll throw the error there if they are different.' This is sort of papering over that while we do all the reading bits, but we could probably throw that here, or maybe even earlier, and not be so defensive here.

cumulus_library/builders/file_upload_builder.py
```diff
@@ -79,6 +65,42 @@ def __init__(self, *args, toml_config_path: pathlib.Path, **kwargs):
                     f"{self._toml_config_path} is invalid."
                 )
 
+    def get_pandas_read_params(
```
mikix, line 68: thought: how hard would it to stop using pandas in Library? I feel like we have grown a few work-arounds for its loosy goosy typing, that maybe if we just used pyarrow directly we could avoid. (doesn't solve all the type conversion that we'd still have to do between like, database types and pandas/pyarrow - but might help with stuff like this)

dogversioning: So I tried doing this with pyarrow directly and was really struggling with it - i think for this case (user supplied data of unknown size/correctness), it's easier to pipe it through pandas for inspection than to use three different tools (csv reader, like openpyxl, and then arrow directly) to perform this step.

PR #526, opened 2026-04-23
Add initial model support (no batching, no writing yet)

mikix:
This adds some new files to `cumulus_library/builds/nlp/`:
- `caching.py`: this is a copy of caching code from the ETL, nothing new here
- `models.py`: this is a copy of "talking directly to models" from the ETL - only interesting new stuff is that I commented out the batching support and made it sync instead of async.
- `driver.py`: this is new code, to iterate over notes and handle nlp responses.
- `workflow.py`: this is existing code from `nlp_builder.py`, just moved into this file for sharing

### Checklist

### DIFF
--- a/cumulus_library/builders/nlp/caching.py
+++ b/cumulus_library/builders/nlp/caching.py
@@ -1,0 +1,66 @@
+"""Local caching of NLP results, to avoid future calls"""
+
+import hashlib
+from collections.abc import Callable
+from typing import TypeVar
+
+import cumulus_fhir_support as cfs
+
+Obj = TypeVar("Obj")
+
+# TODO MIKE: add back when implementing batching
+
+# def _cache_metadata_path(cache_dir: cfs.FsPath, namespace: str, filename: str) -> cfs.FsPath:
+#     return cache_dir.joinpath(f"nlp-cache/{namespace}/{filename}")
+
+
+# def cache_metadata_write(cache_dir: cfs.FsPath, namespace: str, content: dict) -> None:
+#     path = _cache_metadata_path(cache_dir, namespace, "metadata.json")
+#     path.parent.makedirs()
+#     path.write_json(content, indent=2)
+
+
+# def cache_metadata_read(cache_dir: cfs.FsPath, namespace: str) -> dict:
+#     path = _cache_metadata_path(cache_dir, namespace, "metadata.json")
+#     return path.read_json(default={})
+
+
+def _cache_path(cache_dir: cfs.FsPath, namespace: str, checksum: str) -> cfs.FsPath:
+    return cache_dir.joinpath(f"nlp-cache/{namespace}/{checksum[0:4]}/sha256-{checksum}.cache")
+
+
+def cache_checksum(note_text: str) -> str:
+    return hashlib.sha256(note_text.encode("utf8"), usedforsecurity=False).hexdigest()
+
+
+def cache_write(cache_dir: cfs.FsPath, namespace: str, checksum: str, content: str) -> None:
+    path = _cache_path(cache_dir, namespace, checksum)
+    path.parent.makedirs()
+    path.write_text(content)
+
+
+def cache_read(cache_dir: cfs.FsPath, namespace: str, checksum: str) -> str | None:
+    path = _cache_path(cache_dir, namespace, checksum)
+    return path.read_text(default=None)
+
+
+def cache_wrapper(
+    cache_dir: cfs.FsPath,
+    namespace: str,
+    checksum: str,
+    from_file: Callable[[str], Obj],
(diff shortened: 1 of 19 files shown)


cumulus_library/cli.py
```diff
@@ -345,7 +351,8 @@ def run_cli(args: dict):
                     options=args["options"],
                 )
             elif args["action"] == "build":
-                notes = note_utils.NoteSource(args["note_dir"], phi_dir=args["etl_phi_dir"])
+                notes = note_utils.NoteSource(args["note_dir"])
+                nlp_config = note_utils.NlpConfig(args)
```
dogversioning, line 355: how would you feel about making the NlpConfig part of base_utils.StudyConfig? I think there's a lot of plumbing involved in setting things up to take this new class in all the builder logic, and just saying that 'ok, the StudyConfig has this' simplifies that a little? weakly held opinion, since this is semantically kind of its own thing and it builder specific, but worth at least chewing on a bit.

cumulus_library/builders/nlp/caching.py
```diff
@@ -0,0 +1,65 @@
...
+
+# def cache_metadata_write(cache_dir: cfs.FsPath, namespace: str, content: dict) -> None:
+#     path = _cache_metadata_path(cache_dir, namespace, "metadata.json")
+#     path.parent.makedirs()
+#     path.write_json(content, indent=2)
+
+
+# def cache_metadata_read(cache_dir: cfs.FsPath, namespace: str) -> dict:
+#     path = _cache_metadata_path(cache_dir, namespace, "metadata.json")
+#     return path.read_json(default={})
```
dogversioning, line 25: Maybe add a TODO on this? I'm assuming these are things you'll pull back later.

cumulus_library/builders/nlp/driver.py
```diff
@@ -0,0 +1,230 @@
...
+                    rich.print(
+                        f"Could not match span received from NLP server for {note_ref}: {orig_span}"
+                    )
+
+            parsed[key] = new_spans
+
+        return all_found
+
+    def _write_notes_to_output(self) -> None:
+        print("TODO MIKE: write this to parquet", self._notes)  # noqa: T201
```
dogversioning, line 221: this might trip the repo's ruff protections against this kind of thing? maybe rich.print this instead?

tests/builders/test_nlp_builder.py
```diff
@@ -49,21 +74,57 @@ def test_unexpected_config_field(tmp_path, note_source):
...
+
+
+def test_task_without_schema(tmp_path, note_source):
+    workflow_path = f"{tmp_path}/nlp.workflow"
+    with open(workflow_path, "w", encoding="utf8") as f:
+        f.write("""
+config_type="nlp"
+[[task]]
+name="test"
+""")
```
dogversioning: noting that conftest has a `write_toml` method for handling this kind of thing.

dogversioning (approved):

cumulus_library/cli.py
mikix: My thinking was trying to isolate this weird new thing that hopefully only this builder has to care about, but I hear you.

I think all the new plumbing is "just" in cli.py? In which case, it doesn't stress me out. But if that grows, then I could see throwing it in study config. Is that fair?

cumulus_library/builders/nlp/caching.py
mikix: Yeah it's all related to batching - I can add a todo

cumulus_library/builders/nlp/driver.py
mikix: Oh I think I added that noqa for that. Why did I do that instead of rich.print? I dunno, will switch to rich.print.

cumulus_library/cli.py
dogversioning: yeah i buy that

cumulus_library/builders/nlp/caching.py
mikix: Added TODOs above the three commented-out batching blocks - here and two in models.py

tests/builders/test_nlp_builder.py
mikix: Ah so it does! Converted.

cumulus_library/builders/nlp/driver.py
mikix: OK turns out I did it this way to avoid terminal colors in the output, so I could do quick and dirty tests that checked we were "writing" the correct results out. Once this is converted to actually write to disk, those tests will be corrected and this print line will be dropped. I'm just gonna leave it like this for now.

Issue #524, opened 2026-04-17
very minor: build stage should report which file caused the failure


comorbidity:
**Problem Statement**
I want to know which file caused the upload to fail, but its is not clear from the stack trace. 
```
"Error tokenizing data. C error: Expected 3 fields in line 12, saw 5" 
```

**Desired behavior**
Specify which CSV file caused the failure. 

**Workaround**
Obviously I shouldn't upload invalid CSV file, but I dont know which file caused this issue. :) 

**Stack Trace**
```
  File "cumulus-library-ibd-cds/ve/lib/python3.11/site-packages/cumulus_library/actions/builder.py", line 108, in build_study
    w_queries, parallel_allowed = _run_workflow(
                                  ^^^^^^^^^^^^^^
  File "cumulus-library-ibd-cds/ve/lib/python3.11/site-packages/cumulus_library/actions/builder.py", line 611, in _run_workflow
    builder.prepare_queries(
  File "cumulus-library-ibd-cds/ve/lib/python3.11/site-packages/cumulus_library/builders/file_upload_builder.py", line 147, in prepare_queries
    df = pandas.read_csv(
         ^^^^^^^^^^^^^^^^
  File "cumulus-library-ibd-cds/ve/lib/python3.11/site-packages/pandas/io/parsers/readers.py", line 1026, in read_csv
    return _read(filepath_or_buffer, kwds)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "cumulus-library-ibd-cds/ve/lib/python3.11/site-packages/pandas/io/parsers/readers.py", line 626, in _read
    return parser.read(nrows)
           ^^^^^^^^^^^^^^^^^^
  File "cumulus-library-ibd-cds/ve/lib/python3.11/site-packages/pandas/io/parsers/readers.py", line 1923, in read
    ) = self._engine.read(  # type: ignore[attr-defined]
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "cumulus-library-ibd-cds/ve/lib/python3.11/site-packages/pandas/io/parsers/c_parser_wrapper.py", line 234, in read
    chunks = self._reader.read_low_memory(nrows)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "pandas/_libs/parsers.pyx", line 838, in pandas._libs.parsers.TextReader.read_low_memory
[...]

PR #528, opened 2026-04-28
Handle malformed file uploads

dogversioning:
This addresses #524 by catching errors related to pandas parsing when reading data in during a file upload.

### Checklist

### DIFF
--- a/cumulus_library/builders/file_upload_builder.py
+++ b/cumulus_library/builders/file_upload_builder.py
@@ -102,6 +102,64 @@ class FileUploadBuilder(BaseTableBuilder):
             df[col] = pandas.to_datetime(df[col])
         return df
 
+    def _read_file(self, file, table, cache_dir, task_name):
+        table_filename = pathlib.Path(file).name
+
+        # First, we're going to read data into a pandas dataframe.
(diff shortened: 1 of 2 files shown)


cumulus_library/builders/file_upload_builder.py
```diff
@@ -159,51 +160,55 @@ def prepare_queries(
...
-                        dtype_dict, date_cols = self.get_pandas_read_params(
-                            df, pandas_types, task_name
-                        )
-
-                    else:
-                        raise errors.FileUploadError(
-                            f"{table['file']} is not a supported upload file type.\n"
-                            "Supported file types: csv, bsv, tsv, xlsx, parquet"
-                        )
+                    try:
```
mikix, line 205: nit: such a huge indent makes me wish for an indent around a method call. And that makes me feel like this method is too long in general. But doesn't need to be solved here.

mikix (approved):
Bro, just don't upload malformed files. Simple

cumulus_library/builders/file_upload_builder.py
dogversioning: this ended up creating a bit of a debugging headache so i went ahead and moved the file parsing to a dedicated private method.

PR #534, opened 2026-04-30
Actually write out NLP results into parquets and make table

mikix:
This does a few interesting new things:
- Tracks previously uploaded documents in a table_name_v1.ids file, stored in the parent directory of the parquet files. That is, the parquet layout will look like:
  - $BASE/cumulus_user_uploads/{db}/{study}/{table}_v2/*.parquet
  - $BASE/cumulus_user_uploads/{db}/{study}/{table}_v2.ids If we are scanning through notes, and the note is already in the .ids file, we skip it. This saves time and keeps the parquets from having duplicate content. (this exact pattern is used by the ETL, including the same file path)
- Adds actual code behind the --clean-nlp flag to clean up stored parquet files and .ids files for all existing versions of a given task. This is separate from normal Library clean operations, which only touch table data. (this is what the ETL does too, with its own "--clean" arg, which I felt was not specific enough in the Library case)
- Adds "spans" conversion code for model results, to switch from spans of text to spans of character offset integers. (this exact step is used by the ETL)
- Adds get_remote_upload_path(study, topic) to DatabaseBackend. This is very similar to get_remote_path(), but encodes the full path for uploading files.

### Checklist

### DIFF
--- a/cumulus_library/builders/nlp_builder.py
+++ b/cumulus_library/builders/nlp_builder.py
@@ -3,14 +3,17 @@
 import json
 import pathlib
 import sys
+import tempfile
 
 import cumulus_fhir_support as cfs
 import jambo
 import msgspec
+import pandas
+import pyarrow
 import rich
 
 import cumulus_library
-from cumulus_library import note_utils
+from cumulus_library import databases, note_utils
 from cumulus_library.builders.nlp import driver, workflow
 from cumulus_library.template_sql import base_templates
 
@@ -138,6 +141,7 @@ class NlpBuilder(cumulus_library.BaseTableBuilder):
             tasks=self._workflow_config.task,
             filters=note_filters,
             nlp_config=self._nlp_config,
+            db=config.db,
         )
 
         # Print stat block, because that's interesting feedback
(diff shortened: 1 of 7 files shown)


cumulus_library/builders/nlp_builder.py
```diff
@@ -153,31 +157,26 @@ def _table_is_view(self, study_config: cumulus_library.StudyConfig) -> bool:
...
+        """Turns a pyarrow schema into a list of parquet types."""
+        # SQL templates expect a list of parquet types. But the NLP code operates on Pyarrow
+        # schemas for the most part (with a smattering of pydantic).
+        #
+        # The easiest way to convert these is through pandas, because the DatabaseBackend class has
+        # a conversion method for that. and the easiest way through pandas is reading a parquet from
+        # disk.
+        #
+        # So we write an empty parquet to disk, have pandas read it, then have the database convert
+        # the pandas dtype to parquet types. Simple...
```
dogversioning, line 172: you love it! everyone's favorite, the multistage type conversion.

cumulus_library/databases/athena.py
```diff
@@ -112,11 +116,22 @@ def col_parquet_types_from_pandas(self, field_types: list) -> list:
                     )
         return output
 
+    def _get_result_config(self) -> dict:
+        workgroup = self.connection._client.get_work_group(WorkGroup=self.work_group)
+        return workgroup["WorkGroup"]["Configuration"]["ResultConfiguration"]
+
```
dogversioning, line 122: good call on breaking this out

dogversioning (approved):

PR #537, opened 2026-05-01
NLP cleanups

mikix:
- Print token usage to console
- Don't prefix "nlp2_" to table names, just use the task name as given.
- Add some test coverage

### Checklist

### DIFF
--- a/cumulus_library/builders/nlp/models.py
+++ b/cumulus_library/builders/nlp/models.py
@@ -251,8 +251,7 @@ class OpenAIProvider(Provider):
         if self.supports_schema:
             response_format = self.pydantic_to_response_format(schema)
         else:
-            # MIKE TODO: add coverage
-            response_format = {"type": "json_object"}  # pragma: no cover
+            response_format = {"type": "json_object"}
 
         return {
             "model": self.deployment,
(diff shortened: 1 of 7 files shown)


cumulus_library/builders/nlp/models.py
```diff
@@ -278,11 +277,7 @@ def _process_completion_result(
             self.stats.output_tokens += response.usage.completion_tokens
 
         choice = response.choices[0]
-        if self.supports_schema and choice.message.parsed:
-            # TODO MIKE: add coverage
-            parsed = choice.message.parsed  # pragma: no cover
-        else:
-            parsed = schema.model_validate_json(choice.message.content)
+        parsed = schema.model_validate_json(choice.message.content)
```
mikix, line 280: The top path turns out to not be used anymore, it's just a dead path ported from the ETL code - but because of how we pass in the schema to use to openai, it will never give us back an already-parsed model, we have to parse it ourselves.

cumulus_library/builders/nlp_builder.py
```diff
@@ -145,8 +173,8 @@ def _run_nlp(self, config: cumulus_library.StudyConfig) -> None:
         )
 
         # Print stat block, because that's interesting feedback
-        # TODO MIKE: print token stats too
         self._print_note_stats(names=[t.name for t in self._workflow_config.task], stats=self.stats)
+        self._print_token_stats(self.stats)
```
dogversioning: i might want a cli flag to disable this output?

mikix: OK, I added `--no-nlp-stats` to disable token and note stats.

dogversioning (approved):

PR #539, opened 2026-05-04
Add some NLP workflow docs

mikix:
Along the way, change the config layout to match existing workflows ([table.name] dictionary rather than a [[task]] array with a name field).

### Checklist

### DIFF
--- a/docs/workflows/nlp.md
+++ b/docs/workflows/nlp.md
@@ -1,0 +1,171 @@
+---
+title: NLP
+parent: Workflows
+grand_parent: Library
+nav_order: 3
+# audience: clinical researchers, IRB reviewers
+# type: reference
+---
+
+# NLP
+
+This document aims to provide help in configuring natural language processing (NLP) workflows as
+part of a Cumulus Library study.
+
+## Use cases
+
+The coded metadata in FHIR inevitably only tells part of the story.
+There will always be some interesting extra bits in the clinical notes (which aren't easily exposed
+in the core tables for study use).
+In those cases, you'll want to define some NLP workflows that can run a custom prompt and clinical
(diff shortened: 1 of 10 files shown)


docs/workflows/nlp.md
```diff
@@ -0,0 +1,171 @@
...
+# note. If you do not define this field or leave it empty, just the text of the clinical note will
+# be used (i.e. the default is `"%CLINICAL-NOTE%"`).
+user_prompt = ""
+
+# `response_schema` points at a file adjacent to this workflow file. It contains a JSON schema
+# for the expected response from the LLM. If the LLM's response cannot be parsed into this format,
+# it will be ignored. This schema also helps defines the resulting table schema.
+# Often, studies will use a bit of Python code to generate this file from some Pydantic models.
+# See the example_nlp study or other studies for comparison.
+response_schema = "schema.json"
```
dogversioning, line 85: open question: should we not take json, and instead point to a pydantic class, and handle json conversion under the hood?

dogversioning (approved):

docs/workflows/nlp.md
mikix: Maybe... What would pointing at a pydantic class look like? An import path for a model? Like `cumulus_library_ibd.models.MyModel`

Seems plausible / good. (Except I will say that it makes it easier to mess up and change the model without realizing that you need to update the version.)

But maybe we could mark that as a TODO item for the future?

dogversioning: yeah i think that's fine as a TODO/ticket

mikix: https://github.com/smart-on-fhir/cumulus-library/issues/540

PR #543, opened 2026-05-05
Add docker compose file for local vLLM support, with docs

mikix:
### Checklist

### DIFF
--- a/docs/compose-nlp.yaml
+++ b/docs/compose-nlp.yaml
@@ -1,0 +1,57 @@
+services:
+
+  gpt-oss-120b:
+    # Needs 80GB of GPU memory. A g6.12xlarge EC2 instance should work.
+    # Docs: https://recipes.vllm.ai/openai/gpt-oss-120b
+    image: vllm/vllm-openai:v0.10.2
+    environment:
+      - VLLM_ATTENTION_BACKEND=TRITON_ATTN_VLLM_V1  # needed for NVIDIA A10 and A100 chips
+    command:
+      - --download-dir=/data
+      - --port=8086
+      - --model=openai/gpt-oss-120b
+      - --revision=b5c939de8f754692c1647ca79fbf85e8c1e70f8a
+      - --tensor-parallel-size=4
+    shm_size: 32G
+    healthcheck:
+      test: ["CMD", "wget", "localhost:8086/health", "--output-document=/dev/null"]
+      start_period: 40m  # give plenty of time for startup, loading is slow
+    volumes:
+      - vllm-data:/data
+    ports:
+      - 8086:8086
+    deploy:
+      resources:
+        reservations:
+          devices:
+            - capabilities: [gpu]
+
+  llama4-scout:  # WIP, have not gotten it to run successfully yet
+    # Needs 80GB of GPU memory. A g6.12xlarge EC2 instance should work.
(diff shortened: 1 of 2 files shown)


docs/compose-nlp.yaml
```diff
@@ -0,0 +1,57 @@
...
+    # Docs: https://recipes.vllm.ai/meta-llama/Llama-4-Scout-17B-16E-Instruct
+    image: vllm/vllm-openai:v0.11.0
+    environment:
+      - HF_TOKEN
+      - HUGGING_FACE_HUB_TOKEN
+    command:
+      - --download-dir=/data
+      - --port=8087
+      - --model=nvidia/Llama-4-Scout-17B-16E-Instruct-NVFP4
+      - --revision=9417590c9bc99b359f3ac66d1346ecd54249888e
```
mikix, line 40: This file is mostly just a trimmed down version of the ETL's compose file. Except these two lines - I updated to the latest commit (didn't seem like an important commit, just wanted to be up to date) and updated the name - they renamed the model - old name still works, but just again to get up to date.

I used a `compose-nlp.yaml` filename, which won't be found by default so you need a `-f compose-nlp.yaml` arg, but I figured I didn't want to confuse anyone that this was how you could run Library itself.

mikix: Notably, these two stanzas use an old version of vLLM - I tried updating to latest, but got an error and didn't want to debug it. And I've never successfully run the llama4-scout vLLM. So some work for the future.

docs/workflows/nlp.md
```diff
@@ -166,6 +166,80 @@ You can pass these to Cumulus Library when building a study and any NLP workflow
...
+     On Linux, this will be somewhere like `~/.cache/cumulus-library/nlp/{my_study}/{table}_v0/`
+   - The tables themselves (that point to those parquet files) will be in the database path you
+     gave Cumulus Library (i.e. `./testing.db`).
+1. You can inspect the parquet files with a tool like
+   [parquet-tools](https://pypi.org/project/parquet-tools/).
+   - Call `parquet-tools show ~/path/to/parquet/files/*` to see a dump of their contents.
+1. You can inspect the resulting database with a tool like
+   [duckdb-cli](https://pypi.org/project/duckdb-cli/).
+   - Call `duckdb ./testing.db` and use commands like `show tables` and `select * from {table}`
+     to see the table contents.
```
dogversioning: should we suggest using the `-ui` flag for this?

docs/compose-nlp.yaml
```diff
@@ -0,0 +1,57 @@
+services:
+
+  gpt-oss-120b:
+    # Needs 80GB of GPU memory. A g6.12xlarge EC2 instance should work.
```
dogversioning, line 4: 💸 💸 💸 💸 💸 💸

docs/compose-nlp.yaml
dogversioning, line 1: I don't know that this should be at the project root? i'm open to nattering, but maybe under `docs/`?

dogversioning (approved):

docs/compose-nlp.yaml
mikix: Done

docs/workflows/nlp.md
mikix: Done

Issue #544, opened 2026-05-05
Add column "onsetDateTime" to CORE__condition


comorbidity:
**Problem Statement**
CORE__condition recordedDate fields state when the diagnosis was recorded not when the condition started. 

**Desired behavior**
Requesting CORE__condition include a column [onsetdatetime](https://build.fhir.org/condition-definitions.html#Condition.onset_x_) :  

```select count(*) cnt, onsetdatetime  from condition group by onsetdatetime order by cnt desc```

**Additional context**
General feature, not study specific.

**Comment by comorbidity** [comment]
Previous versions of Bulk FHIR downloads did not provide this data, current downloads are showing this data is often populated.

**Comment by comorbidity** [comment]
NOTICE: It is the responsibility of the study designer (me in this case, IBD) to validate that Condition **onset**  is true for the given study use case, for example, in IBD the gold standard is Endoscopy. 

From this Epic query, 2024 is when "onset" date appears, which could be a data load artifact and not truly the date of pathophysiology 

```
select  count(distinct condition_ref) cnt, condition.onsetdatetime
from    condition, ibd__cohort_casedef as casedef 
where   casedef.condition_ref = concat('Condition/', condition.id)
group by condition.onsetdatetime
order by cnt desc 
```

comorbidity:
This has been fixed.

PR #545, opened 2026-05-06
Use bare timestamps for all uncoerced FHIR date fields

dogversioning:
This stops coercion of timestamps to days as the default behavior to better allow for intra-day measurement of care. In many cases, this means adding a new timestamp field where one did not exist before.

This partially addresses #544 by at least not dropping information on the floor, and adding the missing must support fields to condition.

### Checklist
  - If you've changed the structure of a table, you may need to run `generate-md` <- this is broken. #546 


.github/workflows/pypi.yaml
```diff
@@ -2,7 +2,7 @@ name: PyPI
 
 on:
   release:
-    types: [created]
+    types: [published]
```
dogversioning, line 5: Unrelated, but this came up during release cutting - we want to push to pypi on publish, not if we have something in a draft state.

Luke-Sikina (approved):

PR #553, opened 2026-05-08
Fix NLP parquet table schemas

mikix:
Now databases can handle nested schemas and lists.

### Checklist

### DIFF
--- a/cumulus_library/actions/importer.py
+++ b/cumulus_library/actions/importer.py
@@ -2,7 +2,7 @@ import pathlib
 import tempfile
 import zipfile
 
-import pandas
+import pyarrow.parquet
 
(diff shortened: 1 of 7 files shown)


dogversioning (approved):
Just checking out of paranoia since the tests on the athena side are all mock-y - what studies did you run to test this against live databases?

mikix:
> Just checking out of paranoia since the tests on the athena side are all mock-y - what studies did you run to test this against live databases?

Just the example_nlp one.

But also, the "parquet to athena" logic is copied direct from the ETL, which has been using this parquet-upload logic for a couple months. Hopefully it would have surfaced some bugs from Dylan's experiments in that time.

Issue #529, opened 2026-04-29
Create data dictionaries for column metadata


dogversioning:
We should have an artifact that allows a study author to describe the contents of a particular variable used in a study. We'll attempt to create the framework for this for them from the study tables. We'll write it to some default location, but allow a user to override that value in a study manifest

A study data dictionary should contain the following columns (names are subject to modification):
variable name: the name of a column
display name: a user friendly name for a column
short description: a summary of the contents of the column
long description: an optional column that contains additional errata about the column

We may also want a column to indicate if something is in use or not.

dogversioning:
we may also want to include the appropriate dashboard data type in here, though allowing a user to override that would be tricky.

PR #555, opened 2026-05-11
Added support for data dictionaries

dogversioning:
This adds a new data construct available to load from the manifest, a 'data dictionary'. This format is a list of metadata associated with a given column name. If an item in the dictionary has the same name as an item in one of the export tables,
downstream the aggregator will (eventually) make this available to the dashboard.

This addresses #529, and starts progress on https://github.com/smart-on-fhir/cumulus-aggregator/issues/210

### Checklist
- [ ] Consider if documentation in `docs/` needs to be updated <- this will be done after some other issues in the feature are in
- [ ] Update template repo if there are changes to study configuration in `manifest.toml` <- deferred as above

### DIFF
--- a/cumulus_library/study_manifest.py
+++ b/cumulus_library/study_manifest.py
@@ -438,9 +438,6 @@ class StudyManifest:
             return ""
         return self.get_study_prefix()
 
-    def get_data_dictionary(self) -> list[dict] | None:
-        return self._study_config.get("data_dictionary")
-
     def copy_manifest(self, path: pathlib.Path):
         """Writes a copy of the manifest to the provided path.
 
--- a/cumulus_library/study_manifest.py
+++ b/cumulus_library/study_manifest.py
@@ -1,7 +1,9 @@
 """Class for loading study configuration data from manifest.toml files"""
 
 import copy
+import csv
 import dataclasses
+import json
 import pathlib
 import re
 import shutil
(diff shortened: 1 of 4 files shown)


cumulus_library/study_manifest.py
```diff
@@ -156,6 +182,44 @@ def _format_action(self, action):
...
+                    data_dict = msgspec.json.decode(file, type=DataDictionary)
+                elif dict_file.endswith(".toml"):
+                    file = file_buf.read()
+                    data_dict = msgspec.toml.decode(file, type=DataDictionary)
+                else:
+                    raise errors.StudyManifestParsingError(
+                        f"{dict_file} is not a supported format. Expected formats: csv,json,toml"
+                    )
+        except msgspec.ValidationError as e:
+            if "name" in str(e):
```
Luke-Sikina: This seems like a pretty broad check that could lead to misleading error messages if `name` ends up in other error messages. Is there any way you can make this more specific?

Luke-Sikina (approved):

cumulus_library/study_manifest.py
dogversioning: so, i can - the general gist here is that i'm trying to catch two errors, of the same type, thrown by the same library, so the finer grained i make this the more likely it is to get missed, but that brittleness exists regardless, but i suppose we just have to accept that.

Issue #530, opened 2026-04-29
Add a description field to counts workflow


dogversioning:
Each exported counts table should support a freeform description field where a user can define the goal of a particular counts table. We'll use this data to hydrate the manifest during upload to pass this data on to the aggregator.

We may handle this by adding a field to the export actions in the manifest, which would allow a user to specify a description there without actually using the counts workflow, if they desired.

PR #558, opened 2026-05-13
Reading workflows to populate export configs

### LINKED ISSUE #533: Handle upload of data dictionary and column metadata
These new artifacts will need to be processed in some way during the aggregator upload. It will likely involve writing new data to the manifest, but may warrant a standalone file, depending.

dogversioning:
This allows a workflow to be used in exports, allowing a study author to keep all the relevant data in the workflow itself and not worry about the manifest for managing exports.

As a byproduct of this, it breaks out the parsing logic into its own module to prevent circular imports between the CountsBuilder and the StudyManifest.

Finishes #530 and resolves #533 by just updating the manifest itself with the data.

### Checklist
- [ ] Consider if documentation in `docs/` needs to be updated <- will do standalone after this
- [ ] Update template repo if there are changes to study configuration in `manifest.toml` <- will do standalone after docs

### DIFF
--- a/cumulus_library/builders/counts_utils.py
+++ b/cumulus_library/builders/counts_utils.py
@@ -1,0 +1,65 @@
+import pathlib
+import sys
+
+import msgspec
+
+from cumulus_library.builders.statistics_templates import counts_templates
+
+
+class CountsWorkflowAnnotation(msgspec.Struct, forbid_unknown_fields=True):
+    field: str
+    join_table: str
+    join_field: str
+    columns: list[list[str]]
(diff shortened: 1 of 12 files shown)


cumulus_library/builders/counts_utils.py
dogversioning, line 1: noting this is just a copy/paste of what used to be in counts_builder.

Dtphelan1 (approved):
This looks reasonable to me, especially after talking with Matt over slack. Approving conditionally, once tests are fixed you should be good to go.

PR #560, opened 2026-05-15
Data documentation

dogversioning:
This summarizes the last few commit's worth of work with an explanation for end users about how to use the study documentation features.

### Checklist
- [ ] Update template repo if there are changes to study configuration in `manifest.toml` <- todo on 6.1 release

### DIFF
--- a/docs/documenting-data.md
+++ b/docs/documenting-data.md
@@ -1,0 +1,191 @@
+---
+title: Documenting data
+parent: Library
+nav_order: 4
+# audience: Clinical researcher with low to medium familiarity with project
+# type: tutorial
+---
+
+# Documenting data
+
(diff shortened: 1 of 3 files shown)


docs/documenting-data.md
```diff
@@ -0,0 +1,191 @@
...
+type = "string"
+
+[[fields]]
+name = "system"
+description = "Coding System"
+details = ""
+type = "string"
+```
+
+### What fields to document
```
Dtphelan1: Really minor, but for readability I'd suggest this small change 
```suggestion
### Which fields should be documented?
```

Dtphelan1 (approved):
One small inline change to suggest, but otherwise this looks good to me!

Issue #580, opened 2026-06-18
Actions and Submanifests Should Be Able To Be Self-Skippable


Dtphelan1:
**Problem Statement**
As NLP runs get integrated into library manifest definitions, study construction can stall trying to run NLP on entire cohorts of patients. While we want to be able to expose the relevant NLP submanifests in the study manifest, we'd like to skip them by default and run them after constructing our study and exploring the subsequent casedefinition. 

**Desired behavior**
Actions and submanifests should be able to declare their desire to be skipped by default. When building the study, those actions will be skipped; when users need to run those sub-actions, they can point at the appropriate stage/action in particular.

PR #582, opened 2026-06-24
Allow actions to be skipped

dogversioning:
This allows marking actions as skippable in the manifest - those actions will not be run unless the stage is explicitly called. Addresses #580 

### Checklist


Dtphelan1 (approved):
Looks great to me.

Issue #559, opened 2026-05-13
NLP Workflows Should Support Incremental Table Creation


Dtphelan1:
**Problem Statement**
As far as I can tell, `cumulus-library`'s NLP workflows must fully complete before writing any results to Athena, preventing any down-stream work on transforming/evaluating the results of NLP tasks until the full job is complete (or requiring jobs to be manually run in batches by users to side-step this)

**Desired behavior**
In [ETL's NLP pipeline](https://github.com/smart-on-fhir/cumulus-etl/blob/c3c475ab46ba29b4daf171e9d474c71c337b1204/cumulus_etl/etl/pipeline.py#L41-L67), users are able to specify a `--batch-size` command to incrementally write tables of n-many rows as they are processed, instead of waiting for the entire job to finish. A similar sort of CLI-arg driven customization here would be a boon for UX.

**Additional context**
N/A

dogversioning:
So this is doable, but it does create a problem for the existing UX - all of the table progress bars are tied to the BaseTableBuilder's `execute_queries()` method, which assumes the last step you're doing is creating a table, which handles the progress advancement. Some options:

- We could create the table before any data is available in the prepare_queries() method, and then load the data in. I don't like this for two reasons: A) it creates an empty table, which can have negative impacts on column lookup, and B) it decouples everything from the CLI UX.
- We could, in the instance of a batch job, upload the first slug, create the table in the normal method, and then process and upload the remaining slugs in the post execution hook. This avoids A above, and has a lesser version of B, where at least the CLI UX would advance a progress bar, but it would then be semi-incorrect while the remaining notes process
- We could overhaul the execute step to take things other than queries, like a callback, or a specific dataclass tied to a certain kind of job (NLP in this case), which would run something other that just queries. This is probably the correct way to do this, and, of course, is also the most complex.

Dtphelan1:
Update on this: it looks like [library already supports this implicitly](https://github.com/smart-on-fhir/cumulus-library/blob/dfc3a0b226866f795e6e8ef8511dfba4063edc8b/cumulus_library/note_utils.py#L108), with `NlpConfig.note_limit`. We just need a way of being able to set that value via the CLI like we do with other config variables, e.g. `phi_dir`! I can try taking a stab at this next week!

PR #583, opened 2026-06-24
NLP chunking, test cleanup

dogversioning:
This adds a new cli arg, `--nlp-chunksize` (since batching is overloaded here) that allows control over how many results are kept locally before they are written out.

While we're here, this also does some cleanup to the nlp tests, which were all using the global cache directory and thus were behaving poorly when running concurrently.

Addresses #559 

### Checklist


Dtphelan1 (approved):
This looks like exactly what I'm looking for. Tysm and also note this closes #559

PR #588, opened 2026-07-08
Dtp/fix nlp model name resolution

Dtphelan1:
It appears like sometime in the past few months (maybe April?) there was a shift in the alignment between model ids and actual deployment ids in the OpenAI API wrapper. The result is that trying to run the latest library's NLP processes fails (just try running any NLP with `gpt54-nano` as the model id CLI arg). This PR sidesteps an unnecessary failure by just checking to see if a simple request is achieved against the model before failing. 

### Checklist

### DIFF
--- a/cumulus_library/builders/nlp/models.py
+++ b/cumulus_library/builders/nlp/models.py
@@ -237,9 +237,17 @@ class OpenAIProvider(Provider):
             raise UnreachableModel(f"NLP server is unreachable: {exc}") from exc
 
         if self.model_name not in names:
-            raise errors.CumulusLibraryError(
-                f"NLP server does not have model ID '{self.model_name}'."
-            )
+            # Instead of failing, let's do a request to check if the deployment works.
+            # (This is a workaround for Azure, which uses deployment names instead of model names.)
+            try:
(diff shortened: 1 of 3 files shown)


cumulus_library/builders/nlp/models.py
```diff
@@ -237,9 +237,17 @@ def post_init_check(self) -> None:
...
+            # (This is a workaround for Azure, which uses deployment names instead of model names.)
+            try:
+                self.client.chat.completions.create(
+                    model=self.deployment,
+                    messages=[{"role": "system", "content": "Hello"}],
+                )
+            except openai.APIError as exc:
+                raise errors.CumulusLibraryError(
+                    f"NLP server does not have model ID '{self.model_name}: {exc}'."
+                )
```
dogversioning, line 250: can we write a basic test for this?

dogversioning (approved):
Food for thought - should we externalize model definition a bit more?

cumulus_library/builders/nlp/models.py
Dtphelan1: ofc ofc

PR #598, opened 2026-07-28
When a file has changed locally and the file_upload workflow is in use, upload the changes to Athena, regardless of force upload flag

### LINKED ISSUE #589: Minor: config_type = "file_upload" should check SHA hash to automatically upload if needed
**Problem Statement**
Love the upload caching but sometimes it can mask that I have made a change and the --force-upload flag is a good enough fail safe and/but still a bit distracting from staying focused on the study itself.

**Desired behavior**
Hash with MD5/SHA/etc and check if the CSV upload contents have changed, if so, automatically upload on behalf of user. 

**Additional context**
`--force-upload` appears to be working better in latest cumulus-library releases, in the past I have seen scenarios where I was not convinced this was actually working. Would be a non-issue if hash was checked and handled automatically for user.

msa2984:
Resolves #589 
### Testing
- I followed the steps in https://docs.smarthealthit.org/cumulus/library/example.html
   - Manually edited one of the lines in the `snomed_bronchitis.csv`, verified the change was uploaded to S3 + Athena. 

### Checklist

### DIFF
--- a/cumulus_library/databases/athena.py
+++ b/cumulus_library/databases/athena.py
@@ -188,8 +188,10 @@ class AthenaDatabaseBackend(base.DatabaseBackend):
             if res["KeyCount"] > 0:
                 local_file_hash = hashlib.md5(file.read_bytes(), usedforsecurity=False).hexdigest()
 
-                if "ETag" in res and res["ETag"].strip('" ') == local_file_hash:
-                    return f"s3://{bucket}/{s3_key}"
+                if "Content" in res and "ETag" in res["Contents"][0]:
+                    etag = res["Contents"][0].get("ETag", "").strip('" ')
+                    if etag and etag == local_file_hash:
+                        return f"s3://{bucket}/{s3_key}"
 
                 res = s3_client.get_object(
                     Bucket=bucket,
--- a/cumulus_library/databases/athena.py
+++ b/cumulus_library/databases/athena.py
@@ -186,15 +186,16 @@ class AthenaDatabaseBackend(base.DatabaseBackend):
                 Prefix=f"{s3_key}/{remote_filename}",
             )
             if res["KeyCount"] > 0:
-                local_file_hash = hashlib.sha256(
-                    file.read_bytes(), usedforsecurity=False
-                ).hexdigest()
+                local_file_hash = hashlib.md5(file.read_bytes(), usedforsecurity=False).hexdigest()
+
+                if "ETag" in res and res["ETag"].strip() == local_file_hash:
+                    return f"s3://{bucket}/{s3_key}"
 
                 res = s3_client.get_object(
                     Bucket=bucket,
                     Key=f"{s3_key}/{remote_filename}",
                 )
-                res_hash = hashlib.sha256(res["Body"].read(), usedforsecurity=False).hexdigest()
+                res_hash = hashlib.md5(res["Body"].read(), usedforsecurity=False).hexdigest()
 
                 if res_hash == local_file_hash:
                     return f"s3://{bucket}/{s3_key}"
--- a/cumulus_library/databases/athena.py
+++ b/cumulus_library/databases/athena.py
@@ -188,7 +188,7 @@ class AthenaDatabaseBackend(base.DatabaseBackend):
             if res["KeyCount"] > 0:
                 local_file_hash = hashlib.md5(file.read_bytes(), usedforsecurity=False).hexdigest()
 
-                if "ETag" in res and res["ETag"].strip() == local_file_hash:
+                if "ETag" in res and res["ETag"].strip('" ') == local_file_hash:
                     return f"s3://{bucket}/{s3_key}"
 
                 res = s3_client.get_object(
--- a/cumulus_library/databases/athena.py
+++ b/cumulus_library/databases/athena.py
@@ -5,6 +5,7 @@ features, like pyathena's async cursors, to simplify cross-db behavior.
 """
 
 import collections
+import hashlib
 import os
 import pathlib
 from concurrent import futures
@@ -185,7 +186,26 @@ class AthenaDatabaseBackend(base.DatabaseBackend):
                 Prefix=f"{s3_key}/{remote_filename}",
             )
             if res["KeyCount"] > 0:
-                return f"s3://{bucket}/{s3_key}"
+                if not file.exists():
+                    return f"s3://{bucket}/{s3_key}"
+
+                local_file_hash = hashlib.sha256(
+                    file.read_bytes(), usedforsecurity=False
+                ).hexdigest()
+
+                res = s3_client.get_object(
+                    Bucket=bucket,
+                    Key=f"{s3_key}/{remote_filename}",
+                )
+
+                local_file_hash = hashlib.sha256(
+                    file.read_bytes(), usedforsecurity=False
+                ).hexdigest()
+                res_hash = hashlib.sha256(res["Body"].read(), usedforsecurity=False).hexdigest()
+
+                if res_hash == local_file_hash:
+                    return f"s3://{bucket}/{s3_key}"
+
         with open(file, "rb") as b_file:
             s3_client.put_object(
                 Bucket=bucket,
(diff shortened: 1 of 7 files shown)


docs/core-study-details.md
```diff
@@ -22,7 +22,7 @@ for validating the integrity of the data you're extracting from your EHR system
 If you are authoring a study, and are focused only on clinical analysis (i.e. you
 aren't working on data quality/data governance issues), we **strongly** recommend you
 use the core study as the starting point for your own work. See 
-[Creating Studies](./creating-studies.md)
+[Creating Studies](./example.md)
```
msa2984, line 25: I just noticed when I was combing through the Smart Health IT documentation that the creating-studies.md is a dead link, and this seems to be the intended documentation anyways.

tests/databases/test_databases.py
```diff
@@ -237,6 +238,8 @@ def test_upload_file_athena(mock_botocore, args, sse, keycount, expected, raises
     mock_clientobj = mock_botocore.ClientCreator.return_value.create_client.return_value
     mock_clientobj.get_work_group.return_value = mock_data
     mock_clientobj.list_objects_v2.return_value = {"KeyCount": keycount}
+    if keycount > 0:
```
msa2984, line 241: The if isn't really necessary, since `test_athena.py` verifies that get_object is only called under certain conditions so I can just add in the mock

tests/databases/test_athena.py
```diff
@@ -68,16 +70,109 @@ def test_upload_parquet_response_handling(mock_session):
     s3_client = mock.MagicMock()
     with open(path / "test_data/aws/boto3.client.s3.list_objects_v2.json") as f:
         s3_client.list_objects_v2.return_value = json.load(f)
+
+    with open(path / "test_data/file_upload/upload.parquet", "rb") as f:
+        s3_client.get_object.return_value = {"Body": io.BytesIO(f.read())}
+
     mock_session.return_value.create_client.return_value = s3_client
     resp = db.upload_file(
-        file=path / "test_data/count_synthea_patient.parquet",
```
msa2984, line 73: as far as I could tell, this count_synthea_patient.parquet doesn't actually exist at this path, but I can see there are a few upload.parquet in the repo. I've updated this test to actually verify uploading parquet files.

cumulus_library/databases/athena.py
```diff
@@ -185,7 +186,19 @@ def upload_file(
                 Prefix=f"{s3_key}/{remote_filename}",
             )
             if res["KeyCount"] > 0:
-                return f"s3://{bucket}/{s3_key}"
+                local_file_hash = hashlib.sha256(
```
msa2984: we could probably also add a date modified check here? i.e., if the date of the local file is newer AND the content is different, upload the local file

docs/core-study-details.md
dogversioning: nice catch, thanks.

cumulus_library/databases/athena.py
dogversioning: i think this check will cover 99% of our real cases as is, so I think we're probably ok for now.

cumulus_library/databases/athena.py
```diff
@@ -185,7 +186,19 @@ def upload_file(
...
+                    file.read_bytes(), usedforsecurity=False
+                ).hexdigest()
+
+                res = s3_client.get_object(
+                    Bucket=bucket,
+                    Key=f"{s3_key}/{remote_filename}",
+                )
```
dogversioning, line 194: so every once in awhile, the file that we're doing this upload check on is huge - we have a few coding system data loaders that can have files in the GB range, which don't use this currently but conceivably could.

i don't think its flawless, but we might want to think about checking the etag header on the S3 object first? it's not guaranteed to match, but we could skip the download check if it does.

tests/databases/test_athena.py
dogversioning: oh dear - there is the `test_data/upload/upload__count_synthea.patient.cube.parquet`, which should have had the same contents after it got moved, but i would have expected this test to fail when we deleted that file... not something to fix as part of this, but maybe we should make an issue for more assertive error throwing?

tests/databases/test_athena.py
```diff
@@ -68,16 +70,109 @@ def test_upload_parquet_response_handling(mock_session):
...
+
+
+@pytest.mark.parametrize(
+    """force_upload,
+list_objects_result,
+remote_body,
+local_bytes,
+expected_get_object_call_count,
+expected_put_object_call_count
+""",
```
dogversioning: starting off saying: i think this is fine.

nattering about style - in the past for things like this, i think we've tended to prefer joining strings that aren't sql/sql templates via parenthesis concatenation, just so that we're not sensitive to trailing spaces and can maintain indents and the like. here this would get detected via a test failing, so at least there's some programmatic check around it. we don't state this anyplace, and its :very: weakly held as an opinion, so if you have a vibe for this format we can roll with it.

tests/databases/test_athena.py
```diff
@@ -68,16 +70,109 @@ def test_upload_parquet_response_handling(mock_session):
...
+    [
+        pytest.param(
```
dogversioning, line 99: oh my god i had no idea, what a great feature

tests/databases/test_databases.py
dogversioning: never hurts to be paranoid about changes in other modules.

dogversioning (approved):

cumulus_library/databases/athena.py
msa2984: @dogversioning I can add this check, however, I think in quite a few scenarios the ETag won't match anything we can easily calculate... for example, from the AWS [documentation](https://docs.aws.amazon.com/AmazonS3/latest/API/API_Object.html) :

> ETag
> The entity tag is a hash of the object. The ETag reflects changes only to the contents of an object, not its metadata. The ETag may or may not be an MD5 digest of the object data. Whether or not it is depends on how the object was created and how it is encrypted as described below:
> Objects created by the PUT Object, POST Object, or Copy operation, or through the AWS Management Console, and are encrypted by SSE-C or SSE-KMS, have ETags that are not an MD5 digest of their object data.

tests/databases/test_athena.py
msa2984: I wasn't sure if it was the cube file, I can change this test to point to that file instead. 

So... I think interestingly enough, this feature actually exposed that if you specify a file that doesn't exist, the original code would still return the formatted URL `f"s3://{bucket}/{s3_key}"`

tests/databases/test_athena.py
msa2984: 🤦‍♀️ No, parens concatenation is way better, I've changed this

cumulus_library/databases/athena.py
dogversioning: yeah, i agree it's not perfect - we might eke it out here because we're primarily loading stuff with boto. we could run a quick check to see if we're getting hits with the upload method we're using?

tests/databases/test_athena.py
dogversioning: https://github.com/smart-on-fhir/cumulus-library/issues/601

cumulus_library/databases/athena.py
msa2984: so I've tested this locally, and I do not get an ETag match even when I am trying to upload an identical file. This is likely because of the KMS encryption, as called out in the documentation this will not result in an ETag that is an MD5 hash of the file. 

I'm okay with leaving this check in, although I think it will not be a shortcut for 99.9% of scenarios.

dogversioning: i mean - KMS is now default S3 behavior, and this data is sensitive and needs to be encrypted, so we're looking at 100% failure - so maybe this was a bad suggestion

msa2984: > https://github.com/smart-on-fhir/cumulus-library/issues/603

I've also removed this ETag check

PR #600, opened 2026-07-29
Fallback to boto session for credentials

dogversioning:
Our current assumption was 'A user will provide credentials, either via AWS env vars, cli, or via `.aws/credentials` lookup, but this excludes a couple of trailing [valid boto credential options](https://docs.aws.amazon.com/boto3/latest/guide/credentials.html#configuring-credentials) - so if we don't find one via our current methods, we'll just spin up a boto session and call get_credentials, to see if we're in an EC2 and looking for IAM as a source.

### Checklist


dogversioning:
Noting that I'm waiting on validation from the party reporting this issue before merging.

Luke-Sikina (approved):

msa2984 (approved):

PR #605, opened 2026-07-30
Set and get checksum data from S3 to compare file contents.

### LINKED ISSUE #603: Add a way to quickly validate the contents file without downloading it from S3
**Problem Statement**

Spinning off of the comments here: https://github.com/smart-on-fhir/cumulus-library/pull/598/changes#r3683231348

As discussed in the comments, ETag is unlikely to be a reasonable shortcut. However, downloading files from S3 every time can add significant time, especially if the files are large. Perhaps we can add a unique piece of metadata which can be calculated, much like a hash or checksum and append it when we upload the file. 

**Desired behavior**
- Cumulus Library can compare local and remote versions of a file without downloading the full file.

msa2984:
Fixes #603 

### Checklist

### DIFF
--- a/cumulus_library/databases/athena.py
+++ b/cumulus_library/databases/athena.py
@@ -4,6 +4,7 @@ See base.py for design rules of thumb - notably, don't use non-PEP249 database
 features, like pyathena's async cursors, to simplify cross-db behavior.
 """
 
+import base64
 import collections
 import hashlib
 import os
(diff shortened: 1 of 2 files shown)


cumulus_library/databases/athena.py
```diff
@@ -192,22 +193,28 @@ def upload_file(
...
 
-                if res_hash == local_file_hash:
-                    return f"s3://{bucket}/{s3_key}"
+                # If the checksum exists and is equal to the local file's checksum, skip upload.
+                # Otherwise - overwrite if the files are different, or if the checksum doesn't
+                # exist. Metadata in S3 is immutable so the object needs to be overwritten.
+                if "ChecksumSHA256" in res:
+                    res_hash = base64.b64decode(res["ChecksumSHA256"])
+                    if res_hash == local_file_hash:
+                        return f"s3://{bucket}/{s3_key}"
```
dogversioning, line 217: i think this is fine, but i just want to confirm: we agree that the downstream caller doesn't need to care about whether the replacement happened or not? (or at least, we agree until we discover otherwise)

dogversioning (approved):

cumulus_library/databases/athena.py
msa2984: the downstream caller doesn't need to know - but also, since we are passing in the checksum, AWS will actually compare the value with its own calculation and throw an error if they do not match. AKA the user will know because an exception will surface

PR #607, opened 2026-07-31
NLP: Results Tables Follow ETL's Naming Convention, Closing #596

### LINKED ISSUE #596: NLP Results Tables Follow Unexpected Naming Conventions
**Describe the bug**
Separate from NLP cache concerns, NLP tables for completed NLP annotations are materialized using a naming schema that doesn't mirror expectations set by ETL's NLP. We should update the naming scheme used by Library's NLP to use both the preserved `nlp`prefix, and the full name format listed here. 

**To Reproduce**
We'll consider the scenario where we run NLP for Treatment annotations using v7 of the pydantic class and gpt-oss-120b as our LLM: 

The table that library finally exposes to athena is `ibd__treatment`, with source data coming from tables in `cumulus-athena-****-****/cumulus_user_uploads/results/ibd` that follow the following pattern: `treatment_vN`, for each version N of NLP run.

**Expected behavior**
Inspecting data here same location, we can see the `user_uploads` directories made by ETL include more information in their slugs;`nlp_treatment_gpt_oss_120b_v7` tells us 1) this is NLP data, 2) which model generated it; 3)what task table it corresponds to; and 4) what version it was run with. Additionally, the final athena tables - `ibd__nlp_treatment_gpt-oss-120b` - aggregate all versions in one location. Existing workflows are built around the assumption that NLP data will live in tables following these naming conventions.

Dtphelan1:
Should close #596 . Have to pass around the nlp_config a bit more around the driver file, but other than that it wasn't too hard of a fix. Adds some new tests to cover this convention, and updates the example tables around other tests/docs to use the new convention. 

### Checklist

### DIFF
--- a/cumulus_library/studies/example_nlp/calculate_ranges.sql
+++ b/cumulus_library/studies/example_nlp/calculate_ranges.sql
@@ -1,3 +1,6 @@
+-- NLP result tables are named `{study}__nlp_{task}_{model}`, so the model you run NLP with
+-- is part of the table name (this study assumes the gpt-oss-120b model). If you run this study
+-- with a different model, update the table referenced below to match.
 CREATE TABLE example_nlp__range_labels AS
 SELECT
     src.note_ref,
@@ -16,4 +19,4 @@ SELECT
         WHEN src.result.age < 65 THEN 'middle aged (45-64)'
         ELSE 'aged (65+)'
     END AS label
-FROM example_nlp__age AS src
+FROM example_nlp__nlp_age_gpt_oss_120b AS src
--- a/docs/workflows/nlp.md
+++ b/docs/workflows/nlp.md
@@ -48,9 +48,11 @@ contains details on the expectations of each value.
 config_type="nlp"
 
 # You define NLP tasks/tables with the 'tables' dictionary.
-# The keys in the table dictionary define your table names, and the study name will
-# automatically be prepended from the prefix in your manifest.
-# So an entry like [tables.my_table] results in a table like `my_study__my_table`.
+# The keys in the table dictionary define your table names, but the final table name also
+# includes the study prefix from your manifest, an `nlp` marker, and the model you ran with.
+# So an entry like [tables.my_table], run against the gpt-oss-120b model, results in a table
+# like `my_study__nlp_my_table_gpt_oss_120b`.
+# See "Result Table Names" below for more details.
 [tables.table_1]
 
 # `version` will be a number that you will increment every time you change the table definition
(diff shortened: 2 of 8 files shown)


tests/builders/test_nlp_builder.py
```diff
@@ -814,29 +921,32 @@ def test_write_to_athena(mock_openai_client, mock_boto_client, tmp_path, note_so
...
-    id_path = "s3://testbucket/athena/cumulus_user_uploads/testdb/example_nlp/task_v0.ids"
-    with mem_fs.open(id_path, "r") as f:
+    with mem_fs.open(f"{upload_dir}.ids", "r") as f:
         assert f.read() == "DiagnosticReport/hello\n"
 
-    # And confirm the query looks right
+    # And confirm the queries look right
     assert builder.queries == [
-        "CREATE EXTERNAL TABLE IF NOT EXISTS `main`.`example_nlp__task` ( note_ref STRING, "
+        "DROP TABLE IF EXISTS `main`.`example_nlp__nlp_task_gpt_oss_120b`;",
```
dogversioning: I'm noting that this was added midstream and I don't think it remained relevant, so I was going to just pull it. Between this and the other failing test, you should review if this is the behavior you expect, or if something got broken along the way.

tests/builders/test_nlp_builder.py
```diff
@@ -206,6 +208,111 @@ def test_filter(mock_client, tmp_path, mock_db_config):
...
+    assert rows[0]["task_version"] == 0
+
+    # The old version's results are still sitting on disk, but the table should now be
+    # backed by the new version's folder.
+    build(version=1, answer=2)
+    assert cfs.FsPath(tmp_path, "nlp", "example_nlp", "nlp_task_gpt_oss_120b_v0").exists()
+    rows = read_rows(mock_db_config, "example_nlp__nlp_task_gpt_oss_120b")
+    assert rows[0]["result"] == {"hello": 2}
+    assert rows[0]["task_version"] == 1
+
```
dogversioning: related to the other comment, this no longer happens due to that missing drop statement.

tests/builders/test_nlp_builder.py
Dtphelan1: Oh great catch. Yeah I think we're good to drop this. I think I just fundamentally misunderstood what the expectations were for the dropping/not dropping of tables as part of the builders. Good to pull it out

tests/builders/test_nlp_builder.py
Dtphelan1: Heard. We can update this as needed.

cumulus_library/studies/example_nlp/calculate_ranges.sql
```diff
@@ -1,3 +1,6 @@
+-- NLP result tables are named `{study}__nlp_{task}_{model}`, so the model you run NLP with
+-- is part of the table name (this study assumes the gpt-oss-120b model). If you run this study
+-- with a different model, update the table referenced below to match.
```
dogversioning, line 3: philosophizing, no action other than maybe making an issue: should we offer some jinja macros for handling nlp table name formats?

docs/workflows/nlp.md
```diff
@@ -128,6 +130,30 @@ reject_by_word: ["default", "words"]
 reject_by_regex: ["default", regex"]
 ```
 
+### Result Table Names
+
+NLP results are named the same way that [Cumulus ETL](https://docs.smarthealthit.org/cumulus/etl/)
+names the NLP tables it creates, so that all NLP results in your database look alike:
```
dogversioning, line 136: I think by the time this documentation gets to anyone outside the team the note about the v1 version won't be super useful? We're really the only ones affected by it.

tests/test_cli.py
```diff
@@ -1231,3 +1233,99 @@ def test_cohorts_protected():
...
+        # The NLP results the builder wrote to disk for this task version (mock_cache_dir points
+        # the local upload folder at tmp_path)
+        folder = tmp_path / "nlp" / "nlp_versions" / f"nlp_task_gpt_oss_120b_v{version}"
+        con = duckdb.connect()
+        try:
+            return con.execute(f"select result from read_parquet('{folder}/*.parquet')").fetchall()
+        finally:
+            con.close()
+
+    # First run: NLP is hit once and the ETL-style table is created from the v0 results
```
dogversioning, line 1315: this is less critical than the other, but we'll forget about the specifics here and someone will be confused about it in a year (me)

cumulus_library/studies/example_nlp/calculate_ranges.sql
Dtphelan1: Worth thinking about for sure, but probably only once we really have another site using our tool in this way - there still may be some assumptions that break down in our table naming structure that we'll want to revisit.

docs/workflows/nlp.md
Dtphelan1: idk man I think this could reasonably be useful to someone understanding exactly where tables in athena are being fed data created from the NLP workflows (i.e. me in 6 months). If you feel strongly, I can remove this but I'd keep it for now if it were up to me.

tests/test_cli.py
Dtphelan1: Sorry which part are we in need of more specifics about? Do we want more preamble with this test_fn explaining the value of this test? Or is the phrase "ETL-style table" burying a lede?

docs/workflows/nlp.md
dogversioning: ok you know what, i rescind what i said on both ETL-style comments, its fine

PR #612, opened 2026-08-07
fix: remove SQL comments before splitting on semicolons

### LINKED ISSUE #610: Minor: comments containing ";" break cumulus-library SQL execution
**Problem Statement**
-- Comments like this in SQL files cause problems; SQL fails. 

**Desired behavior**
`cumulus-library` first removes comment lines prior to splitting on ";" 

**Additional context**
Has workaround: just don't use ";" in SQL comments. Easy enough but it happens.

AlessandroDiGioacchino:
## Summary
Fixes issue #610 by updating `base_utils.py/sql_parse` to properly handle SQL comments containing semicolons.

Fixes #610

## Changes made
- Updated `cumulus_library/base_utils.py` to fix semicolon splitting logic.

## How has this been tested?
- [x] Ran local unit tests

### Checklist

### DIFF
--- a/cumulus_library/base_utils.py
+++ b/cumulus_library/base_utils.py
@@ -70,15 +70,29 @@ def load_text(path: str) -> str:
 
 
 def parse_sql(sql_text: str) -> list[str]:
-    commands = []
+    tokens = sqlglot.tokenize(sql_text)
 
-    for statement in sql_text.split(";"):
-        parsed = []
-        for line in statement.splitlines():
+    indices = [t.start for t in tokens if t.token_type == sqlglot.TokenType.SEMICOLON]
+
+    commands = []
+    last_idx = 0
+    for idx in indices:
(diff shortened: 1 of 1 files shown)


cumulus_library/base_utils.py
```diff
@@ -70,15 +70,29 @@ def load_text(path: str) -> str:
...
+                parsed_lines.append(line)
+
+        cleaned = "\n".join(parsed_lines).strip()
+        if cleaned:
+            final_commands.append(cleaned)
+
+    return final_commands
 
 
 def filter_strip(commands) -> list[str]:
```
dogversioning: This function was only used by the above, and is now not covered - it's safe to remove

AlessandroDiGioacchino:
in the last two commits, I have added a unit test in `tests/test_base_utils.py` and deleted the now unused `filter_strip` function in `cumulus_library/base_utils.py`

